PR #137, opened 2013-12-17
ncm-metaconfig: support for filename attribute to force the filename instead the service key

stdweird:
Support for defining the filename allows for generated filenames and typical shorter/human readable paths in the templates e.g. 

```
"/software/components/metaconfig/services/{/etc/httpd/conf.d/ssl.conf}/contents" = "foo"
```

could become

```
variable NAME = "ssl";
"/software/components/metaconfig/services/vhost/filename" = format("/etc/httpd/conf.d/%s.conf", NAME);
"/software/components/metaconfig/services/vhost/contents" = "foo"
```

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -7,6 +7,30 @@ declaration template components/${project.artifactId}/schema;
 
 include { 'quattor/schema' };
 
+#
+# Check for conflicts between the filename and the service keys.
+# Each written file should correspond with excatly one service.
+#
+function is_${project.artifactId}_services = {
+    # /software/components/${project.artifactId}/services is passed as only argument
+    esc_filenames=nlist();
+    foreach(key;value;ARGV[0]) {
+        if (exists(value['filename'])) {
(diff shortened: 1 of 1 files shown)


jouvin:
I was about to reopen this issue after @loomis confirmed there was no way (at least no without an impact on perfs) to have interpolated variables in Pan paths (https://github.com/quattor/pan/issues/77, https://github.com/quattor/pan/issues/82). Here the use case is `metaconfig` with not hardcoded paths. But I agree with @ned21 remarks and I think it is better to say with the not-completely-trivial approach:

```
variable MYFILE = '/tmp/test';
'/software/components/metaconfig/services' = {
   SELF[escape(MYTEST)] = dict('contents', 'foo', ....);
   SELF;
}

As for @piojo's use case, I share the opinion that this is an anti-use-case! In the template library (not really to do with SCDB), we really avoid this kind of stuff and if there is the need for a "shared reference", we use a global variable. I don't share the view that global variables are bad by essence, as in a traditional language. In the Pan case, associated with the "default assignment" operator, there are the best way to ensure that every template needing to access the same information will use it consistently. As long as the variable is only assigned with `?=`, the overall result should remain consistent.

PR #405, opened 2015-01-07
ncm-sysctl: NoAction and sysctl.d support

ned21:
Switch ncm-sysctl to using CAF::Process instead of LC::Process. This
enables the component to be run in NoAction mode. It also causes a
non-zero exit code form /sbin/systctl to be logged as an error.
Previously a non-zero exit code was not being caught by the if()
statement.

Also add support for using sysctl.d instead of /etc/sysctl.conf. 

THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS
CONTRIBUTED IN CONNECTION WITH THIS PROGRAM.

THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
"AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE
DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS
BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR
BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF
LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING
NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. THIS
SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING
THIS OR ANOTHER EQUIVALENT DISCLAIMER AS WELL AS ANY OTHER LICENSE
TERMS THAT MAY APPLY.


ned21:
I have another branch that builds on this one adding sysctl.d support. Should I add the commits here or do you want to review them separately? https://github.com/ned21/configuration-modules-core/pull/1

stdweird:
@ned21 add them here, it's still small enough to review

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -22,23 +22,20 @@ use NCM::Component;
...
-
 use NCM::Check;
-use LC::Process qw(trun);
-
-# For convenience
-my $base = '/software/components/sysctl';
 
 
 ##########################################################################
 sub Configure {
```
stdweird: remove the `####` lines, the `{` should start on a new line
(i know this is not your code :smile: )

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -94,13 +91,16 @@ sub Configure {
...
     #
     if ( $changes ) {
-	unless (LC::Process::trun(300,"$sysctl_exe", "-p")) {
-	    $self->error('Failed to load sysctl settings from $configFile');
-	    return;
-	}
+        $self->verbose("Changes made to $configFile, running sysctl on it");
+        my $cmd = CAF::Process->new([$sysctl_exe, '-p', $configFile],
+                                    log => $self);
+	my $output = $cmd->output;
```
stdweird: indentation is wrong.

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -94,13 +91,16 @@ sub Configure {
     # execute /sbin/sysctl -p if any change made to sysctl configuration file
     #
     if ( $changes ) {
-	unless (LC::Process::trun(300,"$sysctl_exe", "-p")) {
-	    $self->error('Failed to load sysctl settings from $configFile');
-	    return;
-	}
+        $self->verbose("Changes made to $configFile, running sysctl on it");
+        my $cmd = CAF::Process->new([$sysctl_exe, '-p', $configFile],
```
stdweird: `sysctl_exe` should be a constant? then use upper case (i can't readily see where it is defined)

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -94,13 +91,16 @@ sub Configure {
...
+                                    log => $self);
+	my $output = $cmd->output;
+        $self->debug(1, "$sysctl_exe output: $output");
```
stdweird: i think `CAF::Process` with logger set already does this

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -94,13 +91,16 @@ sub Configure {
...
+        if ($? != 0) {
```
stdweird: just `if($?)` is fine

ncm-sysctl/src/main/perl/sysctl.pm
stdweird: ignore this.

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -94,13 +91,16 @@ sub Configure {
...
+            $self->error("Error loading sysctl settings from $configFile: ".
+                            "$output");
```
stdweird: `$output` is a string, and all log methods join the arguments ot a string, no need for the `.`.
```perl
$self->error("Error loading sysctl settings from $configFile: ", $output);
``` 
should be fine

ncm-sysctl/src/main/perl/sysctl.pm
ned21: It's pulled from the profile, so cannot  be a constant without breaking backwards compatibility.

stdweird:
looks good to me, but i'd like a new issue opened to discuss moving towards a more strict behaviour (via schema or directly in the component)
* files should end in `.conf`
* no arbitrary file location
 * either relative path or `/etc/sysctl.conf`
* no arbitrary commands
  * ideally detect correct command based upon OS?
  * systemd capable systems shoudl use `systemd-sysctl.service` to activate changes?

ned21:
I added the restriction on the filename. I doubt it's going to break anyone, and if it does, there's an easy workaround (drop the restriction from the schema). The additional breaking enhancements filed as #406

ned21:
@stdweird - I have tested it locally in our lab. Since the code looks good then I'll look into deploying it to our production plant in the next few weeks.

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -22,85 +22,109 @@ use NCM::Component;
...
-	    $changes += $st;
-	}
+    if ($configFile =~ m{/}) {
+        $self->debug(1, "confFile setting is a path");
+        $changes = $self->sysctl_file($configFile, $sysctl_exe, $variables);
+    } else {
+        $self->debug(1, "confFile is a relative filename, using sysctl.d");
+        $changes = $self->sysctl_dir($configFile, $sysctl_exe, $variables);
+        # update the configFile path for the sysctl execution
+        $configFile = "/etc/sysctl.d/$configFile";
```
stdweird: @ned21 remark wrt the file name. i think it would be better to make this 
```perl
$configFile = "/etc/sysctl.d/zzz-quattor-$configFile";
```
so on reboot, the quattor changes are always winning.

stdweird: actually, if you push this reasoining a bit further, relative file names should be mapped to `/etc/sysctl.d/zzz-quattor.conf`. i don't see any added value in allowing an arbitrary name (because ncm-sysctl can only manage one file), so for the new-style relative filenames we can already force the mapping.

ncm-sysctl/src/main/perl/sysctl.pm
```diff
@@ -22,85 +22,109 @@ use NCM::Component;
...
-				   goodre => '\s*'.$key.'\s*=\s*'.$value,
-				   good => "$key = $value",
-				   add => 'last'
-				  );
-	if ($st < 0) {
-	    $self->error("Failed to update sysctl $key (value=$value)");
-	} else {
-	    $changes += $st;
-	}
+    if ($configFile =~ m{/}) {
```
piojo: Mmmm... in this case, if `$configFile` is `foo/bar`, you'll get your file in `/tmp/foo/bar`. Do you want this?

ncm-sysctl/src/test/perl/configure.t
```diff
@@ -0,0 +1,34 @@
...
+
+
+my $cmp = NCM::Component::sysctl->new('sysctl');
+my $cfg = get_config_for_profile('simple');
+
+is($cmp->Configure($cfg), 1, "Configure succeeds");
+my $fh = get_file("/etc/sysctl.d/50-quattor.conf");
+ok($fh, "A file was actually created");
+isa_ok($fh, "CAF::FileWriter");
+like("$fh", qr{^kernel.sysrq = 1$}m, "Found kernel.sysrq variable");
```
piojo: You probably also want to check that `sysctl` is also executed if something changes:

```perl
my $cmd = get_command("sysctl ...");
# do something with $cmd here
```

piojo:
Couple of small comments, can go in after that.

ncm-sysctl/src/main/perl/sysctl.pm
ned21: I don't think we should restrict the component's flexibility by being so prescriptive. Sites may need to configure their filename to fit their use case (e.g. higher or lower precedence than some badly written rpm that drops a file into the same directory), and actually by using the component aliasing it's possible to use the same code to configure multiple files.

ncm-sysctl/src/main/perl/sysctl.pm
ned21: No, but that's not a regression over the current code that has the same issue! How would you suggest fixing it? A schema check that prevents any unless the first one is? Would this regexp match that? ^(/.+)|[^/]+)$

ncm-sysctl/src/test/perl/configure.t
ned21: Is that actually possible with a released version of CAF? AFAIK, the current version of CAF::FileWriter will always return undef on close() when running in NoAction mode, which means the unit tests will not trigger the a run of the command following a change.

ncm-sysctl/src/main/perl/sysctl.pm
ned21: I have tightened the schema validation for the config file to disallow / except when the name starts with a /

ncm-sysctl/src/test/perl/configure.t
stdweird: current build tools can do this, but you hva eto set it explicitly:
https://github.com/quattor/maven-tools/blob/master/build-scripts/src/main/perl/Test/Quattor.pm#L566

ned21:
@piojo - addressed both points

jouvin:
Any good reason to postpone to 15.4. From the comments, I don't see any negative comment and it has carefully reviewed by @stdweird and @piojo

stdweird:
it was opened after `14.12` window closed, but i'll leave it up to @jrha

jrha:
It's a new feature, I'd rather not.

PR #411, opened 2015-01-20
Required unit-test fixes for EL7 (and root)

stdweird:
These changes are required to pass the unittests on EL7 (some of them are required when running the unittests as root).
Part of the effort of https://github.com/quattor/release/pull/72


stdweird:
hmmm, the failure is because of https://github.com/quattor/CAF/pull/80; but i have no clue why the code path that triggers the error is used on jenkins (test is working on EL7 and fedora20)

ncm-icinga/src/test/perl/test-mkdirs.t
```diff
@@ -6,25 +6,30 @@ use Test::Quattor;
...
-	 check_result_path => DIR,
-	};
-
+$cmp->{ERROR} = 0;
 
+$cmp->make_dirs({check_result_path => $DIR});
 
-$cmp->{ERROR} = 0;
+ok(-d $DIR, "Directory created");
+is($cmp->{ERROR}, 0, "No realistic errors when creating the directory");
```
jouvin: What does mean the error message?

ncm-postfix/src/main/perl/postfix.pm
```diff
@@ -55,13 +55,15 @@ sub handle_config_file
     my ($self, $tree, $files) = @_;
     my $fh = CAF::FileWriter->new($files->{file}, log => $self);
     printf $fh "# File generated by %s. Do not edit\n", __PACKAGE__;
+    my $ok=1;
     if (!$self->template()->process($files->{template}, $tree, $fh)) {
         $self->error("Unable to process template for $files->{file}: ",
                  $self->template()->error());
         $fh->cancel();
-        return;
+        $ok = undef;
```
jouvin: Why returning `undef` rather than 0?

ncm-spma/src/test/perl/yum-cleanup-repos.t
```diff
@@ -82,13 +82,22 @@ In this case, the error must be reported
...
+  diag("Bad test for root user. Trying to set immutable bit.");
+  system("chattr +i $REPO_DIR/*repo");
+}
 is($cmp->cleanup_old_repos($REPO_DIR, $repos), 0,
    "Error reported when an outdated repo cannot be removed");
 is($cmp->{ERROR}, 2, "Error in unlink is reported");
 
 # Restore permissions on the repository for future executions.
-chmod(0700, $REPO_DIR);
+if ($>) {
```
jouvin: Using `$EUID` would be more readable than `$>`

ncm-spma/src/test/perl/yum-cleanup-repos.t
```diff
@@ -82,13 +82,22 @@ In this case, the error must be reported
 =cut
 
 initialize_repos();
-chmod(0500, $REPO_DIR);
+if ($>) {
```
jouvin: Using `$EUID` would be more readable than `$>`

ncm-spma/src/test/perl/yum-cleanup-repos.t
```diff
@@ -82,13 +82,22 @@ In this case, the error must be reported
...
+  chmod(0500, $REPO_DIR);
+} else {
+  diag("Bad test for root user. Trying to set immutable bit.");
+  system("chattr +i $REPO_DIR/*repo");
+}
 is($cmp->cleanup_old_repos($REPO_DIR, $repos), 0,
```
jouvin: Is there no risk to let repo files with wrong permissions or immuatable bit set in case of error or CTRL/C?

ncm-postfix/src/main/perl/postfix.pm
stdweird: i don't want to touch any behaviour like changing the return value on failure .

ncm-spma/src/test/perl/yum-cleanup-repos.t
stdweird: yes, there is. `mvn clean` will fail. the alternative here would be to not run this test when running the tests as root. (would adding a message how to clean up be sufficient?)

ned21:
Why are the unit tests being run as root? That has the possibility of screwing up the test system if someone accidentally runs a unit test that fails to observe the NoAction flag (and I came across one of these recently - possibly ncm-spma). IMHO the tests should fail if run as root.

stdweird:
@ned21 https://github.com/quattor/release/pull/72 currenlty runs as root, and for now, i'm fine with it trashing my VMs that way. (although switching to sudo for yum parts should be easy). but i don't see why the tests should fail as root (esp since all our code runs as root). 
to be clear, i personally never run them as root when developing, also jenkins does not run the tests as root.

jrha:
retest this please

PR #417, opened 2015-02-10
ncm-metaconfig: service perfsonar: Config::General does not respect order

stdweird:


stdweird:
@jrha nt the cleanest solution ever, but config general is not the prettiest thing ever either :smile:

PR #420, opened 2015-02-18
ncm-network: allow for bonding with lacp while keeping validation

wdpypere:
When bonding with LACP is configured (mode=4), updelay and primary are not required but lacp_rate is.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -36,12 +36,13 @@ type structure_bonding_options = {
     "primary" ? string with exists("/system/network/interfaces/" + SELF)
     "lacp_rate" ? long(0..1)
 } with {
-    if ( SELF['mode'] != 4 ) {
+    if ( SELF['mode'] == 1 || SELF['mode'] == 5 || SELF['mode'] == 6 ) {
         if ( ! exists(SELF["primary"]) ) {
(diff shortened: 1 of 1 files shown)


stdweird:
@wdpypere set milestaone and add `ncm-network` prefix to the title

jrha:
LGTM, could `mode` be restricted? IIRC only modes 1..6 are valid.

wdpypere:
@jrha it is updated.

Pansanel:
Are you sure that lacp_rate is required by mode=4? We are using this mode with only miimon set and it works fine. According to https://www.centos.org/docs/5/html/5.1/Deployment_Guide/s3-modules-bonding-directives.html, lacp_rate has a default value (0).

wdpypere:
@Pansanel You are correct, I'll update the pull request.

Pansanel:
As you can read on:
https://www.kernel.org/doc/Documentation/networking/bonding.txt
primary option is only valid for active-backup mode (active-backup(1), balance-tlb (5) and balance-alb (6)). It has not to be required by other mode, and we could also check that primary is not set with other mode than 1, 5 or 6.

wdpypere:
@Pansanel can you check again? I made primary mandatory for mode 1, 5 and 6, and made sure it is not configured for the other modes. 
I also removed the updelay check as it has a default and is only needed with miimon.

Pansanel:
It is ok for me.

PR #422, opened 2015-02-19
ncm-accounts: check for duplicate uids or gids in schema

kwaegema:


ncm-accounts/src/main/pan/components/accounts/schema.pan
```diff
@@ -6,6 +6,22 @@ declaration template components/accounts/schema;
 
 include { 'quattor/schema' };
 
+function has_unique_attr = {
+    values = ARGV[0];
+    attr = ARGV[1];
+    cattrs = list();
```
ned21: I am not sure about the name - is the c prefix for seen? If so I would prefer to spell it out -  "seen_attrs". I think it would also be more efficient to store this as an nlist Then the check below is just "if exists(seen_attrs[el[attr]])" instead of a search through a list.

ncm-accounts/src/main/pan/components/accounts/schema.pan
```diff
@@ -6,6 +6,22 @@ declaration template components/accounts/schema;
...
+
+    foreach(ni;el;values) {
+        if(index(el[attr], cattrs) != -1) {
+            error(format("Duplicate attr %s : %s ", attr, el[attr]));
+            return(false);
```
ned21: Remove this line - it's redundant since error() already stops processing.

ncm-accounts/src/main/pan/components/accounts/schema.pan
```diff
@@ -6,6 +6,22 @@ declaration template components/accounts/schema;
...
+        } else {
+            append(cattrs, el[attr]);
+        };
+    };
+    return(true);
```
ned21: IIRC, it's more efficient to just say "true" here and not explicitly call return.

kwaegema:
remarks processed

ned21:
Thanks for fixing that. The second commit message isn't very informative for anyone just reading the git commit history since there's no way for them to know what/who's remarks were processed. I think for a simple change like this the best way to keep the commit history easy to follow is just to squash the second commit into the first, if you don't mind doing that? I've also just noticed that the first commit message also has a typo in it -- uids *of* gids instead of *or* -- so would you mind fixing that up at the same time? I think we might have a convention that commit messages should start the first line with the name of the component too as lots of them seem to, but I am not sure if that is policy or not.

kwaegema:
@ned21 done!

jrha:
AFAIK there is no policy that commit messages start with the name of the component, it is very helpful when reviewing PRs that touch more than one component though.

It is policy that issues and PRs against a single component start with the name of that component though.

jrha:
How about a unit test for this important new functionality?

PR #424, opened 2015-02-25
ncm-systemd: initial support for systemd unit control

stdweird:
Provide inital framework to handle systemd units.


stdweird:
It's a lot of code, and it can't do much yet (but it's not so trivial as chkconfig :smile: )

Known issues
 * no target handling
 * too much verbose logging
 * not used to run anything :smile:
 * no ncm-chkconfig drop in code

Feedback is welcome, no need to review the `cmddata` modules :smile:

stdweird:
@ned21 don't review per commit. There is some major rework done at the end, i'm open to suggestion how to rewrite the git history.

ncm-systemd/src/main/perl/Systemd/Service.pm
```diff
@@ -0,0 +1,519 @@
...
+# TODO protecting network assumes ncm-network is being used
+# TODO shouldn't these services be "always on"?
+Readonly::Hash my %DEFAULT_PROTECTED_SERVICES => (
+    network => 1,
+    messagebus => 1,
+    haldaemon => 1,
+    sshd => 1,
+);
+
+Readonly my $BASE => "/software/components/systemd";
```
ned21: Use $self->prefix instead as this then allows module "aliasing"

jouvin:
Could we merge this one (with or without @ned21 suggested change) to allow give it some early exposure for those starting to look at EL7. I think this is inline with Grenoble discussion. As it is still not use in any part of the template library, it is harmless if the componenent is not yet rock solid. We may add a disclaimer in the doc that this is an early/beta version and that many things are subject to change in the next release, including the schema... What do you think?

stdweird:
@jouvin i'd like things reviewed before merged in. i also found that the `possible_missing` work is not sufficient, so expect a few more commits. people who want to work with ncm-systemd can build there own package (`15.4` is only 1.5 months away btw).

jouvin:
Ok, as you prefer. As I said, personnally I don't mind putting a beta version of a new component (not an existing one!) in a release if we manage to put a disclaimer somewhere...

ncm-systemd/src/main/perl/Systemd/Service.pm
stdweird: @ned21 what do you mean with aliasing?

ned21: See https://github.com/quattor/ncm-ncd/pull/11

stdweird: hmm, it makes things complicated on my end as `Systemd::Service` is not a component. i'll have to think about this.

the current usage i have in mind is to use unmodified `ncm-chkconfig` profiles under EL7 by including the following template in EL7:
https://github.com/stdweird/configuration-modules-core/blob/systemd/ncm-systemd/src/main/pan/components/systemd/legacy/chkconfig.pan

it sets an alias on `ncm-chkconfig` for a partial `ncm-systemd`, namely the part of `ncm-systemd` that deals with services. (`ncm-systemd` will also be able to do e.g. unit configuration, the equivalent of current `ncm-sysconfig`).

stdweird:
current state: we run this in production as pure replacement for ncm-chkconfig using nothing more then 
```
include 'components/systemd/legacy/chkconfig';
```
in the 
`config/core/base` for centos7 os

and a modification to the `structure_component` type in the the core `quattor/schema`
```
    "ncm-module"        ? string with match(SELF, '^([a-zA-Z_]\w*(?:::[a-zA-Z_]\w*)*)$')
```

everything else is unmodified wrt templates and it seems to work quite ok.

stdweird:
see https://github.com/quattor/template-library-core/pull/63 for the schema change

jrha:
I really want this in 15.4, I'll try and find some time to sit down and go through the code, but the more reviewers the better!

ncm-systemd/src/main/perl/Systemd/Service.pm
```diff
@@ -0,0 +1,529 @@
...
+                       join(',',@limit_units));
+        $possible_missing = $self->{unit}->possible_missing($relevant_units);
+    }
+
+    # A sysv service that is not listed in chkconfig --list
+    #   you can run systemctl enable/disable on it (it gets redirected to chkconfig)
+    #   they do show up in list-units --all
+    #     even when only chkconfig --add is used
+    #   systemctl mask removes it from the output of chkconfig --list
+    #   systemctl umask restores it to last known state
```
ned21: umask => unmask

ncm-systemd/src/main/perl/Systemd/Service/Chkconfig.pm
```diff
@@ -0,0 +1,504 @@
...
+
+sub current_units
+{
+    my $self = shift;
+
+    my %current;
+
+    my $data = CAF::Process->new(
+        [$CHKCONFIG, '--list'],
+        log => $self,
```
ned21: This needs a keepstate=>1 to work under NoAction mode.

ncm-systemd/src/main/perl/Systemd/Systemctl.pm
```diff
@@ -0,0 +1,367 @@
...
+
+Returns undef on failure.
+
+=cut
+
+sub systemctl_show
+{
+    my ($logger, $unit) = @_;
+    my $proc = CAF::Process->new([$SYSTEMCTL, "--no-pager", "--all", "show"],
+                                  log => $logger,
```
ned21: Requires keepstate=>1 to work correctly in NoAction mode.

ncm-systemd/src/main/perl/Systemd/Systemctl.pm
```diff
@@ -0,0 +1,367 @@
...
+=cut
+
+sub systemctl_list_deps
+{
+    my ($logger, $unit, $reverse) = @_;
+
+    # no --all !
+    my $proc = CAF::Process->new(
+        [$SYSTEMCTL, '--no-pager', '--no-legend', '--full', '--plain', 'list-dependencies'],
+        log => $logger,
```
ned21: Add keepstate

ncm-systemd/src/main/perl/Systemd/Systemctl.pm
```diff
@@ -0,0 +1,367 @@
...
+Output that does not match the regexp is skipped, if the regexp matches but
+there is no C<name> value in the named group, it is also skipped and
+logged as error.
+
+=cut
+
+sub systemctl_list
+{
+    my ($logger, $spec, $regexp, $type) = @_;
+    my $proc = CAF::Process->new(
```
ned21: Add keepstate

ned21:
The code looks OK although I am not sure it will give very useful output in NoAction mode without the "list" commands to provide data to work with.  I don't know anything about systemd to conclude whether this is the "right" way to implement this but I guess the schema and use of the component was reviewed at the workshop?

stdweird:
@ned21 
wrt NoAction, without any list data, it's pretty much useless.
the component was presented at the workshop, the schema briefly reviewed. main target use for now is to use the configuration from existing `/software/components/chkconfig`. i guess when people will actually use it, we'll see how it goes.
what do you want to know about systemd wrt the "right" way? main i can provide some further info.

jouvin:
I'm trying to use `ncm-systemd`but when I include the chkconfig compatibility, I get the following error:

```
     [panc] validation error [/scratch/jouvin/quattor/cdb/cfg/clusters/lal/service-machines-test/profiles/tracsrv2.lal.in2p3.fr.pan]
     [panc] record definition contains unexpected field(s): ncm-module 
     [panc] element path: '/software/components/chkconfig'
     [panc] type: 'component_chkconfig_type' [/scratch/jouvin/quattor/cdb/cfg/quattor/15.4.0-dev/components/chkconfig/schema.pan:67.33-71.1]
     [panc] path '/software/components/chkconfig' bound to type component_chkconfig_type in [/scratch/jouvin/quattor/cdb/cfg/quattor/15.4.0-dev/components/chkconfig/schema.pan:73.41-73.64]
 ```

Do I need an updated schema?

stdweird:
@jouvin yes, you need https://github.com/quattor/template-library-core/blob/master/quattor/types/component.pan#L22

jouvin:
Works like a charm :) Thanks for this very useful contribution. I'll add the suggested include as part of the base EL7 config in the template library: this will make the transition very easy.

As far as I'm concerned, I think it is ready for merging.

jrha:
Agreed.

jouvin:
Were @ned21's comments about `keepstat`addressed (https://github.com/quattor/configuration-modules-core/pull/424#discussion-diff-28652297)? If yes, I am ready to merge it. We'll open issues when they arise...

stdweird:
@jouvin yes, the second last commit should do that

PR #433, opened 2015-03-04
Fix pod-check unittest to include the pod files

stdweird:


stdweird:
@wdpypere voila!

wdpypere:
@stdweird nice! Thank you! :thumbsup:

PR #448, opened 2015-03-16
ncm-icinga: contact group definition is not mandatory for contacts. 

wdpypere:
They can also be defined in the contact group definition. Both are valid.


jrha:
Trivial, thanks!

PR #451, opened 2015-03-25
ncm-spma: Restrict repository name

### LINKED ISSUE #409: ncm-spma: repository name can contain invalid characters
The validation for repository names is currently:
```pan
"name" ? string # "Repository name"
```
This is not strict enough as the name is used as the repository id for YUM configurations, it seems (from experimentation) that YUM will accept `[A-Za-z0-9_-]`, I haven't checked how any of this maps onto IPS for Solaris however.

jrha:
Fixes #409.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/schema.pan
+++ b/ncm-spma/src/main/pan/components/spma/schema.pan
@@ -42,7 +42,7 @@ type SOFTWARE_REPOSITORY = {
     "gpgcheck" : boolean = false
     "excludepkgs" ? string[]
     "includepkgs" ? string[]
-    "name" ? string  # "Repository name"
+    "name" ? string with match(SELF, '^[\w-]+$') # "Repository name"
     "owner" ? string  # "Contact person (email)"
     "priority" ? long(1..99)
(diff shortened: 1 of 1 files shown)


ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -42,7 +42,7 @@ type SOFTWARE_REPOSITORY = {
     "gpgcheck" : boolean = false
     "excludepkgs" ? string[]
     "includepkgs" ? string[]
-    "name" ? string  # "Repository name"
+    "name" ? string with match(SELF, '^[A-Za-z0-9_-]+$') # "Repository name"
```
stdweird: use `'^\S+$` to exclude withespace; or `^[\w-]+$` for current regexp (but no dots either?)

jrha:
See my comments on #409: dots may work but I'm going by the scant documentation I managed to find.

stdweird:
@jrha LGTM

jrha:
If happy can someone merge this so that I don't have to merge my own change (which would make me sad).

PR #453, opened 2015-03-25
ncm-nfs: Support PanFS mounts

### LINKED ISSUE #452: ncm-nfs: schema too restrictive to use for panfs
PanFS mounts are very similar to nfs mounts, so we should extend the component to support them.
As requested by @drossy

jrha:
Fixes #452 as requested by @drossy

### DIFF
--- a/ncm-nfs/src/main/pan/components/nfs/schema.pan
+++ b/ncm-nfs/src/main/pan/components/nfs/schema.pan
@@ -15,7 +15,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device'     : string
     'mountpoint' : string
-    'fstype'     : string with match(SELF, '^nfs|panfs|none$')
+    'fstype'     : string with match(SELF, '^(nfs|panfs|none)$')
     'options'    ? string
     'freq'       ? long(0..)
     'passno'     ? long(0..)
--- a/ncm-nfs/src/main/pan/components/nfs/schema.pan
+++ b/ncm-nfs/src/main/pan/components/nfs/schema.pan
@@ -15,7 +15,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device'     : string
     'mountpoint' : string
-    'fstype'     : string with match(SELF, '^(nfs|panfs|none)$')
+    'fstype'     : string with match(SELF, '^(nfs4?|panfs|none)$')
     'options'    ? string
     'freq'       ? long(0..)
     'passno'     ? long(0..)
--- a/ncm-nfs/src/main/pan/components/nfs/schema.pan
+++ b/ncm-nfs/src/main/pan/components/nfs/schema.pan
@@ -15,7 +15,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device'     : string
     'mountpoint' : string
-    'fstype'     : string with match(SELF, '^nfs|none$')
+    'fstype'     : string with match(SELF, '^nfs|panfs|none$')
     'options'    ? string
     'freq'       ? long(0..)
     'passno'     ? long(0..)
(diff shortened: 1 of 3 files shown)


jrha:
Needs tests and an example in the documentation

stdweird:
there is code in the component that checks for nfs and should be modified like `($fstype =~ /^nfs/)` (but the unittests will reveal that :smile: )

jrha:
Hmm, there are only 00-load and pod-syntax tests present right now :sob:

drossy:
Looks good to me, is it worth adding a comment to the documentation saying that extending the regexp is the way to go for others in a similar situation?

jrha:
Good point!

ncm-nfs/src/main/pan/components/nfs/schema.pan
```diff
@@ -15,7 +15,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device'     : string
     'mountpoint' : string
-    'fstype'     : string with match(SELF, '^nfs|none$')
+    'fstype'     : string with match(SELF, '^nfs|panfs|none$')
```
stdweird: you are aware that the string `"nfspanfsnone"` matches this regexp, right :smile: use
```pan
match(SELF, '^(nfs|panfs|none)$')
```

(pan really needs a `in` test)

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -114,7 +114,7 @@ sub Configure($$@) {
       $freq = 0 unless defined($freq);
       $passno = 0 unless defined($passno);
       
-      if ( ($fstype =~ /^nfs/) || (($fstype eq 'none') && ($opt eq 'bind')) ) {
+      if ( ($fstype =~ /^(pa)?nfs/) || (($fstype eq 'none') && ($opt eq 'bind')) ) {
```
stdweird: use `/^(pa)?nfs$/` ?

stdweird:
@jrha minor regexp remarks (i prefer strict regexps)

ncm-nfs/src/main/perl/nfs.pm
jrha: Pffft, blame who ever wrote it first :stuck_out_tongue:

ncm-nfs/src/main/pan/components/nfs/schema.pan
jrha: Ugh, well so does `nfsbanana`

jrha: But then nfsbanana and elephantnone would have matched the old regexp.

stdweird: i didn't mean to imply that you introduced the support for nfsbanana, but might as well fix it :smile:

jrha: :smiling_imp: understood!

jrha:
Is something wrong with Jenkins?

stdweird:
retest this please

stdweird:
@jrha disk full, should be ok again

stdweird:
test this please

jrha:
Yeah it looks sick. :hankey:

jrha:
Yay! :smile:

drossy:
nfs4 is a valid filesystem type that would have been accepted by the old regexp (and I think I saw a mention of it somewhere in the quattor code but can't find it right now)

stdweird:
@drossy true, then it should be allowed in the schema and the perl regexp like
```pan
match(SELF, '^(nfs4?|panfs|none)$')
```
and 
```perl
$fstype =~ /^(nfs4?|panfs)$/
```

jrha:
@stdweird happy?

stdweird:
@jrha LGTM

PR #456, opened 2015-03-25
ncm-metaconfig: remove daemon property

### LINKED ISSUE #431: ncm-metaconfig: daemon field is deprecated and should be removed
`daemon` was intended as a list of services to restart when a file was changed, this proved to not be flexible enough and `daemons` was implemented to replace it by providing a `dict()` of pairs of services and actions (`restart`, `reload`, `stop_sleep_start`).

We should remove `daemon` in a future release.

To preserve exisiting behaviour for templates using `daemon`, migrate the list of services to `daemons` with the `restart` action.

jrha:
The daemon property was a good start, but is now deprecated and replaced with the daemons structure.
This removes the deprecated property, the code associated with it and anything that was testing it.

Fixes #431

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -15,7 +15,6 @@ type ${project.artifactId}_config =  {
      'mode' : long = 0644
      'owner' : string = 'root'
      'group' : string = 'root'
-     'daemon' ? string[] with { deprecated(0, "daemon property has been deprecated, daemons should be used instead"); true; }
      'daemons' ? caf_service_action{}
      'module' : string
      'backup' ? string


jrha:
NOT FINISHED YET

jrha:
I think this is done now, but at least @stdweird should look carefully at my slapdash coding.

ncm-metaconfig/src/main/metaconfig/ganesha/pan/config.pan
```diff
@@ -9,7 +9,7 @@ include 'metaconfig/ganesha/schema';
 bind "/software/components/metaconfig/services/{/etc/ganesha/ganesha.nfsd.conf}/contents" = ganesha_config;
 
 prefix "/software/components/metaconfig/services/{/etc/ganesha/ganesha.nfsd.conf}";
-"daemon" = {if (CTDB_MANAGES_GANESHA) { null } else { list(GANESHA_SERVICE) }};
+"daemon" = {if (CTDB_MANAGES_GANESHA) { null } else { dict(GANESHA_SERVICE, "restart") }};
```
stdweird: `daemons`

ncm-metaconfig/src/main/metaconfig/perfsonar/pan/bwctl/config.pan
```diff
@@ -30,11 +30,11 @@ prefix "/software/components/metaconfig/services/{/var/lib/bwctl/.bwctlrc}";
     l = value("/software/components/metaconfig/services/{/var/lib/bwctl/.bwctlrc}");
     l["owner"] = "perfsonar";
     l["group"] = "perfsonar";
-    l["daemon"] = list("perfsonarbuoy_bw_master");
+    l["daemon"] = dict("perfsonarbuoy_bw_master", "restart");
```
stdweird: `daemons`

ncm-metaconfig/src/main/metaconfig/perfsonar/tests/profiles/bwctl.pan
```diff
@@ -20,7 +20,7 @@ prefix "/software/components/metaconfig/services/{/var/lib/bwctl/.bwctlrc}/conte
     l = value("/software/components/metaconfig/services/{/var/lib/bwctl/.bwctlrc}");
     l["owner"] = "perfsonar";
     l["group"] = "perfsonar";
-    l["daemon"] = list("perfsonarbuoy_bw_master");
+    l["daemon"] = dict("perfsonarbuoy_bw_master", "restart");
```
stdweird: `daemons`

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -15,7 +15,6 @@ type ${project.artifactId}_config =  {
      'mode' : long = 0644
      'owner' : string = 'root'
      'group' : string = 'root'
-     'daemon' ? string[]
```
stdweird: how nicely backwards incompatible.
for migration purposes you could use 
```
'daemon' ? string with error("use daemons")
```
did we agree on backwards incompatible changes for this issue?

ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -44,16 +42,6 @@ sub prepare_action
         }
     }
 
-    if ($srv->{daemon}) {
```
stdweird: i'm sorry, but this is even more of a stretch. this code has to stay in to deal with old templates. we can deprecated it (or even throw a warning/error here, but not just rip it out).

stdweird:
@jrha can you do a `grep -r daemon | grep -v daemons` in the metaconfig subdirectoyr. should have no hits
but i don't agree with the aggressive deprecation policy (esp not in the component). for the schema it's ok i guess, but the code needs proper deprecation path and new issue to says remove the code (and even then, it should at least throw an error).

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
jrha: My recollection is that we agreed to remove the functionality before it started to be used at many sites.

jrha:
If you are worried about the deprecation path we can move this to 15.6 and just modify the schema for 15.4.

ncm-metaconfig/src/main/perl/metaconfig.pm
jrha: Fair enough, but how would this code get exposed to old templates? If the component is update the profile must be rebuilt and the schema validation would throw an error.

ncm-metaconfig/src/main/metaconfig/perfsonar/pan/buoy/mesh.pan
```diff
@@ -6,9 +6,9 @@ bind "/software/components/metaconfig/services/{/opt/perfsonar_ps/perfsonarbuoy_
 
 prefix "/software/components/metaconfig/services/{/opt/perfsonar_ps/perfsonarbuoy_ma/etc/owmesh.conf}";
 "module" = "perfsonar/owmesh";
-"daemon" = list(
-    "perfsonarbuoy_bw_master", 
-    "perfsonarbuoy_bw_collector",
-    "perfsonarbuoy_owp_master", 
-    "perfsonarbuoy_owp_collector"
-    );
+"daemon" = dict(
```
stdweird: `daemons`

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
stdweird: ok, but afaik MS were also using metaconfig already so they probably also have an opinion
@ned21 any takes how to deprecate this?

stdweird:
@jrha the deprecation should at least log an error in the component for a a couple of releases (if not forever).

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: ah yes, except in our case, we don't releases for lots of things, esp at this moment. after a while it converges, but not now :stuck_out_tongue: 
if we really start to rely on the coupling, we need to enforce this (or at least make it possible to enforce it).

jrha: I'm not sure I understand...
Ae you saying that you update the component RPMs without updating the schema?

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
jrha: I was under the impression from the workshop that it is already deprecated (hence the issue) and  why I'm stripping it out now. If i misunderstood then we should push it back a release or two, but I really don't want anyone else to start the deprecated structure in the mean time.

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: yes, it has happened before and might/will happen again (we have quiet a few self-packaged parts of quattor with master branch + PRs merged in (or branches under development), and not always sync back the schema of each new package with the actual schema).
it's not good practise, but quattor does not protect against this. enforcing the schema with the component should be possible (and is probably quite easy to do).
it's actually worse then that considering all non-locked quattor parts like CAF and CCM (at least we don't lock them). our packages have no version dependencies, and there are quite some things that won't work without aligned versions. (e.g. imagine your repo sync failed and you skipped CCM or CAF; and you deploy 15.2).

anyway, deprecation should done with care :smile:

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
stdweird: i'm fine with throwing it out of the schema, but a clearer error message is nice (like it is now).
i'm also fine with throwing the functionality out of the code, but there should always be an error logged. ideally, wrt the code, functionality should stay in there for another release (but it can already log warning or error).

ncm-metaconfig/src/main/perl/metaconfig.pm
jrha: :fearful:

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
jrha: No that's cool, I didn't realise anyone did such strange things with components :scream:, I'll bump this to 14.6 and open a new PR for the schema change.

ned21:
I am with @stdweird - this should be done over a few releases. First release change the schema to emit a warning if its used (and the code as well if you like).
```
  'daemon' ? string with deprecated(0, "Long explanatory message here")
```

should do it. Then target the schema and code clean up in the next release.

stdweird:
@ned21 ok to deprecate in 15.12?

jrha:
It was deprecated in #479 which was in 15.8, this removes it entirely.

stdweird:
@ned21 ok to remove it in 15.12?

ned21:
Yep

stdweird:
@jrha more greyed out buttons...
@ned21 can you merge this one too?

stdweird:
@jrha seems the rebase/out-of-date stuff doesn't allow me to merge this. i could e.g. press the green button in https://github.com/quattor/configuration-modules-core/pull/596

jrha:
I don't understand it, I can see:
> All checks have passed
> 1 successful check
> This branch is up-to-date with the base branch
> Merging can be performed automatically.

With a green button.

stdweird:
@jrha not for me
no_merge_pr

jrha:
What about now?

stdweird:
seems to work just fine :smile:

jrha:
Ok, it's related to the mandatory checks, I'll have to contact GitHub support :disappointed:

PR #464, opened 2015-03-31
ncm-metaconfig: add service carbon-relay-ng

stdweird:


jrha:
Looks fine to me (compared to example configs at https://github.com/graphite-ng/carbon-relay-ng)

PR #467, opened 2015-04-06
ncm-metaconfig: pass the contents as Element instance CAF::TextRender

stdweird:


stdweird:
requires https://github.com/quattor/CAF/pull/85 (and thus also https://github.com/quattor/CCM/pull/51)

stdweird:
Will be replace with CCM::TextRender implementation

PR #474, opened 2015-04-17
ncm-ssh: Add support for AuthorizedKeysCommand* options

drossy:
Add support for AuthorizedKeysCommand and AuthorizedKeysCommandRunAs to sshd, which I think are new options in openssh in RHEL6


jrha:
please test this

PR #478, opened 2015-04-27
ncm-ceph: Add `straw2` as possible bucket algorithm

gvasilak:
The original implementation of the `straw` algorithm is flawed as Sage Weil describes [here](http://comments.gmane.org/gmane.comp.file-systems.ceph.devel/22317)
With the release of Hammer, `straw2` is now available which fixes what was wrong with the old algorithm.
There is no reason for any new Ceph cluster running Hammer to use straw.
Ideally, straw2 would be automatically chosen as the default algorithm for a cluster running Hammer or later.

### DIFF
--- a/ncm-ceph/src/main/pan/components/ceph/schema.pan
+++ b/ncm-ceph/src/main/pan/components/ceph/schema.pan
@@ -245,7 +245,7 @@ type ceph_radosgwh = {
     arg = bucket algoritm
 }
 function is_ceph_crushmap_bucket_alg = {
-    if (!match(ARGV[0], '^(uniform|list|tree|straw)$')){
+    if (!match(ARGV[0], '^(uniform|list|tree|straw|straw2)$')){
         error(ARGV[0] +  'is not a valid bucket algorithm');
         return(false);
     };
--- a/ncm-ceph/src/main/pan/components/ceph/schema.pan
+++ b/ncm-ceph/src/main/pan/components/ceph/schema.pan
@@ -245,7 +245,7 @@ type ceph_radosgwh = {
     arg = bucket algoritm
 }
 function is_ceph_crushmap_bucket_alg = {
-    if (!match(ARGV[0], '^(uniform|list|tree|straw|straw2)$')){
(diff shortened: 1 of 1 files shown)


jrha:
Also change `defaultalg` in schema?

gvasilak:
`straw2` is only available starting with the *Hammer* release of Ceph.
If it were to be made the default any deployment using an older release will need to have `straw` explicitly specified.

This might not be such a bad thing. *Hammer* is the latest long-term stable/supported release of Ceph and `straw2` should be used always be used over `straw`, the exception being a cluster already significantly filled with data. Even then, the Ceph team are working on ensuring the absolutely minimal amount of data movement when switching to `straw2`.

kwaegema:
There is one exception: using CephFS :smile: 
Crush settings should be supported by the Ceph Clients, otherwise the clients can't even connect, and straw2 buckets will only be supported in the 4.1 kernel.
Of course it should become the default in most cases, using Hammer, but I would't just set it here, that would brake all CephFS clusters when upgrading ncm-ceph. (I'd rather work without a default value then, then the algorithm should be set in the profiles per cluster or so..)

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -245,7 +245,7 @@ type ceph_radosgwh = {
     arg = bucket algoritm
 }
 function is_ceph_crushmap_bucket_alg = {
-    if (!match(ARGV[0], '^(uniform|list|tree|straw)$')){
+    if (!match(ARGV[0], '^(uniform|list|tree|straw|straw2)$')){
```
stdweird: the regexp would better be `'^(uniform|list|tree|straw2?)$`

gvasilak:
@kwaegema you're right about the CephFS bit, didn't think of that. Perhaps it's too early to be changing defaults, let's get it supported first.

stdweird:
test this please

jrha:
Looks good, update the regex and it can go into this release.

gvasilak:
I've updated the regex to what @stdweird proposed.

Issue #250, opened 2014-07-16
ncm-mysql: user problem


stdweird:
Some issues with ncm-mysql
* `FLUSH PRIVILEGES` after any user/privileges modifications
* mysql doesn't like `\` characters in the passwords. maybe check this in the schema?
* it's quite a hassle to create an empty database with just a user that has `ALL PRIVILEGES` rights on a none existing database. current trickery involves setting `createDb` to `false` and using `/dev/null` as `initScript/file`.
* the code allows `ALL` (just a string) for rights, but the schema enforces lists. might want to add the default privileges as default in schema (` = list('ALL PRIVILEGES')`)

jrha:
Bump.

ned21:
Let's clear the milestone on anything where there's not a realistic expectation of making progress, and it's not a must have feature/bug fix. Milestones can always be added to PRs if someone suddenly finds a patch down the back of the sofa...

stdweird:
or close this issue as only #482 is left?

PR #484, opened 2015-04-27
ncm-mysql: Schema fixes for `component_mysql_db_user` type

jrha:
See #250 for background.

### DIFF
--- a/ncm-mysql/src/main/pan/components/mysql/schema.pan
+++ b/ncm-mysql/src/main/pan/components/mysql/schema.pan
@@ -36,7 +36,7 @@ function component_mysql_check_db_script = {
 
 type component_mysql_db_user = {
   'password'  : string with match(SELF, '^[^\\]+$')
-  'rights'    : string[]
+  'rights'    : string[] = list('ALL PRIVILEGES')
   'shortPwd'  : boolean = false
 };
 
--- a/ncm-mysql/src/main/pan/components/mysql/schema.pan
+++ b/ncm-mysql/src/main/pan/components/mysql/schema.pan
@@ -34,9 +34,11 @@ function component_mysql_check_db_script = {
   };
 };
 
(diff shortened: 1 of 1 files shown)


stdweird:
retest this please

ncm-mysql/src/main/pan/components/mysql/schema.pan
```diff
@@ -35,8 +35,8 @@ function component_mysql_check_db_script = {
 };
 
 type component_mysql_db_user = {
-  'password'  : string
-  'rights'    : string[]
+  'password'  : string with match(SELF, '^[^\\]+$')
+  'rights'    : string[] = list('ALL PRIVILEGES')
```
stdweird: is `ALL PRIVILIGES` really the default in mysql? it seems very very wrong (and maybe we should settle for something lower; like readonly?)

stdweird: also something restricted like
```pan
type component_mysql_user_right = string with match(SELF, '^(ALL( PRIVILEGES)?|ALTER( ROUTINE)?|CREATE( (ROUTINE|TEMPORARY TABLES|USER|VIEW))?|DELETE|DROP|EVENT|EXECUTE|FILE|GRANT OPTION|INDEX|INSERT|LOCK TABLES|PROCESS|REFERENCES|RELOAD|REPLICATION (CLIENT|SLAVE)|SELECT|SHOW (DATABASES|VIEW)|SHUTDOWN|SUPER|TRIGGER|UPDATE|USAGE)$');

    'rights' : component_mysql_user_right[] = list('SELECT')
...
```
(completely untested)

jrha: I don't know, it was your suggestion...

jrha: `SELECT` might be a better default?

jrha:
I checked your code, it seemed fine so I've added it :grin:

stdweird:
@jrha thanks for the extended testing :grinning:

jrha:
@jouvin, could you take a look at this and merge if happy?

PR #496, opened 2015-05-07
ncm-spma: allow '.' in repository names

### LINKED ISSUE #495: ncm-spma 15.4.0-rc1: repository name validation too restrictive
In 15.4.0, there is a validation rule for repository names. Unfortunatly, it lacks `.` in the list of allowed characters...

This is a blocking issue for 15.4.0 (https://github.com/quattor/release/issues/85).

jouvin:
Fixes #495.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/schema.pan
+++ b/ncm-spma/src/main/pan/components/spma/schema.pan
@@ -42,7 +42,7 @@ type SOFTWARE_REPOSITORY = {
     "gpgcheck" : boolean = false
     "excludepkgs" ? string[]
     "includepkgs" ? string[]
-    "name" ? string with match(SELF, '^[\w-]+$') # "Repository name"
+    "name" ? string with match(SELF, '^[\w-.]+$') # "Repository name"
     "owner" ? string  # "Contact person (email)"
     "priority" ? long(1..99)
     "protocols" ? SOFTWARE_REPOSITORY_PROTOCOL []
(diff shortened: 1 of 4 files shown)


stdweird:
can you add a unittest for the schema?
make a template in `test/resources` that has a `include 'component/spma/schema` or `config`, add required values and make a new perl `schema.t` which has 

```perl
use Test::Simple;
use Test::Quattor qw(name_of_template_with_schema);
ok(1, "Would have failed earlier if schema was invalid");
done_testing;
```

jouvin:
Sure! I didn't think about this way of doing it...

jouvin:
@stdweird unfortunately this is not so simple... I tried your suggestion but in this case you need to include the component schema to have the binding and this doesn't work because a component schema includes `quattor/schema` (which can be found if you have a clone of template-library-core at the same level as the clone of configuration-module-core) which includes `monitoring/lemon/schema` which is in another repo...
I don't see a solution that is not an overkill for this particular case, even though it would have been desirable to have a unit test...

stdweird:
you can use `types/component` now, i refactored the `quattor/schema` to get around the missing monitoring issue for exactly this reason. (https://github.com/quattor/template-library-core/pull/58)

jouvin:
Right after my previous option, I found another solution... add a dummy `quattor/schema`. The advantage is that there is no dependency over `template-library-core` as here we just need `structure_component` to be an existing type but we don't care about its definition...

stdweird:
well, it defeat a bit the purpose of testing the templates that are shipped, but it's certainly an improvement.

stdweird:
LGTM

jouvin:
@stdweird no, I don't think it defeat anything. What we want to test here is only what is provided in the component schema, we don't care about other part of the schema (in fact, if the test fails because of something wrong in another part of the quattor or structure_component schema, this will not be a unit test anymore!).

PR #497, opened 2015-05-08
ncm-nrpe: Fix documentation issue

### LINKED ISSUE #491: ncm-nrpe: Documentation is wrong
Hi,

There's some problems with the current documentation regarding the nrpe module:
* highlighted paths are not accurate when the path comprises an underscore (e.g. log_facility)
* Contrary to what the documentation says (all fields are mandatory, unless...), all fields are optional except ```command``` and ```allowed_hosts```
* all fields should go into ```/software/components/nrpe/options``` and not ```/software/components/nrpe``` (though I don't see the point of it)

Julien

jrha:
Fixes #491.

### DIFF
--- a/ncm-nrpe/src/main/perl/nrpe.pod
+++ b/ncm-nrpe/src/main/perl/nrpe.pod
@@ -11,7 +11,7 @@ way.
 
 =head1 COMPONENT STRUCTURE
 
-All fields are mandatory, unless it is explicitly stated:
+All fields are required (but most have sensible defaults unless otherwise stated).
 
 =over 4
 
@@ -44,6 +44,8 @@ Group the daemon will run as. For instance, 'nagios'.
 =item * C<< /software/components/nrpe/allowed_hosts : type_hostname[] >>
 
 List of hosts allowed to order the NRPE daemon to run commands.
+
+Must be specified, no default is provided.
 
 =item * C<< /software/components/nrpe/dont_blame_nrpe : boolean >>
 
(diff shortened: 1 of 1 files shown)


ncm-nrpe/src/main/perl/nrpe.pod
```diff
@@ -11,80 +11,84 @@ way.
 
 =head1 COMPONENT STRUCTURE
 
-All fields are mandatory, unless it is explicitly stated:
+All fields are required (but most have sensible defaults) unless otherwise stated:
```
stdweird: should be `All fields are required (but most have sensible defaults unless otherwise stated)`? (because you add phrases like `Must be specified, no default is provided.`).

ncm-nrpe/src/main/perl/nrpe.pod
```diff
@@ -11,80 +11,84 @@ way.
...
 
 Group the daemon will run as. For instance, 'nagios'.
 
-=item * /software/components/nrpe/allowed_hosts : type_hostname[]
+=item * C<< /software/components/nrpe/options/allowed_hosts : type_hostname[] >>
 
 List of hosts allowed to order the NRPE daemon to run commands.
 
-=item * /software/components/nrpe/dont_blame_nrpe : boolean
+Must be specified, no default is provided.
```
stdweird: i would put the options that have no default and are required on top of the pod (because those must be configured)

stdweird:
LGTM

Issue #380, opened 2014-11-19
ncm-symlink: replace dirempty does not work as intended


stdweird:
`dirempty` behaviour uses `stat->size` to "determine" if a directory is empty or not. but that only gives the directory inode size, which is never 0 and highly FS-dependent (and probably can't shrink in certain filesystems).

``` bash
mkdir new && cd new && perl -e 'use File::stat;print $ENV{PWD}," ",stat($ENV{PWD})->size, "\n";'
```

jrha:
Just tried making empty directories on various file-systems on a cross section of our nodes, getting the size from `stat` as follows:
```bash
df --output=fstype .; mkdir empty-dir; stat -c %s empty-dir; rmdir empty-dir
```

Results:

Filesystem | empty-dir size in bytes
---|---
ext2 | 1024
ext3 | 4096
ext4 | 4096
tmpfs | 40
xfs | 6
zfs | 2

So the logic in the code looks like it should never work.

jouvin:
https://github.com/quattor/configuration-modules-core/issues/683

jouvin:
To be done as part of the [full rewrite](https://github.com/quattor/configuration-modules-core/issues/683) planned and the new `ncm-path` discussed at Annecy workshop (March 17).

PR #500, opened 2015-05-08
ncm-symlinks: documentation reformatting and clarification

### LINKED ISSUE #381: ncm-symlink: clarify intended behaviour of replace
the current behaviour of `ncm-symlink` can't really be derived form the pod
the schema allows all suboptions, the code however doesn't; so ideally the schema is restricted or at least the (intended) behaviour made clear in the pod (and some examples are welcome too; e.g. what is the exact behaviour of `replace = nlist('all', 'yes')`?).

because of the power of `ncm-symlink`, this has to be clarified and then the code checked with the intended behaviour.
in particular, the "eastereggs" `none=no` and `all=no` should be documented/fixed

jouvin:
Fixes #381.

### DIFF
--- a/ncm-symlink/src/main/perl/symlink.pod
+++ b/ncm-symlink/src/main/perl/symlink.pod
@@ -15,11 +15,8 @@ symlink : symlink NCM component.
 
 =head1 DESCRIPTION
 
-Object to create/delete symbolic links. When creating symlinks, target existence can be checked. And clobbering can be disabled.
-
-Target definition may contain contextual variable, using syntax '{variable}'. Variables are defined globaly for all links and allow to use the same target definition for different contexts (e.g. platforms).
-
-Target definition and contextual variable value may also contain expression interpreted as shell commands (.eg. uname). These expressions must enclosed between pair of '@@'. Unless shell command must reevaluated for each link, it is better to associate the shell command with a contextual variable and use the variable in the target definition, as a contextual variable is evaluated once (global).
+Object to create/delete symbolic links. When creating symlinks, target existence can be checked. And clobbering can be disabled. Also, target definition can be simplified by
+the use of contextual variables and command outputs.
 
 
 =head1 RESOURCES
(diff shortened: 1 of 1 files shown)


jouvin:
@stdweird let me know if you find the new description clearer. And if you think I should add more examples...

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -17,9 +17,9 @@ symlink : symlink NCM component.
 
 Object to create/delete symbolic links. When creating symlinks, target existence can be checked. And clobbering can be disabled.
 
-Target definition may contain contextual variable, using syntax '{variable}'. Variables are defined globaly for all links and allow to use the same target definition for different contexts (e.g. platforms).
+Target definition may contain contextual variable, using syntax C<{variable}>. Variables are defined globaly for all links and allow to use the same target definition for different contexts (e.g. platforms).
```
stdweird: `globally`

stdweird: and maybe add a newline to avoid the long lines?

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -17,9 +17,9 @@ symlink : symlink NCM component.
...
 
-Target definition and contextual variable value may also contain expression interpreted as shell commands (.eg. uname). These expressions must enclosed between pair of '@@'. Unless shell command must reevaluated for each link, it is better to associate the shell command with a contextual variable and use the variable in the target definition, as a contextual variable is evaluated once (global).
+Target definition and contextual variable value may also contain expression interpreted as shell commands (.eg. uname). These expressions must enclosed between pair of C<@@>. Unless shell command must reevaluated for each link, it is better to associate the shell command with a contextual variable and use the variable in the target definition, as a contextual variable is evaluated once (global).
```
stdweird: `contain expression` -> either `an expression` or `expressions`
`.eg.` -> `e.g.`
an example of `enclosed between pair of C<@@>` with uname?
`must` -> `must be`

when is the contextual variable evaluated? at the begin or as encountered during the run?

(and maybe some newlines, this is a very long line)

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+fields: 
 
-'name' is the symbolic link name, 'target' is the link target, other fields are boolean. 'delete' means delete the link rather than create. 'exists' means check the target exists when creating or check the link exists when deleting. 
+=over
 
-'replace' is an option taking one or several of the several following suboptions : all, dir, dirempty, file, link, none. Each suboption can have the following values : 'yes', 'no', a string interpreted as a file extension used to rename the existing object with the link name. If a suboption value is empty, it is interpreted as 'yes'. If value is 'yes', replacement is enabled but existing object is not renamed, except if a default extension has been defined with 'none' or 'all' or if it is a non empty directory (renamed with a default extension).
+=item *
 
-When 'all' or 'none' suboption is specified, other suboptions can be specified too to redefine extension or to prevent replacement of a specific object type. 'dirempty' means directory is replaced only if empty. 'none=extension' can be used to establish a default backup extension without enabling replacement (as it is with 'all').
+C<name>: symbolic link name
```
stdweird: this is probably just me not being able to remember, but can you add clarify name and target here (the `ln` man has `create a link to TARGET with the name LINK_NAME`; and maybe add the familiar command `ln TARGET LINK_NAME` to clarify what is meant with target and name.

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+
+=item *
+
+C<target>: link target path
+
+=item *
+
+C<delete>: (boolean) delete the symlink rather than creating it
```
stdweird: delete the `name`, not the `target`; also how does this mix with `exists`? will delete trigger an erro when the target does not exist?

stdweird: also, do you need to specify the target, as only the name is involved here?

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+
+C<target>: link target path
+
+=item *
+
+C<delete>: (boolean) delete the symlink rather than creating it 
+
+=item *
+
+C<exists>: (boolean) check that the target exists when creating it or check that the link exists when deleting it
```
stdweird: use `name` instead of `link` for the delete part. what happens when you don't set this and the link does not exists? error logged? and if it is set, does it ignore anything missing

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+
+=item *
+
+C<replace>: (nlist) option used to specify the action to do when an object with the same name as the symlink already exists, depending on the object type.
```
stdweird: `action to do` -> `action to take`? (what happens if not set and name exists)
`an object with the same name as the symlink` you mean if the name already exists, and what sort of objects are defined: existing links, or files and/or directories? and is there a backup option, esp if you support replacing files/directories?

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+
+C<replace>: (nlist) option used to specify the action to do when an object with the same name as the symlink already exists, depending on the object type. 
+Possible actions are: do not define the symlink, replace the
+object by the symlink or define the symlink after renaming the object. The nlist keys and values can be:
+
+=over
+
+=item *
+
+key: the existing object type. Valid values are: C<all>, C<dir>, C<dirempty>, C<file>, C<link>, C<none>.  C<dirempty> means an empty directory only, C<dir> means any directory.
```
stdweird: dirempty needs to work first, i wouldn't mention it here. (i would actually remove it from the schema)

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+C<all> and C<none> are mutually exclusive but can be combined with other object types to define the extension to use when renaming a given object type or to prevent/enable replacement for a specific object type.
```
stdweird: is the exclusive part forced in the schema?
and this needs an example, i really don't understand the `can be combined` part (it probably very useful and powerful, but i just don't get it)

ncm-symlink/src/main/perl/symlink.pod
stdweird: ok, most of this is explained below.

ncm-symlink/src/main/perl/symlink.pod
stdweird: maybe double check the terminology used below. i guess when you write `symlink` you mean `name`

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,13 +28,62 @@ Target definition and contextual variable value may also contain expression inte
...
+
+=back
+
+C<replace> option allows a lot of flexibility in specifying what should be done in case of conflict with an existing object. It implements the following advanced features:
+
+=over
+
+=item *
+
+C<none=extension> can be used to establish a default rename extension without enabling replacement. This extension will be used with object types for which replacement is enabled.
```
stdweird: example please

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -91,25 +141,26 @@ A list of global options used as default for all links creation/deletion. Suppor
...
-None known.
-
-=head1 AUTHOR
-
-Michel Jouvin <jouvin@lal.in2p3.fr>
-
-=head1 MAINTAINER
-
-Michel Jouvin <jouvin@lal.in2p3.fr>
+No known bugs.
```
stdweird: remove this whole block starting from DEPENDENCIES, it does not really add much

stdweird:
@jouvin some more examples of the allowed combination really is necessary

jouvin:
I'll try to add them this week end...

ncm-symlink/src/main/perl/symlink.pod
jouvin: Do you mean it doesn't? In this case, I'd suggest opening an issue rather than removing from the doc...

ncm-symlink/src/main/perl/symlink.pod
jouvin: No it is not enforced by the schema... The schema would certainly benefit from a few validation functions, please could you open an issue for 15.6?

jouvin:
@stdweird please, let me know if you think more examples are needed, I tried to add some but there are so many combinations possible that I'm wondering which ones are the most useful.

ncm-symlink/src/main/perl/symlink.pod
stdweird: see https://github.com/quattor/configuration-modules-core/issues/380

ncm-symlink/src/main/perl/symlink.pod
stdweird: https://github.com/quattor/configuration-modules-core/issues/501

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
-=head2 /software/components/symlink/context
+=item *
 
-A list of contextual variables to use in target definitions.  Each entry
-must be of the structure_symlink__context_entry type which has the following
-fields: name, value.  Contextual variables are global.
+C<target>: link target path. The target path can be built using a command output with the command to execute between a pair of C<@@> or a contextual variable using the syntax
+C<{variable}> (variables are defined in C</software/components/symlinks/context>). Unless the shell command between C<@@> must be reevaluated for each link, it is better 
+to associate the shell command
+with a contextual variable and use the variable in the target definition, as a contextual variable is evaluated once (global).
```
stdweird: add example `C<"/software/components/symlinks/context/myvariable" = "@mycommand --option@";>` (unless i'm mistaken with the inner details)

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
+C<replace> option allows a lot of flexibility in specifying what should be done in case of conflict with an existing object. It implements the following advanced features:
 
-None known.
+=over
 
-=head1 AUTHOR
+=item *
 
-Michel Jouvin <jouvin@lal.in2p3.fr>
+C<none=extension> can be used to establish a default rename extension without enabling replacement. This extension will be used with object types for which replacement is enabled.
```
stdweird: the `without enabling replacement`  vs `for which replacement is enabled` is still a bit unclear. you mean it's a default and thus you don't need to specify it each time you enable replacement?

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
+=item *
 
-Michel Jouvin <jouvin@lal.in2p3.fr>
+Action can be empty. If a default rename extension was defined with C<none=extension>, the object will be renamed before defining the symlink. Else it is interpreted as C<yes>.
 
-=head1 VERSION
+=back
 
-1.3.2
+=head2 /software/components/symlink/context
```
stdweird: in the description above, i would simply refer to this section (otherwise it's twice the same information and might get out of sync).

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
+with the exception of C<delete>.
+
+=head1 EXAMPLES
 
-=head1 SEE ALSO
+    # Define global variable osdir so that it can be use to define symlink targets
+    "/software/components/symlink/context" = {
+      append(nlist(
+               "name",    "ostype",
+               "value",   "@@uname@@",
```
stdweird: ah, this is what you mean a pair of `@@` :smile:

ncm-symlink/src/main/perl/symlink.pod
stdweird: so should be `C<"/software/components/symlinks/context/myvariable" = "@@mycommand --option@@";>` if i follow the exmaple below

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
+              "exist",    true
+      ));
+
+      # Define /lhcb with a target actual value including a contextual variable.
+      # The contextual variable can be defined before or later in the configuration.
+      append(nlist(
+              "name",    "/lhcb",
+              "target",   "/lhcb_prod/{ostype}",
+              "exists",    true
+      ));
```
stdweird: close the `}`

stdweird: or remove the block below (as it's also there at the end of the example)

ncm-symlink/src/main/perl/symlink.pod
```diff
@@ -28,92 +25,148 @@ Target definition and contextual variable value may also contain expression inte
...
+
+      # Define symlink /etc/alpine/conf, replacing an existing
+      # file by the symlink without renaming it
+      append(nlist(
+               "name", "/etc/alpine/pine.conf",
+               "target", "/lal/gen/etc/pine.conf",
+               "replace",  nlist("all", "yes"),
+      ));
+
+      # Define /htdocs as a link only if /htdocs doesn't exist or is a symlink
```
stdweird: with `or is a symlink` you mean `or is a symlink from a different target`?

stdweird:
@jouvin it's a lot more clear with the examples. thanks for clarifying!

ncm-symlink/src/main/perl/symlink.pod
jouvin: Yes it was not a mistake, this is not a pair of `@`... I'm open to any clearer wording you may have in mind!

stdweird: no clearer wording needed, the example makes it very clear. maybe simialr to the way you introduce the variable with `{variable}`, you can do command `@@command@@`?

ncm-symlink/src/main/perl/symlink.pod
jouvin: See also #502 on this topic...

ncm-symlink/src/main/perl/symlink.pod
jouvin: Hopefully clarified in upcoming mod...

ncm-symlink/src/main/perl/symlink.pod
jouvin: Thanks for noticiing. This was a leftover in the previous edit.

jouvin:
@stdweird I tried to clarify the points in your last comments. Thanks for this careful review... As usual, when you try to write a proper documentation, you realize all the things that could be improved!

stdweird:
LGTM

PR #504, opened 2015-05-09
ncm-metaconfig: service ganesha service change stat_exporter Access type

### LINKED ISSUE #499: ncm-metaconfig: service ganesha unittest fails on EL5
output from diag
```
# STAT_EXPORTER {
#     Access = "";
#     Port = 10401;
# }
```
test that fail
```
not ok 609 - Found at least one match (total 0 matches) for test idx 1 (pattern (?-xism:(?m:^\s{4}Access\s=\s"localhost";)))
not ok 610 - Order ok for test idx 1 (pattern (?-xism:(?m:^\s{4}Access\s=\s"localhost";))) (lastpos 1420 before -1)
```

the TT code is join on, the schema assumes a string instead of a list (do more recent TT don't mind the `.join` vmethod on scalar?) anyway, should be fixed

not blocking for rc1, but still for the release https://github.com/quattor/release/issues/85

stdweird:
The Access attribute is assumed a comma separated list in the TT files (and auto-promotion from scalar to list with the CCM::TT::Scalar instances seems to fail with TT on EL5)

Fixes #499

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/ganesha/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/ganesha/pan/schema.pan
@@ -226,7 +226,7 @@ type ganesha_snmp_adm = {
 } = nlist();
 
 type ganesha_stat_exporter = {
-    "Access" : string = "localhost"
+    "Access" : string[] = list("localhost")
     "Port" : long = 10401
 } = nlist();
 


jrha:
Is this not backwards incompatible?

stdweird:
yes, but it's a bug in the schema. not sure what the rules are for that.

jrha:
Rules? :smile:

PR #507, opened 2015-05-14
ncm-mysql: allow an empty password for a DB user

### LINKED ISSUE #506: ncm-mysql 15.4.0-rc2: empty password should be allowed for db users
Even if not recommended, this is a valid configuration from the MySQL point of view and it used to be supported (in this case the password is not set by the component, this doesn't prevent setting it manually if I'm right). Validation introduced in https://github.com/quattor/configuration-modules-core/pull/484 doesn't allow this anymore.
To be fixed before 15.4 final release(https://github.com/quattor/release/issues/92).

jouvin:
Required for backward compatibily. If desirable to change this, it would be good to open a separate issue and discuss it as part of 15.6 release.

Also applies the same validation rules for admin password (empty password excluded)

Fixes #506

### DIFF
--- a/ncm-mysql/src/main/pan/components/mysql/schema.pan
+++ b/ncm-mysql/src/main/pan/components/mysql/schema.pan
@@ -34,10 +34,19 @@ function component_mysql_check_db_script = {
   };
 };
 
+# The function validating password imposes a non-empty string.
+function component_mysql_password_valid = {
+  if (  match(SELF, '^[^\\]+$') ) {
+    true;
+  } else {
+    false;
+  };
+};
+
 type component_mysql_user_right = string with match(SELF, '^(ALL( PRIVILEGES)?|ALTER( ROUTINE)?|CREATE( (ROUTINE|TEMPORARY TABLES|USER|VIEW))?|DELETE|DROP|EVENT|EXECUTE|FILE|GRANT OPTION|INDEX|INSERT|LOCK TABLES|PROCESS|REFERENCES|RELOAD|REPLICATION (CLIENT|SLAVE)|SELECT|SHOW (DATABASES|VIEW)|SHUTDOWN|SUPER|TRIGGER|UPDATE|USAGE)$');
 
 type component_mysql_db_user = {
-  'password'  : string with match(SELF, '^[^\\]+$')
+  'password'  : string with (length(SELF) == 0) || component_mysql_password_valid(SELF)
   'rights'    : component_mysql_user_right[] = list('SELECT')
   'shortPwd'  : boolean = false
 };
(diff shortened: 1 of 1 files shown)


ncm-mysql/src/main/pan/components/mysql/schema.pan
```diff
@@ -34,10 +34,19 @@ function component_mysql_check_db_script = {
...
+  } else {
+    false;
+  };
+};
+
 type component_mysql_user_right = string with match(SELF, '^(ALL( PRIVILEGES)?|ALTER( ROUTINE)?|CREATE( (ROUTINE|TEMPORARY TABLES|USER|VIEW))?|DELETE|DROP|EVENT|EXECUTE|FILE|GRANT OPTION|INDEX|INSERT|LOCK TABLES|PROCESS|REFERENCES|RELOAD|REPLICATION (CLIENT|SLAVE)|SELECT|SHOW (DATABASES|VIEW)|SHUTDOWN|SUPER|TRIGGER|UPDATE|USAGE)$');
 
 type component_mysql_db_user = {
-  'password'  : string with match(SELF, '^[^\\]+$')
+  'password'  : string with (length(SELF) == 0) || component_mysql_password_valid(SELF)
```
jrha: Much simpler to just change the `+` to `*`.

jouvin: The idea is that validation function consider that a valid password is not empty and for some property we allow it to be either empty or valid... this is the idea.

stdweird: the length 0 check should be moved to the function, and (this part is not a clokcer for me, just my opinion) imho only allow it after some variable `ALLOW_INSECURE_EMPTY_PASSWORD` is set. i really don't like support for crippled security; might just as well make the password optional (allowing empty passwords is a bug, not something that requires backwards compatibility).

jouvin: I really disagree on both points. MySQL allows empty passwords so this is not crippy to allow them in the configuration module... And as soon as you want to support them, there is no reason why the function should impose it and use a complex mechanism to reallow it (the global variable solution is particularly weird as you may want to enforce something different for the admin password and the user passwords.
As I said, I'd appreciate that this PR is merged in 15.4 as it really breaks backward compatibility in the sense that you cannot install it without updating your local config, if you configure some grid components with the template library and have not defined a password for some services... And I also agree that this default behaviour should be rediscussed and it would be fine with me if we open an issue to do it in 15.6.

jrha:
Would a better change not be to set a non-empty default in the template library?

jouvin:
I'm resisting any change like this without the time to assess the potential side effects. I'm ready for all these changes in 15.6...

stdweird:
@jouvin ok, i opened an issue for it (and assigned it to you)
https://github.com/quattor/configuration-modules-core/issues/509
lets continue the discussion there

jrha:
Yeah ok.

PR #511, opened 2015-05-18
ncm-metaconfig: daemon user validation/deprecation should return true

### LINKED ISSUE #510: ncm-metaconfig: daemon deprecation must return boolean
validation with deprecated `daemon` field fails with
```
user-defined validation function returned an element of type undef; boolean value required
```
fix is trivial, expect PR soon

stdweird:
fixes #510


ned21:
oops! Bad bug. :( I believe Cal has said that just saying "true;" is more performant than an explicit return(true) but happy to let this one go by to get the release out...

stdweird:
@ned21 fixed

jrha:
bleh :frowning: thanks for the catch!

PR #513, opened 2015-05-21
ncm-metaconfig: service elasticsearch

### LINKED ISSUE #404: ncm-metaconfig: Include elasticsearch TT files
Split off from #311 

Needs new maven-tools and a PR with the ES config and schema.

stdweird:
Fixes #404


stdweird:
@jrha these are unittested with json_typed. when using json but no json_typed, all longs and doubles are strings (`2` becomes `'2' etc); so your mileage may vary if this works or not.

jrha:
Tested: longs and doubles seem to work fine with `''` around them.

PR #514, opened 2015-05-21
ncm-metaconfig: add kibana service

stdweird:
Adds support for kibana4 config (tested with 4.1.0 prerelease)

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/kibana/pan/config.pan
+++ b/ncm-metaconfig/src/main/metaconfig/kibana/pan/config.pan
@@ -1,0 +1,3 @@
+unique template metaconfig/kibana/config;
+
+include 'metaconfig/kibana/schema';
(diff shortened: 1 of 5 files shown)


stdweird:
Requires `json_typed` to render correctly, otherwise your mileage may vary (doubles and logs become quoted strings).

stdweird:
This is valid kibana config for kibana 4.1.0 (proper rpms with init scripts should be in the rpms, if not use https://github.com/elastic/kibana/pull/3212 to make your own)

the only thing missing to make this work is a simle proxy rule in httpd like
```
    <location ~ "^/kibana/(.*)$">
        proxypassmatch  http://localhost:5601/$1
        proxypassreverse http://localhost:5601/$1
    </location>
```
or if you prefer using metaconfig,

```pan
prefix "/software/components/metaconfig/services/{/etc/httpd/conf.d/ssl.conf}/contents/vhosts/base/";
"locations" = append(dict(
        "regex", true,
        "name", "^/kibana/(.*)$",
        "proxy" , dict(
            "pass", list(dict(
                "match", true,
                "url", "http://localhost:5601/$1",
                )),
            "passreverse", list(dict(
                "url", "http://localhost:5601/$1",
                )),
            ),
        ));
```

jrha:
LGTM

PR #515, opened 2015-05-22
ncm-metaconfig: service logstash add support for filter kv

stdweird:
Add support for the `bytes2human` and `kv` filters.


itkovian:
It looks ok to me. But I am no expert.

jrha:
Where does the `bytes2human` filter come from? It's not in upstream logstash AFAIK.

stdweird:
we package it as part of our patterns (https://github.com/hpcugent/logstash-patterns/blob/master/logstash/filters/bytes2human.rb), someone opened an issue for it https://github.com/elastic/logstash/issues/3515

stdweird:
and to be clear: this PR does not make it mandatory or anything like that

PR #522, opened 2015-06-04
ncm-metaconfig: service logstash: change forwarder configfile location 

stdweird:
With the first 0.4.0 rpm released by elastic, the default configfile used by the rpm is now in `/etc/logstash-forwarder.conf`


stdweird:
this is highly non-backwards compatible, but i think it is a valid change (otherwise you are required to change the location via sysconfig to the old value). the old location is/was the default built-in, but with rpm packages available now from elastic website (before, you had to make your own rpms incl init scripts), one can consider this a new default location (imho).

@jrha @jouvin  any thoughts?

jrha:
I'm okay with it

PR #524, opened 2015-06-08
ncm-network: Add fqdn to interface_alias type

jrha:
This is inserted by aquilon for any aliases with hostnames.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -18,7 +18,7 @@ type structure_interface_alias = {
     "ip"      ? type_ip
     "netmask" : type_ip
     "broadcast" ? type_ip
-    "fqdn"    ? string
+    "fqdn"    ? type_fqdn
 };
 
 @documentation{
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -18,6 +18,7 @@ type structure_interface_alias = {
     "ip"      ? type_ip
     "netmask" : type_ip
     "broadcast" ? type_ip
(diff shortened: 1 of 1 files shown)


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -18,6 +18,7 @@ type structure_interface_alias = {
     "ip"      ? type_ip
     "netmask" : type_ip
     "broadcast" ? type_ip
+    "fqdn"    ? string
```
stdweird: what is this supposed to do/achieve?

ned21: Every interface has a fqdn in DNS. e.g. myhost-hb.dc1.quattor.org is an interface of myhost.dc1.quattor.org.

stdweird: that makes sense, but how is this used? e.g. is there a component that uses this? (i'm just curious)

gombasg:
Currently not used directly for configuration, but it gets into various reports.

ncm-network/src/main/pan/components/network/core-schema.pan
jrha: But perhaps it should be `type_hostname`?

jrha:
They show up in `aq dump_dns` and so can be used to configure a DNS server.

stdweird:
retest this please

stdweird:
@jrha add `ncm-network:` in title

ncm-network/src/main/pan/components/network/core-schema.pan
stdweird: yes please

stdweird:
LGTM

PR #526, opened 2015-06-11
ncm-ssh: Fix small bugs

piojo:
* Add missing options to the schema.
* Fix lines that match partially the profile.


piojo:
Mmm... this fails in ncm-opennebula, which I didn't touch. It seems you'll have to update panc in the Jenkins server?

stdweird:
@piojo it fails because the current `template-library-core-pr-builder` has invalid templates.

stdweird:
retest this please

PR #532, opened 2015-07-13
ncm-download timeouts

ajf8:
We've been having problems with the ncm-download component taking a long time to timeout because there is no timeout set.

I found some code in sudo.pm which adds a timeout switch to the curl command (though it is not in the schema). However, there is no timeout on the LWP client which does the HEAD request before this.

So I've added a head_timeout parameter at the component level, with a default of 3 seconds set. The curl timeout has been added to schema.pan, and it is possible to override it per file.

### DIFF
--- a/ncm-download/src/main/perl/download.pm
+++ b/ncm-download/src/main/perl/download.pm
@@ -121,6 +121,9 @@ sub Configure {
             $min_age = $inf->{files}->{$f}->{min_age};
         }
 
+	my $timeout = $inf->{files}->{$f}->{timeout};
+	$timeout = $inf->{timeout} unless defined($timeout);
+
         my $success = 0;
         if (@proxyhosts && $inf->{files}->{$f}->{proxy}) {
             my $attempts = scalar @proxyhosts;
@@ -133,7 +136,8 @@ sub Configure {
                 $success += $self->download(
                     file => $file,
                     href => $source,
-                    timeout => $inf->{timeout},
+                    timeout => $timeout,
+                    head_timeout => $inf->{head_timeout},
                     proxy => $proxy,
                     gssneg => $gss,
                     cacert => $inf->{files}->{$f}->{cacert},
@@ -216,7 +220,9 @@ sub get_head
     local $ENV{'HTTPS_CA_FILE'} = $opts{cacert} if (exists($opts{cacert}));
     local $ENV{'HTTPS_CA_DIR'} = $opts{capath} if (exists($opts{capath}));
 
-    return LWP::UserAgent->new->head($source);
+    my $lwp = LWP::UserAgent->new;
+    $lwp->timeout($opts{head_timeout}) if (defined($opts{head_timeout}));
+    return $lwp->head($source);
 }
 
 sub download {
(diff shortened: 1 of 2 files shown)


ncm-download/src/main/perl/download.pm
```diff
@@ -133,7 +133,8 @@ sub Configure {
                 $success += $self->download(
                     file => $file,
                     href => $source,
-                    timeout => $inf->{timeout},
+                    timeout => $inf->{files}->{$f}->{timeout} || $inf->{timeout},
```
stdweird: if the timeout of a file `$inf->{files}->{$f}->{timeout}` is set to 0, this code will override it with whatever is set in as global timeout `$inf->{timeout}`.

ncm-download/src/main/perl/download.pm
```diff
@@ -216,7 +217,9 @@ sub get_head
     local $ENV{'HTTPS_CA_FILE'} = $opts{cacert} if (exists($opts{cacert}));
     local $ENV{'HTTPS_CA_DIR'} = $opts{capath} if (exists($opts{capath}));
 
-    return LWP::UserAgent->new->head($source);
+    my $lwp = LWP::UserAgent->new;
+    $lwp->timeout($opts{head_timeout}) if (exists($opts{head_timeout}));
```
stdweird: if i read the code correct, `$opts{head_timeout}` will always exist, but could be undefined. so this should probably be `if (defined($opts{head_timeout}))`

stdweird:
test this please

ncm-download/src/main/perl/download.pm
```diff
@@ -121,6 +121,9 @@ sub Configure {
             $min_age = $inf->{files}->{$f}->{min_age};
         }
 
+	my $timeout = $inf->{files}->{$f}->{timeout};
+	$timeout = $inf->{timeout} unless defined($timeout) && $timeout == 0;
```
stdweird: why the ` && $timeout == 0`? the `unless defined($timeout)` should be enough (and it should be possible to force `timeout=0` for a file; unless that makes no sense)

ncm-download/src/main/pan/components/download/config.pan
```diff
@@ -33,4 +33,4 @@ prefix '/software/components/${project.artifactId}';
 
 'active' ?= true;
 'dispatch' ?= true;
-
+'head_timeout' ?= 3;
```
stdweird: if this is the default, it's better to set it via the schema. about the value, i have no clue if this is a lot or not . but since there was no default previously, a larger default value to mimc no timeout might be appropriate.

ncm-download/src/main/perl/download.pm
ajf8: Right, I've removed that condition. Thanks.

ncm-download/src/main/pan/components/download/config.pan
ajf8: Since it's just the HEAD request to check if a file has changed, it should be very fast, so for us 3 seconds is a long time. But it's a good point, it may not be a good default for others. So I think it's best we don't set any default on this, have changed it.

stdweird:
LGTM

PR #533, opened 2015-07-13
ncm-sudo fixes

ajf8:
Hi,

One change is for ncm-sudo on Solaris, it fails because /proc/self/fd/0 is not a valid handle for stdin. Use a - instead on Solaris (this doesn't seem to work on a Linux tested).

Also, while trying to upgrade to the latest upstream, we are having a problem with the schema change in e79f799. The privilege options parameter was changed to require a colon at the end, which we do not have in our templates. It is difficult because it requires that we change the schema, templates and components all at once. So I propose this change which, I hope, will allow them to be changed independently. It does this by making the colon optional in the schema, and adding it in the code if it's not present.

Finally, a true (1) return was missing at the end of the package.

### DIFF
--- a/ncm-sudo/src/main/perl/sudo.pm
+++ b/ncm-sudo/src/main/perl/sudo.pm
@@ -44,7 +44,9 @@ use constant {
     INCLUDES_DIRS    => PROFILE_BASE . "includes_dirs"
 };
 
-use constant VISUDO_CHECK => qw(/usr/sbin/visudo -c -f /proc/self/fd/0);
+use constant VISUDO_CHECK => qw(/usr/sbin/visudo -c -f);
+use constant VISUDO_STDIN_LINUX => '/proc/self/fd/0';
+use constant VISUDO_STDIN_SOLARIS => '-';
 
 # All possible "Default" options separated by type.
 use constant BOOLEAN_OPTS => qw(
@@ -210,6 +212,7 @@ sub generate_privilege_lines {
 sub is_valid_sudoers {
     my ($self, $fh) = @_;
     my ($err, $out);
+
     my $proc = CAF::Process->new (
         [VISUDO_CHECK],
         stdin => "$fh",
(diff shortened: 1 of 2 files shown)


stdweird:
test this please

ajf8:
test this please

stdweird:
test this please

ncm-sudo/src/main/perl/sudo.pm
```diff
@@ -197,8 +199,11 @@ sub generate_privilege_lines {
         my $ln = $info{PRIVILEGE_USER()}->getValue;
         $ln .= "\t" . $info{PRIVILEGE_HOST()}->getValue;
         $ln .= "= (". $info{PRIVILEGE_RUNAS()}->getValue . ")\t";
-        $ln .= $info{PRIVILEGE_OPTS()}->getValue . "\t"
-        if defined $info{PRIVILEGE_OPTS()};
+	if (defined($info{PRIVILEGE_OPTS()})) {
+		$ln .= $info{PRIVILEGE_OPTS()}->getValue;
+		$ln .= ':' if $ln !~ /:$/;
```
stdweird: just to understand this, the ending colon is mandatory, https://github.com/ajf8/configuration-modules-core/commit/e79f799ab4bfa0025b869c7bf7e124be0b048f7f tried to force that via schema, the new code forces it via the component?

stdweird:
@ajf8 only limited list of people can trigger the tests (and i don't know who you are). since this touches solaris and you are london based, i'm going to guess you are working for MS?
@ned21 @gombasg i'll add @ajf8 to the list if you could confirm.

ncm-sudo/src/main/perl/sudo.pm
ajf8: Yeah, this is enforcing it by the component, but only if it isn't set via the templates. And it makes the colon optional in the templates.

The code used to always add a colon at the end (changed in 31333b8). This means if we upgrade the schema and change the templates before the component is upgraded, we will have two colons at the end of the line, and the sudoers file will not apply.

If we upgrade the component before the schema, we will have no colon after the option, and again sudoers will not apply.

Because we cannot upgrade both at the same time, it makes upgrading to this latest upstream in its current state quite awkward. This change should allow either way to work.

stdweird: sounds like some backwards incompatible commits were added, and lack of unittests didn't catch this change.
if you can, you may always add a unittest for this behaviour (ensure the trailing `:` in the privilige opts). if not, open an issue to add this unittest.

stdweird:
@jrha LGTM

gombasg:
@stdweird Yup, we're happy to have @ajf8 on board, please allow him to trigger the tests. Thanks!

Issue #525, opened 2015-06-10
ncm-spma: el7 yum has fastmirror enabled by default


stdweird:
and it will use (external) mirrors if it can.
we need a way to disable it, e.g. if a proxy is defined.

stdweird:
also EL7 sets up default yum cron jobs that download rpms and things like that. should also be controlled by spma.

stdweird:
on the cron jobs: `yum` installs the the cron files,
```
# rpm -q --filesbypkg yum |grep cron
yum                       /etc/cron.daily
yum                       /etc/cron.daily/0yum-daily.cron
yum                       /etc/cron.hourly
yum                       /etc/cron.hourly/0yum-hourly.cron
yum                       /etc/yum/yum-cron-hourly.conf
yum                       /etc/yum/yum-cron.conf
```

however, the actual binary that does the work is part of `yum-cron`, so if that  is not installed, no issues with cron doing yum stuff.

stdweird:
fixed by #535

PR #535, opened 2015-07-16
ncm-spma: support yum plugin configuration (including default versionlock and fastestmirror)

### LINKED ISSUE #534: ncm-spma: adding plugin support
currently, yum-based spma relies on the `versionlock` yum plugin; and on EL7, the `fastestmirror` plugin is also in use by default and can cause unwanted behaviour when running spma (see also #525).
none of these are currenlty controlled/managed by quattor.

i propose to add support for both in ncm-spma, but i'm struggling with the default behaviour. 
i.e. what does ncm-spma do when nothing is configured wrt plugins.

i would create the following configs by default: 
for versionlock
```
[main]
enabled=1
locklist = /etc/yum/pluginconf.d/versionlock.list
```

and for fastestmirror (generating the configfile, not making it a dependency of spma as is the case with versionlock)
```
[main]
enabled=0
```

any thoughts?

stdweird:
Also includes some cleanup and TT unittests
Fixes #534 #525

By default (i.e. unless the plugins are explicitly configured), this code will enable the versionlocklist plugin with config 
```
[main]
enabled=1
locklist=/etc/yum/pluginconf.d/versionlock.list
```
and disable the fastestmirror plugin with config
```
[main]
enabled=0
```


stdweird:
tests will fail until we can use https://github.com/quattor/maven-tools/pull/56

ned21:
Code looks OK to me.

stdweird:
@ned21 are you also ok with the default behaviour? (i updated the description)

ned21:
What you propose looks sensible, yes.

stdweird:
@ned21 test run now with new buildtools

PR #541, opened 2015-08-05
ncm-metaconfig: perfsonar: replace module general with TT

stdweird:
Due to https://github.com/quattor/CAF/issues/81 and https://github.com/quattor/CAF/pull/97, module `general` can't be used anymore.


jrha:
LGTM

PR #542, opened 2015-08-05
ncm-metaconfig: named: add newline after each include

stdweird:
Although multiple includes separated by `;` is valid, inserting a newline after each include makes it much more readable


jrha:
LGTM

PR #546, opened 2015-08-11
ncm-network: Actually use realhostname if it has been provided

### LINKED ISSUE #462: ncm-network: realhostname not used
Current realhostname code
```perl
if ($config->elementExists($path."/realhostname")) {
    $text .= "HOSTNAME=".$config->getValue($path."/hostname")."\n";
}
```
which has behaviour that setting `realhostname` to anything sets the short hostname as `hostname`.
fixing this bug might break current templates that rely on this "feature".

jrha:
Fixes #462


stdweird:
title should start with `ncm-network`

stdweird:
:+1:  on the simple refactoring
LGTM

PR #547, opened 2015-08-11
ncm-opennebula: Fix database schema

alvarosimon:
This PR includes a small schema fix to check opennebula db backend.
    https://github.com/quattor/template-library-core/issues/69 : schema should allow for the possibility to use sqlite as one db backend


stdweird:
LGTM

PR #550, opened 2015-08-12
ncm-ceph: Extend schema with rgw options, and spread over multiple files

### LINKED ISSUE #333: ncm-ceph: validate radosgw configuration schema
radosgw config schema not validated yet in #320 . The config parameters should be checked and tested

kwaegema:
fixes #333

### DIFF
--- a/ncm-ceph/src/main/pan/components/ceph/schema-rgw.pan
+++ b/ncm-ceph/src/main/pan/components/ceph/schema-rgw.pan
@@ -2,7 +2,7 @@ declaration template components/${project.artifactId}/schema-rgw;
 
 type type_quoted_string = string with match(SELF, '^".*"$');
 
-@documentation{ ceph rados gateway config }
+@documentation{ configuration options for a ceph rados gateway instance }
 type ceph_radosgw_config = { 
     include ceph_daemon_config
     'host'      : string
@@ -24,7 +24,7 @@ type ceph_radosgw = {
     'config' ? ceph_radosgw_config
 };
 
-@documentation{ ceph rados gateway host }
+@documentation{ ceph rados gateway host, defining all gateways on a host }
 type ceph_radosgwh = {
     'fqdn'      : type_fqdn
     'gateways'  : ceph_radosgw{}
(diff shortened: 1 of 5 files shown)


ncm-ceph/src/main/pan/components/ceph/schema-rgw.pan
```diff
@@ -0,0 +1,30 @@
+declaration template components/${project.artifactId}/schema-rgw;
+
+@documentation{ ceph rados gateway config }
+type ceph_radosgw_config = { 
+    include ceph_daemon_config
+    'host'      : string
+    'keyring'   : string
+    'rgw_socket_path' : string = ''
+    'log_file'  : string = '/var/log/radosgw/client.radosgw.gateway.log'
+    'rgw_frontends' : string = "\"civetweb port=8000\""
```
stdweird: should be list, and actual string generated by component

kwaegema: Hmm, i don't find this option documented somewhere, but I've found some examples where it is not quoted, so I guess It don't has to be this dirty. @gvasilak can I remove the inner quotes?

stdweird:
retest this please

ncm-ceph/src/main/pan/components/ceph/schema-rgw.pan
gvasilak: It is currently unclear to me what it will actually parse properly. Judging by [this bug report](https://bugzilla.redhat.com/show_bug.cgi?id=1228845), it seems that leaving the quotes in is the safest bet.
Ken Dreyer, in comment 3 on that is referring to [this PR](https://github.com/ceph/ceph/pull/5103), which suggests that even the Ceph developers are not quite sure what it can and should parse properly.
I suggest we leave the quotes, watch that PR and revisit this when they've resolved it on the Ceph side.

stdweird: the component should handle the quotes, or you need to make a type that test this

stdweird:
LGTM

jrha:
Looks find code wise, but most of the `@documentation` blocks are essentially pointless (they just repeat the name of the type)

kwaegema:
@jrha extended documentation a little bit. don't know what more to say :smile:

jrha:
Thanks, no I don't know either :smile:

PR #554, opened 2015-08-13
ncm-ccm: change profile_failover from hostURI to hostURI[].

ajf8:
CCM will be supporting multiple failover URLs (comma seperated in the
config file) so this change will allow that to be configured by this
component. Code is backwards compatible with old schema.

### DIFF
--- a/ncm-ccm/src/main/perl/ccm.pm
+++ b/ncm-ccm/src/main/perl/ccm.pm
@@ -52,7 +52,8 @@ sub Configure
     delete($t->{version});
 
     while (my ($k, $v) = each(%$t)) {
-        print $fh "$k $v\n";
+        my $value = ref($v) eq 'ARRAY' ? join(',', @$v) : $v;
+        print $fh "$k $value\n" if length($value);
     }
 
     if (_is_noquattor()) {
--- a/ncm-ccm/src/main/pan/components/ccm/schema.pan
+++ b/ncm-ccm/src/main/pan/components/ccm/schema.pan
@@ -12,7 +12,7 @@ type component_ccm = {
     include structure_component
     'configFile'       : string = '/etc/ccm.conf'
     'profile'          : type_hostURI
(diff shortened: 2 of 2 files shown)


jrha:
LGTM

ncm-ccm/src/main/perl/ccm.pm
```diff
@@ -52,7 +52,11 @@ sub Configure
     delete($t->{version});
 
     while (my ($k, $v) = each(%$t)) {
-        print $fh "$k $v\n";
+        if (ref($v) eq 'ARRAY') {
```
stdweird: this will print warnings with scalars. `ref($v)` of scalar `$v` is undef, and `undef eq "string"` makes perl unhappy.
i would replace this with
```perl
my $value = ($ref($v) && $ref($v) eq 'ARRAY') ? join(',', @$v) : $v;
print $fh "$k $value\n";
```

ncm-ccm/src/main/perl/ccm.pm
```diff
@@ -52,7 +52,11 @@ sub Configure
...
+            print $fh "$k " . join(',', @$v) . "\n" unless scalar(@$v) == 0;
```
stdweird: the `scalar(@$v) == 0` is not really needed, yo can do logic tests with `unless @$v`.
but the not printing seems odd. it's not needed for the schema, and if you want this, you should also do this for regular scalars. so the above code can become
```perl
print $fh "$k $value\n" if $value;
```

stdweird: hmmm, this will also skip the print if eg `$v` is 0.
so better to do 
```perl
print $fh "$k $value\n" if length($value);
```

ajf8:
Thanks @stdweird, have made your suggested improvements and tested them.

stdweird:
LGTM

PR #557, opened 2015-08-14
ncm-authconfig: correct module regexp for pamadditions entry attribute

### LINKED ISSUE #529: Quattor 15.4.0 authconfig component problem
Hi,
In Quattor 15.4.0, the authconfig component schema is defined as follows:
```pan
/type authconfig_pamadditions_line_type = {
  "order"       : string with match(SELF, '^(first|last)$')
  "entry"       : string with match(SELF, '\S+\.so$')
};
```

while in 15.2.0, it corresponds to it:
```pan
type authconfig_pamadditions_line_type = {//
  "order"       : string with match(SELF, 'first|last')//
  "entry"       : string with match(SELF, '\S+\.so')//
};
```

What makes the following statement fails:
```pan
"/software/components/authconfig/pamadditions/system"=nlist(
    "conffile","/etc/pam.d/sshd",
    "section","account",
    "lines",list( nlist(
           "order","first",
           "entry","required      pam_access.so accessfile=/etc/security/access_sshd.conf"
    ))
);
```

stdweird:
Fixes #529

### DIFF
--- a/ncm-authconfig/src/main/pan/components/authconfig/schema.pan
+++ b/ncm-authconfig/src/main/pan/components/authconfig/schema.pan
@@ -17,7 +17,7 @@ type yesnostring = string with match(SELF, "yes|no");
 
 type authconfig_pamadditions_line_type = {
   "order"       : string with match(SELF, '^(first|last)$')
-  "entry"       : string with match(SELF, '^\s*\w+\s+\S+\.so(\s|$)')
+  "entry"       : string with match(SELF, '^\s*(required|requisite|sufficient|optional|include|substack)\s+\S+\.so(\s|$)')
 };
 
 type authconfig_pamadditions_type = {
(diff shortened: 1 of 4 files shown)


ncm-authconfig/src/main/pan/components/authconfig/schema.pan
```diff
@@ -17,7 +17,7 @@ type yesnostring = string with match(SELF, "yes|no");
 
 type authconfig_pamadditions_line_type = {
   "order"       : string with match(SELF, '^(first|last)$')
-  "entry"       : string with match(SELF, '\S+\.so$')
+  "entry"       : string with match(SELF, '^\s*\w+\s+\S+\.so(\s|$)')
```
ned21: Is this unnecessarily restrictive? how about 
'^\s*\w+\s+\S+\.so'

\w+ could possibly be restricted to the set of words (e.g. (required|sufficient|etc) allowed in that position in the pam file?

stdweird: the regex you propose is not that much simpler, and allows typos like `required      pam_access.soaccessfile=/etc/security/access_sshd.conf`

stdweird: i'll lookup the allowed words in to replace the `\w+`

stdweird:
@ned21 i added the allowed control values for the entry

PR #558, opened 2015-08-16
ncm-ntpd: introduce useserverip boolean to force hostname to IP

### LINKED ISSUE #531: ncm-ntpd - Use servers names instead of resolved IPs
Hello,

I see in the component code, we translate servers names to IP in order to put them in the ntp.conf (https://github.com/quattor/configuration-modules-core/blob/master/ncm-ntpd/src/main/perl/ntpd.pm#L64).
Using pools (time.belnet.be, be.pool.ntp.org, ...) will resolve an IP from the round-robin DNS, which may lead to issues if the resolved server is down. 
Couldn't we use server names (via an option or by default) in the config file instead of IP's addresses ?

stdweird:
The component will always try to resolve the configured timeservers, but uses the configured hostname instead of the resolved IP address. If one really wants the resolved IP address rather then the hostname, a new boolean `useserverip` is introduced to force this behaviour.

Fixes #531


stdweird:
@jrha this looked easy, so i fixed it. could be pushed to 15.8 if wanted

stdweird:
@ned21 @jrha any remarks?

jrha:
What's with the merge commit?

stdweird:
@jrha is that from pushing the 'update branch' button?

jrha:
I think so...

stdweird:
so we shouldn't press that button? i just asked some people to do so in their PRs

jrha:
I hate merge commits. Also: merging from master into topic is just wrong.

stdweird:
@jrha so pressing the update branch button is a big no-no adn we should only rebase?

jrha:
I would prefer that, yes.

stdweird:
cleaned up

ncm-ntpd/src/main/pan/components/ntpd/config.pan
```diff
@@ -5,5 +5,14 @@
...
+include 'components/${project.artifactId}/schema';
+
+bind "/software/components/${project.artifactId}" = component_ntpd_type;
+
+"/software/packages" = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
+
+prefix '/software/components/${project.artifactId}';
+'active' ?= true;
+'dispatch' ?= true;
+'version' ?= '${no-snapshot-version}';
```
ned21: What's the purpose of this attribute? I thought it was to trigger a re-run of the component after the version changed but I recently discovered that ncm-cdispd watches /software/packages/ncm-$component and re-runs any components that has been upgraded. Is this still useful or should be it be dropped?

stdweird: it will become useful with https://github.com/quattor/ncm-ncd/issues/41
we can remove all version attributes now, and readd them later

ned21: If we have a use for them then leave them in. Just trying to make sure we have a consistent view of what is required in this files!

PR #561, opened 2015-08-19
ncm-ccm: add tabcompletion option

stdweird:
Allows ncm-ccm to enable tabcompletion generation, which is disabled by default in CCM


jrha:
LGTM

PR #562, opened 2015-08-20
ncm-fstab: add protected filesystems and strict option

kwaegema:
Should not be merged before https://github.com/quattor/ncm-lib-blockdevices/pull/52


ncm-fstab/src/main/pan/components/fstab/schema.pan
```diff
@@ -8,15 +8,29 @@ include {'quattor/schema'};
...
+type fstab_protected_entries = {
+    "mounts" : string[] = list (
 	    "/", "/usr", "/home", "/opt", "/var", "/var/lib", "/var/lib/rpm",
 	    "/usr/bin", "/usr/sbin", "/usr/lib", "/usr/local/bin", "/usr/lib64",
 	    "/bin", "/sbin", "/etc", "swap", "/proc", "/sys", "/dev/pts", "/dev/shm",
 	    "/media/floppy", "/mnt/floppy", "/media/cdrecorder", "/media/cdrom",
 	    "/mnt/cdrom", "/boot"
-	)
+    )
+    "filesystems" ? string[]
```
ned21: What's the difference between this field and mounts?

ncm-fstab/src/main/pan/components/fstab/schema.pan
```diff
@@ -8,15 +8,29 @@ include {'quattor/schema'};
...
+    "strict" : boolean = false
```
ned21: Instead of making this binary, should it instead be two separate lists? never-remove and never-modify?

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +46,42 @@ sub update_entries
     return %mounts;
 }
 
+# build the protected hash from the template
+sub protected_hash 
+{
+    my ($self, $config) = @_;
+
+    my $mounts_depr = $config->getElement("/software/components/fstab/protected_mounts")->getTree();
```
ned21: This element in the tree is optional - what happens if it is not specified? Do you get a fatal caused by undef->getTree()?

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +46,42 @@ sub update_entries
...
+    my $protect = $config->getElement("/software/components/fstab/protected")->getTree();
```
ned21: It's preferable to use $self->prefix instead of hardcoding paths.

ncm-fstab/src/main/perl/fstab.pm
stdweird: or use 
```perl
my $tree = $config->getTree($self.prefix());
$mounts_depr = $tree->{proctected_mounts}
```
should not give you errors

ned21:
It would be nice if this component also had a configure.t test. I think the only thing needed to make that possible is to replace [this line](https://github.com/kwaegema/configuration-modules-core/blob/protect/ncm-fstab/src/main/perl/fstab.pm#L138) with something that supports NoAction and then set the variable that indicates NoAction mode is supported.

ncm-fstab/src/main/pan/components/fstab/schema.pan
kwaegema: these are the filesystem types that are protected, not the mointpoints itself

ncm-fstab/src/main/pan/components/fstab/schema.pan
kwaegema: @ned21 Can we use positive names for these, eg. `keep` instead of `never-remove` and `static` instead of `never-modify` ?

ned21: Yes  - sorry the names were terrible. Pick better ones. :-)

ncm-fstab/src/main/pan/components/fstab/schema.pan
ned21: OK, so if something in this array appears in the fs_spec column of fstab then that line will be protected? If so, please describe that in the documentation block.

kwaegema: it is the 3rd field, the fs_vfstype. We use this for our gpfs filesystems. I will document this..

kwaegema:
@ned21 @stdweird  I've made a configure.t, but got into an issue: ncm-fstab uses NCM:Filesystem [here](https://github.com/kwaegema/configuration-modules-core/blob/protect/ncm-fstab/src/main/perl/fstab.pm#L41) 
I could mock this, but then I can't test a lot. 
It would be nice to have the integration with ncm-lib-blockdevices tested I guess, but this adds a dependency on ncm-lib-blockdevices for the test of ncm-fstab to work. Is this an issue?

stdweird:
no, go ahead. i think this is already the case for running AII unittests. i'll update jenkins to include the ncm-lib-blockdeivces path

stdweird:
it was already the case on jenkins

ncm-fstab/src/main/pan/components/fstab/schema.pan
ned21: Oh - thanks for clarifying that, I had completely mis-interpreted the name. The docs make sense but would fs_types be a better name for this field?

ncm-fstab/src/main/pan/components/fstab/schema.pan
```diff
@@ -8,15 +8,37 @@ include {'quattor/schema'};
...
-	)
+    )
+    "filesystems" ? string[]
+};
+
+@documentation{
+fstab component structure
+keep entries are always kept, but can be changed
+static entries can not be changed, but can be deleted
+protected_mounts is still here for backwards compability, and is the same as keep/moints
```
ned21: s/moints/mounts/

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -38,7 +38,7 @@ sub update_entries
         my $el2 = $el->getNextElement();
         my $fs = NCM::Filesystem->new ($el2->getPath()->toString(), $config);
         $self->debug (4, "Update fstab entry at $fs->{mountpoint}");
```
ned21: debug lvl 4 is high. Should this be changed to verbose or even info?

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -110,7 +141,7 @@ sub remount_everything
     while (my $f = <$fstab>) {
     	my $mount = $self->mount_from_entry($f);
     	$mount && $mount ne 'swap' or next;
-    	LC::File::makedir ($mount);
+    	LC::Check::directory ($mount);
```
ned21: Thanks for making this change. Can you confirm this now works safely under NoAction mode and then add the NoActionSupported attribute?

ncm-fstab/src/test/perl/configure.t
```diff
@@ -0,0 +1,46 @@
...
+my $fh = get_file('/etc/fstab');
+like($fh, qr{^/dev/mapper/vg_sl65-lv_root\s+/\s+ext4\s+defaults\s+1\s+1\s*[^/]*$}m, 'root not changed');
+like($fh, qr{^UUID=f6452f58-99b1-41fe-9840-f688157171f8\s+/boot\s+ext4\s+defaults\s+1\s+2\s*[^/]*$}m, '/boot not changed');
+like($fh, qr{^/dev/mapper/vg_sl65-lv_swap\s+swap\s+swap\s+defaults\s+0\s+0\s*[^/]*$}m, 'swap mount ok');
+like($fh, qr{^sysfs\s+/sys\s+sysfs\s+defaults\s+0\s+0\s*[^/]*$}m, '/sys mount ok');
+like($fh, qr{^/dev/gpfsfs\s+/gpfs/fs1\s+gpfs\s+defaults\s+0\s+0\s*[^/]*$}m, 'GPFS mount ok');
+like($fh, qr{^10.10.10.10:6789:/\s+/cephfs\s+ceph\s+name=admin\s+0\s+0\s*[^/]*$}m, 'Mount /cephfs not deleted');
+like($fh, qr{^/dev/sda3\s+/new\s+ext4\s+auto\s+0\s+0\s*[^/]*$}m, 'Mount /new entry added to fstab');
+like($fh, qr{^LABEL=FRIETJES\s+/food\s+chokotoFS\s+auto\s+0\s+0\s*[^/]*$}m, 'Mount /food entry added to fstab');
+
```
ned21: Perhaps add a check that the remount command was run?

ncm-fstab/src/main/perl/fstab.pm
kwaegema: @ned21 $fh->close() is only called when no filehandle is passed, so this won't work here. Unless a new fh is created each call..

ncm-fstab/src/main/perl/fstab.pm
kwaegema: @ned21 ncm-fstab itself  seem to work safe under NoAction mode, using CAF or LC::Check. But is depends on ncm-lib-blockdevices.. I don't know if someone has checked this for NoAction..

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +47,48 @@ sub update_entries
...
+    my ($self, $config) = @_;
+
+    my $tree = $config->getTree($self->prefix());
+    my $mounts_depr = $tree->{protected_mounts}; 
+    
+    my $protected = {};
+
+    foreach my $type ('keep', 'static'){
+        my $protect = $tree->{$type};
+        my %mounts=();
```
stdweird: space around `=`

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +47,48 @@ sub update_entries
...
+        if($type eq 'keep' && $mounts_depr){
```
stdweird: maybe
```perl
my $mountlist = ($type eq 'keep' && $mounts_depr) ? $mounts_depr : $protect->{mounts};
my %mounts = map { $_ => 1 } @$mountlist;
```

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +47,48 @@ sub update_entries
...
 sub valid_mounts
 {
-    my ($self, $config, %mounts) = @_;
+    my ($self, $protected, $fstab, %mounts) = @_;
 
-    my $t = $config->getElement("/software/components/fstab/protected_mounts")
-        ->getTree();
-    foreach my $i (@$t) {
-        $mounts{$i} = 1;
+    @mounts{ keys %{$protected->{mounts}} } = values %{$protected->{mounts}};
```
stdweird: add a comment `# update mounts with protected mounts`

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +47,48 @@ sub update_entries
...
+            mounts => \%mounts,
+            fs_types => \%fs_types,
+        };
+    }
+    return $protected;
+}
+
+
 # Returns a hash with all the mount points that are accepted in the
 # system as keys.
```
stdweird: add that `fstab` hast to be a FileFamily instance

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -46,17 +47,48 @@ sub update_entries
...
 
-    my $t = $config->getElement("/software/components/fstab/protected_mounts")
-        ->getTree();
-    foreach my $i (@$t) {
-        $mounts{$i} = 1;
+    @mounts{ keys %{$protected->{mounts}} } = values %{$protected->{mounts}};
+    my $txt = "$fstab";
+    my $re = qr!^\s*([^#\s]\S+)\s+(\S+?)\/?\s+(\S+)\s!m;
+    while ($txt =~m/$re/mg) {
+        @mounts{$2} = 1 if ($protected->{fs_types}->{$3});
```
stdweird: add a comment clarifying what this does

stdweird:
LGTM
@kwaegema i guess unittests fail because the ncm-libblockdevices PR isn't merged yet?

kwaegema:
@stdweird indeed

ncm-fstab/src/main/perl/fstab.pm
ned21: ah - good point. So I think the wording of this message should be changed to "Checking" not "updating". Did we ever agree on what levels of debug should be used? The fact that /etc/fstab has been modified will get logged when we fix the logging in CAF::FileWriter to say what files it touched.

stdweird: https://github.com/quattor/CAF/blob/master/src/main/perl/Reporter.pm#L82

kwaegema:
test this please

stdweird:
retest this please

stdweird:
retest this please

stdweird:
@kwaegema can you rebase? and confirm we have this running?

kwaegema:
@stdweird rebased. We are indeed running this

kwaegema:
rebased again

ncm-fstab/src/main/pan/components/fstab/schema.pan
```diff
@@ -8,15 +8,37 @@ include {'quattor/schema'};
...
 include {'quattor/filesystems'};
 
-type structure_component_fstab = {
-	include structure_component
-	"protected_mounts" : string[] = list (
+@documentation{ 
+Protected mountpoints and filesystem types. 
+mounts is looked for on the second field of fstab, fs_file
+fs_types is looked for on the third field of fstab, fs_vfstype
+Default mounts is the same list as before in "protected_mounts"
```
ned21: "before" what?

ned21:
I see the dependent block-device change has been merged so I think with one tweak to the docs this is OK to go?

jrha:
Indeed

ncm-fstab/src/main/pan/components/fstab/schema.pan
kwaegema: @ned21 yes indeed, I will include that in the doc

PR #563, opened 2015-08-20
ncm-nagios & ncm-icinga: Seperate type definitions

jrha:
Removes duplicate type definitions which allows both components to be included
which gets us one step closer to having unit tests pass on template-library-core.

### DIFF
--- a/ncm-icinga/src/main/pan/components/icinga/schema.pan
+++ b/ncm-icinga/src/main/pan/components/icinga/schema.pan
@@ -5,7 +5,8 @@
 
 declaration template components/icinga/schema;
 
-include 'quattor/schema';
+include 'quattor/types/component';
+include 'pan/types';
 
 # Please note that the "use" directive is not supported in order to make
 # validation code easier. If you want hosts to inherit settings then use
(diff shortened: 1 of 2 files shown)


ncm-icinga/src/main/pan/components/icinga/schema.pan
```diff
@@ -5,52 +5,52 @@
 
 declaration template components/icinga/schema;
 
-include {'quattor/schema'};
+include 'quattor/schema';
```
stdweird: can probably be replaced with 
```pan
include 'quattor/types/component';
include 'pan/types';
```

ncm-nagios/src/main/pan/components/nagios/schema.pan
```diff
@@ -4,408 +4,408 @@
 
 declaration template components/nagios/schema;
 
-include {'quattor/schema'};
+include 'quattor/schema';
```
stdweird: same remark
```pan
include 'quattor/types/component';
include 'pan/types';
```

stdweird:
LGTM
@jrha as this has no uniitests, i guess we can merge this and test this on templates with and rc of template-library-core.

stdweird:
LGTM

PR #564, opened 2015-08-21
ncm-spma: purge_rep_list - Do nothing if SELF is not defined

jrha:
Needed to get unit-tests passing on `template-library-core`

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/functions.pan
+++ b/ncm-spma/src/main/pan/components/spma/functions.pan
@@ -101,8 +101,10 @@ function resolve_pkg_rep = {
     arg = repository list
 }
 function purge_rep_list = {
-    foreach (i;rep;SELF) {
-        rep['contents'] = null;
+    if (is_defined(SELF)) {
+        foreach (i;rep;SELF) {
+            rep['contents'] = null;
+        };
     };
     SELF;
 };


ncm-spma/src/main/pan/components/spma/functions.pan
```diff
@@ -101,8 +101,10 @@ function resolve_pkg_rep = {
     arg = repository list
 }
 function purge_rep_list = {
-    foreach (i;rep;SELF) {
-        rep['contents'] = null;
+    if (is_list(SELF)) {
```
stdweird: and how about the 
```pan
else {
   deprecated("lists don't do anything anymore");
```
or better yet use `error` (why would you use a function named `_list` on a non-list)

jrha: The primary trigger for this is that the repository list is undefined when running the unit-tests.

stdweird: then do a `if defined(SELF)` instead of `is_list` (the current code will happily take string, dicts and what not and ignore it. i'd rather have it fail in those situations then just do nothing.

jrha: Fair enough, but type checking is what the schema is for.

stdweird:
LGTM

PR #565, opened 2015-08-23
ncm-spma: support yum noaction

### LINKED ISSUE #536: ncm-spma: NoAction and /etc/yum.d issues
[sub cleanup_old_repos](https://github.com/quattor/configuration-modules-core/blob/master/ncm-spma/src/main/perl/spma/yum.pm#L77) executes a change to the filesystem even in NoAction mode.

Guarding the [call point](https://github.com/quattor/configuration-modules-core/blob/master/ncm-spma/src/main/perl/spma/yum.pm#L794) with a ```if (!$NoAction)``` seems to break the yum-configure.t unit test (I have not debugged this in depth but it appears that $NoAction is not set when running under unit tests).

sub generate_repos does respect NoAction but this can lead to mis-leading results. --noaction appears to show now change to the installed packages, but when run without --noaction there is an update to /etc/yum.d and package changes are performed. 

A related issue, that may require it's own ticket, is that when userpkgs = yes then /etc/yum.d is not cleared of invalid repositories. While this is a useful configuration, it should really be its own setting, user-repositories.

(Meta: is it time to move the management of /etc/yum.d/* to its own component to allow finer grained control over yum configuration versus installed package sets?)

stdweird:
Fixes #536


stdweird:
@jrha wrt the 15.8 milestone, this code is still lacking unittests and actual testing.

stdweird:
@ned21 @jnovy this glorified patchwork is what i came up with. can't wait for proper rewrite of spma-yum :smile:

stdweird:
@ned21 noaction unittest added. but even more patchwork.

stdweird:
@ned21 the current code seems to do as expected.
@jrha as there probably will be some remarks, don't delay 15.8 for this (i'm off till sept 3rd)

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -22,9 +22,15 @@ type spma_yum_plugin_versionlock = {
     'follow_obsoletes' ? boolean
 };
 
+type spma_yum_plugin_priorities = {
+    'enabled' : boolean = true
```
ned21: Is being enabled by default backwards compatible with older SL releases?

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -753,44 +855,191 @@ sub update_pkgs_retry
...
+        # Make sure the returned path ends with a /
+        $tmppath .= "/" if ($tmppath !~ m/\/$/);
+
+        return if (! $self->_copy_files_and_dirs($tmppath, @data));
+
+        $self->verbose("Created noaction prefix $tmppath.");
+
+    } else {
+        # Do nothing, return empty string is no noaction
+        $self->verbose("Nothing to do for noaction prefix.");
```
ned21: Is verbose correct for this? Seems like a developer level message (i.e. debug)?

ncm-spma/src/main/pan/components/spma/yum/schema.pan
stdweird: `ncm-spma` has a dependency on `yum-priorities` package, which ships a conf file that enables the plugin, so i'd say this is backwards compatible.

stdweird:
@ned21 remarks addressed

ned21:
LGTM

stdweird:
cleaned up master merge commit

jrha:
LGTM

ncm-spma/src/test/resources/noaction.pan
```diff
@@ -0,0 +1,19 @@
...
+prefix "/software/packages";
+
+"ConsoleKit/_2e4_2e1_2d3_2eel6/arch/x86_64" = "sl620_x86_64";
+
+prefix "/software/repositories/0";
+
+"name" = "sl620_x86_64";
+"owner" = "me@here.com";
+"protocols/0/name" = "http";
+"protocols/0/url" = "http://www.here.com";
```
jrha: Use `www.example.com` as per RFC2606.

stdweird:
@jrha fixed

PR #566, opened 2015-08-24
ncm-iptables: use secure temporary file from File::Temp

ajf8:
Hi,

We were reviewing the changes to ncm-iptables as we need to upgrade, and found usage of an insecure temporary file. Here's a PR to use a file created by File::Temp instead.

### DIFF
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -18,7 +18,7 @@ use vars qw(@ISA $EC);
 $EC=LC::Exception::Context->new->will_store_all;
 use NCM::Check;
 use File::Copy;
-use File::Path;
+use File::Temp qw(tempfile);
 use LC::Process qw(run);
 use LC::Check;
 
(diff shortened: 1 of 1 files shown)


ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -852,16 +852,9 @@ sub Configure($$@) {
     ######################################################################
     # Create tmpdir if necessary
     ######################################################################
-    my $tmpdir = "/var/tmp/ncm/";
-    mkpath($tmpdir, 0, 0755 ) unless ( -e $tmpdir );
+    my ($iptc_temp_fh, $iptc_temp) = tempfile();
```
stdweird: has to use `eval{}` to catch any failures (or the following line makes no sense).

stdweird: what are the default file permissions set by tempfile? and why not specify a template (like `ncm-iptables-XXXXX`)

ajf8:
I've just checked, since the docs don't seem to mention it explicitly, it's 600 as you would expect.

Fixed the template and error checking.

stdweird:
LGTM

PR #569, opened 2015-08-28
ncm-iptables: move from config-{common,rpm}.pan into config.pan

ajf8:


stdweird:
@ajf8 looks like a bug in the old code, but there's no predependency on spma for this component? can you add one?

ajf8:
Sure @stdweird, I've just added that in.

ajf8:
Done.

stdweird:
LGTM
@ajf8 thanks

stdweird:
@ajf8 update branch action required :smile:

stdweird:
@ajf8 please ignore my remark on pressing the update branch button, in fact do never press that evil shiny green button, rebase instead

ajf8:
Got it, I went through all my open PRs and rebased all the branches hopefully :)

stdweird:
@ajf8 this one however is not rebased (you updated the master branch before rebasing right?)

here's a function that should work, it assumes upstream is the name name of the upstream repo etc etc (and that you are using bash)

@jrha should we put this somewhere?

```bash
function rebase_on_master {
    local curr upstream origin
    origin=origin # remote repo name with my fork
    upstream=upstream # remote repo of project
    curr=$(git symbolic-ref -q --short HEAD)
    if [ $? -gt 0 ]; then
        echo 'Failed to determine current branch'
        return 1
    fi
    git checkout master
    if [ $? -gt 0 ]; then
        echo 'Failed to switch to master branch'
        return 1
    fi
    git pull $upstream master
    if [ $? -gt 0 ]; then
        echo "Failed to pull update master from $upstream"
        return 1
    fi
    git checkout $curr
    if [ $? -gt 0 ]; then
        echo 'Failed to switch back to current branch'
        return 1
    fi
    git rebase master
    if [ $? -gt 0 ]; then
        echo 'Failed to rebase current branch on master'
        return 1
    fi
    git push $origin $curr --force
    if [ $? -gt 0 ]; then
        echo "Failed to push force current branch to $origin"
        return 1
    fi
}
```

jrha:
In the release repo?

Issue #432, opened 2015-03-04
tests fail "randomly" on CAF::FileWriter::DESTROY


wdpypere:
Sometimes tests fail on the destroy. output from useraccess fail:

```
[INFO] --- exec-maven-plugin:1.2:exec (prove) @ useraccess ---
src/test/perl/00-load.t ..... ok
src/test/perl/close.t ....... ok
src/test/perl/configure.t ... ok
src/test/perl/files.t ....... ok
Uncaught exception!!! Calling stack is:
  LC::Exception::throw_error called at /home/wdpypere/quattor-libs/lib/perl/LC/Fatal.pm line 204
  LC::Fatal::stat called at /home/wdpypere/quattor-libs/lib/perl/LC/Check.pm line 883
  LC::Check::file called at /home/wdpypere/quattor-libs/lib/perl/CAF/FileWriter.pm line 195
  CAF::FileWriter::close called at /home/wdpypere/quattor-libs/lib/perl/CAF/FileWriter.pm line 280
  CAF::FileWriter::DESTROY called at src/test/perl/kerberos.t line 0
  (eval) called at src/test/perl/kerberos.t line 0
*** stat(target/test/ssh_url_test): No such file or directory
src/test/perl/kerberos.t .... 
Dubious, test returned 1 (wstat 256, 0x100)
```

output from sysctl fail:
```
src/test/perl/yum-cleanup.t ............... ok
	(in cleanup) Can't call method "error" on an undefined value at /home/wdpypere/quattor-libs/lib/perl/LC/Check.pm line 897 during global destruction.
Uncaught exception!!! Calling stack is:
  LC::Exception::throw_error called at /home/wdpypere/quattor-libs/lib/perl/LC/Fatal.pm line 204
  LC::Fatal::stat called at /home/wdpypere/quattor-libs/lib/perl/LC/Check.pm line 883
  LC::Check::file called at /home/wdpypere/quattor-libs/lib/perl/CAF/FileWriter.pm line 195
  CAF::FileWriter::close called at /home/wdpypere/quattor-libs/lib/perl/CAF/FileWriter.pm line 280
  CAF::FileWriter::DESTROY called at src/test/perl/yum-configure.t line 0
  (eval) called at src/test/perl/yum-configure.t line 0
*** stat(/home/wdpypere/workspace/configuration-modules-core/ncm-spma/target/test/cache/simple/profile.2/profile.xml): No such file or directory
src/test/perl/yum-configure.t ............. 
[...]

PR #577, opened 2015-09-11
ncm-modprobe: close the mocked filehandles in the test 

stdweird:
Avoids DESTROY issue (see #432)


jrha:
As if by magic, a Perl Guru appeared.

PR #583, opened 2015-09-22
Metaconfig example service generic

stdweird:


stdweird:
Closed on urpose, only for example purposes

PR #587, opened 2015-09-25
ncm-spma: Next Generation Yum Module [WIP / early review]

ned21:
We wanted to publish this code for discussion before next week's workshop. Assistance/guidance on how to produce good unit tests and any other improvements would be much appreciated!

(Sorry about the merge conflicts, they and the commit history will get cleaned up in due course!)


ncm-spma/pom.xml
```diff
@@ -27,6 +27,8 @@
   <properties>
     <spma.ips.author>Mark R. Bannister</spma.ips.author>
     <spma.ips.email>Mark.Bannister@morganstanley.com</spma.ips.email>
+    <spma.yum-ng.author>Jindrich Novy</spma.yum-ng.author>
```
stdweird: is it `yum-ng` or `yumng`?

ncm-spma/src/main/pan/components/spma/config.pan
```diff
@@ -8,7 +8,7 @@ unique template components/${project.artifactId}/config;
 variable CONFIG_MODULES_CONFIG_SUFFIX ?= 'rpm';
 
 include { 'components/${project.artifactId}/schema' };
-include { 'components/${project.artifactId}/functions' };
+include { 'components/${project.artifactId}/functions-'+PACKAGER };
```
stdweird: can we create subdirs per flavour? e.g. 
```pan
include { format('components/${project.artifactId}/%s/functions', PACKAGER) };
```
(maybe keep a default around for backwards compatibility)?

or make a `components/spma/ips/functions.pan` which is just a `include components/spma/functions`?

ncm-spma/src/main/pan/components/spma/functions-ips.pan
```diff
@@ -0,0 +1,553 @@
+# ${license-info}
```
stdweird: where does this code come from? (i guess it's the old `functions`?)

ncm-spma/src/main/pan/components/spma/functions-ips.pan
```diff
@@ -0,0 +1,553 @@
+# ${license-info}                                                                                                                                                                                                                            
+# ${developer-info}                                                                                                                                                                                                                          
+# ${author-info}
+
+declaration template components/spma/functions-ips;
+
+variable DEF = "";
+
+############################################################
```
stdweird: if you are touching this, please move these blocks into annotations
```pan
@documentation{
Utility function to determine if a particular child element
exists.  Used in the checking of the repositories.
}
```

ncm-spma/src/main/pan/components/spma/functions-rpm.pan
```diff
@@ -0,0 +1,572 @@
+# ${license-info}
```
stdweird: and this is also new/old code?

ncm-spma/src/main/pan/components/spma/functions-ips.pan
stdweird: what imean is: do you expect us to review this?

ncm-spma/src/main/pan/components/spma/functions-yum.pan
```diff
@@ -0,0 +1,560 @@
+# ${license-info}
```
stdweird: same question

ncm-spma/src/main/pan/components/spma/functions-rpm.pan
stdweird: also is the the `yumng` or old `yum`? (i'm confused with the `-rpm` and the `-yum`, and the lack of `-yumng`)

ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -40,13 +40,15 @@ type SOFTWARE_REPOSITORY_PROTOCOL = {
 type SOFTWARE_REPOSITORY = {
     "enabled" : boolean = true
     "gpgcheck" : boolean = false
+    "mirrorlist" : boolean = false
     "excludepkgs" ? string[]
     "includepkgs" ? string[]
-    "name" ? string with match(SELF, '^[\w-.]+$') # "Repository name"
+    "name" ? string  # "Repository name"
```
stdweird: you really want any char as name? might want to check why the old schema had the restrictions

ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -63,7 +66,7 @@ type component_spma_type = {
     include component_spma_ips
     include component_spma_yum
     "cmdfile"       : string = "/var/tmp/spma-commands" # where to save commands for spma-run script
-    "packager"      : string = "yum" with match (SELF, '^(yum|ips)$') # system packager to be used (yum,ips)
+    "packager"      : string = "yum" with match (SELF, '^(yum|ips|rpm)$') # system packager to be used (yum,ips,rpm)
```
stdweird: `rpm

ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -90,6 +93,6 @@ type component_spma_type = {
 };
 
 bind "/software/components/spma" = component_spma_type;
-bind "/software/groups" = SOFTWARE_GROUP{};
+bind "/software/groups" = SOFTWARE_GROUP;
```
stdweird: backwards incompatible. is this really needed/intended?

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -7,4 +7,10 @@ declaration template components/spma/yum/schema;
 type component_spma_yum = {
     "userpkgs_retry" : boolean = true  
     "fullsearch" : boolean = false  
+    "excludes"      ? string[] # packages to be excluded from metadata
+    "yumconf"       ? string # /etc/yum.conf YUM configuration
+    "whitelist"     ? string[] # packages not shipped by repositories but generated by 3rd party installer
+    "quattor_os_file" ? string # file to write quattor_os_release as confirmation of successful YUM spma pass
+    "quattor_os_release" ? string # string to write to quattor_os_file
+    "suffix"        ? string # suffix of alternative packaging module to include instead of standard one
```
stdweird: looking at the code below, this should be restricted (only allow `ng`?)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
+# #
```
stdweird: use the templates for author/license/...

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+use EDG::WP4::CCM::Element qw(unescape);
+
+use CAF::Process;
+use CAF::FileWriter;
+use Set::Scalar;
+use File::Path qw(mkpath rmtree);
+use LWP::Simple;
+use Text::Glob qw(match_glob);
+
+use constant HOSTNAME            => "/system/network/hostname";
```
stdweird: would be nice to be able to indentify the common ones with "regualr" yum, instead of redefining them.

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+use constant REPOS_DIR           => "/etc/yum.repos.d";
+use constant REPOS_TREE          => "/software/repositories";
+use constant PKGS_TREE           => "/software/packages";
+use constant GROUPS_TREE         => "/software/groups/names";
+use constant CMP_TREE            => "/software/components/spma";
+use constant YUM_PACKAGE_LIST    => "/etc/yum/pluginconf.d/versionlock.list";
+use constant YUM_CONF_FILE       => "/etc/yum.conf";
+use constant RPM_QUERY_INSTALLED => qw(rpm -qa --nosignature --nodigest --qf %{EPOCH}:%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n);
+use constant REPO_AVAIL_PKGS     => qw(repoquery -C --show-duplicates --all --qf %{EPOCH}:%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH});
+use constant YUM_PLUGIN_OPTS	 => "--disableplugin=\* --enableplugin=fastestmirror --enableplugin=versionlock";
```
stdweird: no priorities?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+use constant YUM_TEST_CHROOT     => qw(/tmp/spma_yum_testroot);
+use constant SPMA                => "/software/components/spma";
```
stdweird: this should become `$self->prefix()` in the code

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+use constant SPMAPROXY           => "/software/components/spma/proxy";
+
+our $NoActionSupported = 1;
+
+# If user packages are not allowed, removes any repositories present
```
stdweird: can you point out if this old code or new code?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+even upon errors.  In those cases, we have to detect errors by looking
+at stderr.  This method just encapsulates all that logic, keeping the
+callers clean.
+
+=cut
+
+my $yum_warned;
+my $yum_error;
+
+sub execute_yum_command
```
stdweird: is this old code or new code? (ie need reveiw or not?)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+
+    if ( !defined($status) ) {
+	my @lines = split /\n/, $yum_error;
+	foreach my $line (@lines) {
+	    my $cf = index($line, 'conflicts with file from package');
+	    if ( $cf != -1) {
+		my $pkg = $line;
+		$pkg =~ s,^.*package ,,;
+		$pkg =~ s,-[0-9]*:,-,;
+		$self->execute_yum_command(["rpm -e --nodeps $pkg"], "removing $pkg to avoid file conflict");
```
stdweird: the commands should be a arrayref like `['rpm', '-e', '--nodeps', $pkg]`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+sub execute_yum_command_with_recovery
+{
+    my ( $self, $command, $why, $keeps_state, $stdin, $error_logger, $error_ok ) = @_;
+    my $status = $self->execute_yum_command($command, $why, $keeps_state, $stdin, $error_logger, $error_ok);
+    my $packages_modified = 0;
+
+    if ( !defined($status) ) {
+	my @lines = split /\n/, $yum_error;
+	foreach my $line (@lines) {
+	    my $cf = index($line, 'conflicts with file from package');
```
stdweird: you might want a regex here, incase the syntax and/or spelling ever changes

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: youalos don't check exitcode? (or do a forced removal?)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    return $status;
+}
+
+sub Configure
+{
+    my ( $self, $config ) = @_;
+
+    # Make sure stdout and stderr is flushed every time to not to
+    # put mess on serial console
+    autoflush STDOUT 1;
```
stdweird: why is this set here? is/should be part of `ncm-ncd`

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: you might want to document what this is trying to achieve. (because this is not forced, not too many teriible things can happen, but on the otehr hand, removal of packages with dependencies between them might fail if removed in wrong order).

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    my $repos = $config->getElement(REPOS_TREE)->getTree();
+    my $t     = $config->getElement(CMP_TREE)->getTree();
+    
+    # Display system info
+    if ( defined($t->{quattor_os_release}) ) {
+	$self->info("target OS build: ", $t->{quattor_os_release});
+    }
+    
+    my $spmaver = $self->execute_yum_command( ["rpm -q --qf %{NAME}-%{VERSION}-%{RELEASE}.%{ARCH} ncm-spma"], "checking for spma version", 1);
+    $self->info("SPMA version: ", $spmaver);
```
stdweird: any particular reason? would https://github.com/quattor/maven-tools/pull/62 help instead? (all components version should better be logged by ncm-ncd. https://github.com/quattor/ncm-ncd/issues/48)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+
+    # Convert these crappily-defined fields into real Perl booleans.
+    $t->{run}      = $t->{run} eq 'yes';
+    $t->{userpkgs} = defined( $t->{userpkgs} ) && $t->{userpkgs} eq 'yes';
+
+    # test for sharp spma execution early
+    my $spma_run = 0;
+    if ( -e "/.spma-run" ) {
+        if ( !unlink "/.spma-run" ) {
+            $self->warn("unable to remove file /.spma-run: $!");
```
stdweird: only warn, no error or return?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    $self->info("SPMA version: ", $spmaver);
+    $self->info("User packages permitted: $t->{userpkgs}");
+
+    # Convert these crappily-defined fields into real Perl booleans.
+    $t->{run}      = $t->{run} eq 'yes';
+    $t->{userpkgs} = defined( $t->{userpkgs} ) && $t->{userpkgs} eq 'yes';
+
+    # test for sharp spma execution early
+    my $spma_run = 0;
+    if ( -e "/.spma-run" ) {
```
stdweird: location of file must be a constant (and all things quattor are better tracked in e.g. `/var/run/quattor-components/`).

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        if ( !unlink "/.spma-run" ) {
+            $self->warn("unable to remove file /.spma-run: $!");
+        }
+	$spma_run = 1;
+    }
+
+    # Detect legacy OS (RHEL5 and older)
+    my $legacy_os = 0;
+    my $fhi;
+    if ( open( $fhi, '<', "/etc/redhat-release" ) ) {
```
stdweird: use `CAF::FileReader`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        while ( my $line = <$fhi> ) {
+            if ( index( $line, 'release 5' ) >= 0 ) {
```
stdweird: should be a regex.

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    # test for sharp spma execution early
+    my $spma_run = 0;
+    if ( -e "/.spma-run" ) {
+        if ( !unlink "/.spma-run" ) {
+            $self->warn("unable to remove file /.spma-run: $!");
+        }
+	$spma_run = 1;
+    }
+
+    # Detect legacy OS (RHEL5 and older)
```
stdweird: should be a method, and if you claim `older`, you should also support that (the current code only handles `5` imho)

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -7,4 +7,10 @@ declaration template components/spma/yum/schema;
 type component_spma_yum = {
     "userpkgs_retry" : boolean = true  
     "fullsearch" : boolean = false  
+    "excludes"      ? string[] # packages to be excluded from metadata
+    "yumconf"       ? string # /etc/yum.conf YUM configuration
```
stdweird: a "string" is hardly valid yum configuration

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    print $yum_conf_file "exclude=" . join ( " ", sort @$excludes );
+    $yum_conf_file->close();
+
+    if ( !$NoAction ) {
+        # Remove all repositories if userpkgs is not defined
+        $self->cleanup_old_repos( REPOS_DIR, $repos, $t->{userpkgs} ) or return 0;
+    }
+
+    # Generate new installation repositories from host profile
+    my $proxy = !defined($t->{proxy}) ? 'yes' : $t->{proxy};
```
stdweird: why not use `0` and `1`?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+			$url =~ s,[^/]*/,,;
+			$url = $repo->{proxy} . $url;
+		    }
+		}
+	    }
+	}
+	if (!defined($urls)) {
+	    $urls = $url;
+	}
+        my $repofile = "[$repo->{name}]\nname=$repo->{name}\nbaseurl=$urls\nenabled=$repo->{enabled}\ngpgcheck=$repo->{gpgcheck}\n";
```
stdweird: use TextRender?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        if ( defined( $repo->{mirrorlist} ) && $repo->{mirrorlist} ) {
+            $repofile = "[$repo->{name}]\nname=$repo->{name}\nmirrorlist=$url\nenabled=$repo->{enabled}\ngpgcheck=$repo->{gpgcheck}\n";
+        }
+        print $fh "# File generated by ", __PACKAGE__, ". Do not edit\n";
+        print $fh $repofile;
+        $fh->close();
+    }
+
+    # Preprocess required packages and separate version-locked
+    #    - also skip package dualities - e.g. both kernel and kernel-2.6.32-504.1.3.el6.x86_64
```
stdweird: is this an issue in regular spma yum?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    my $pkgs               = $config->getElement(PKGS_TREE)->getTree();
+    my $wanted_pkgs        = Set::Scalar->new();
+    my $wanted_pkgs_locked = Set::Scalar->new();
+    my $found_spma         = 0;
+    my @pkl;
+    my @pkl_v;
+    my @pkl_a;
+
+    for my $name ( keys %$pkgs ) {
+        if ( !$found_spma && substr( unescape $name, 0, 8 ) eq 'ncm-spma' ) {
```
stdweird: regexp will be more readbale than `substr`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+            $found_spma = 1;
+        }
+        my $vra = $pkgs->{$name};
+        while ( my ( $vers, $a ) = each(%$vra) ) {
```
stdweird: `each` is evil. https://github.com/quattor/CAF/issues/109

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+            my $arches = $a->{arch};
+            if ( exists( $a->{repository} ) ) {
+                foreach my $arch (@$arches) {
+                    if ( $vers ne '_' ) {
```
stdweird: since you have the `else` block too, use positive conditions like

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    #      will install the latest what is undesired. Simply keep only version-locked variant.
+    my $pkgs               = $config->getElement(PKGS_TREE)->getTree();
+    my $wanted_pkgs        = Set::Scalar->new();
+    my $wanted_pkgs_locked = Set::Scalar->new();
+    my $found_spma         = 0;
+    my @pkl;
+    my @pkl_v;
+    my @pkl_a;
+
+    for my $name ( keys %$pkgs ) {
```
stdweird: sort the keys fro repoducibility

stdweird: better to call this vairbale `$escname` and do `my $name = unescape($escname);` as first thing in the loop

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+                        } else {
+                            push ( @pkl_a, unescape $name . ';' . $arch );
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    if ( !$found_spma ) {
```
stdweird: maybe use
```perl
if (! $pkgs->{escape('ncm-spma')}) {
  ...
}
```
before the above loop (and remove the logic looking for spma in the loop)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        $self->error('Package ncm-spma is not present among required packages.');
+        return 0;
+    }
+
+    my $wanted_pkgs_uv = Set::Scalar->new(@pkl);
```
stdweird: this looks like something regular yum spma could use too

stdweird: can you add a comment for each set describing what sort of packages it holds?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    my $wanted_pkgs_v  = Set::Scalar->new(@pkl_v);
+    my $wanted_pkgs_a  = Set::Scalar->new(@pkl_a);
+    while ( defined( my $p = $wanted_pkgs_uv->each ) ) {
+        while ( defined( my $p2 = $wanted_pkgs_v->each ) ) {
+            if ( index( $p2, $p ) == 0 ) {
```
stdweird: isn't this a bit too loose? any package that is a prefix will match.

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        if ( !$found_spma && substr( unescape $name, 0, 8 ) eq 'ncm-spma' ) {
+            $found_spma = 1;
+        }
+        my $vra = $pkgs->{$name};
+        while ( my ( $vers, $a ) = each(%$vra) ) {
+            my $arches = $a->{arch};
+            if ( exists( $a->{repository} ) ) {
+                foreach my $arch (@$arches) {
+                    if ( $vers ne '_' ) {
+                        push ( @pkl_v, ( unescape $name) . ';' . ( unescape $vers) . '.' . $arch );
```
stdweird: why is this joined as string and not a 2 element tuple?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        $wanted_pkgs_locked->insert($p);
+    }
+    while ( defined( my $p = $wanted_pkgs_a->each ) ) {
+        $p =~ s/;/./;
+        $wanted_pkgs->insert($p);
+    }
+
+    if ( !$NoAction ) {
+        # Import GPG keys
+        foreach my $file ( glob "/etc/pki/rpm-gpg/RPM-GPG-KEY*" ) {
```
stdweird: where do these come from?
(also interesting for regular yum spma)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+            my $gpg = $self->execute_yum_command( ["rpm -v --import $file"], "importing GPG key $file" );
+            return 0 if !defined($gpg);
+        }
+    }
+
+    # Get list of packages installed on system before any package modifications.
+    my $preinstalled_rpms = $self->execute_yum_command( [RPM_QUERY_INSTALLED], "getting list of installed packages", 1 );
+    return 0 if !defined($preinstalled_rpms);
+    $preinstalled_rpms =~ s/\(none\)/0/g;
```
stdweird: what does this do?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    my $preinstalled = Set::Scalar->new( split ( /\n/, $preinstalled_rpms ) );
+
+    if ( !$NoAction && $spma_run ) {
+	# Attempt to update YUM and plugins to latest available release.
```
stdweird: also relevant for regular yum spma

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: where do you configure e.g. fastestmirror

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+	my $updateyum = $self->execute_yum_command( ["yum update -y yum\* " . YUM_PLUGIN_OPTS], "updating YUM stack" );
```
stdweird: this command and some of the others look like fixed strings, you also put thos in a constant/readonly

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    my $yum_clean = $self->execute_yum_command( ["yum clean expire-cache " . YUM_PLUGIN_OPTS], "expiring YUM caches", 1 );
+    if ($yum_warned) {
+        # Expiration of cache is not enough sometimes.
+        # https://bugzilla.redhat.com/show_bug.cgi?id=1151074
+        my $yum_clean = $self->execute_yum_command( ["yum clean all " . YUM_PLUGIN_OPTS], "resetting hard YUM state", 1 );
+        return 0 if !defined($yum_clean);
+    }
+
+    return 0 if !defined($yum_clean);
+    $yum_clean = $self->execute_yum_command( ["yum makecache " . YUM_PLUGIN_OPTS], "generating YUM cache", 1 );
```
stdweird: also usefull in regular yum spma

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    return 0 if !defined($yum_clean);
+
+    my @files = glob "{/tmp/*.yumtx,/var/lib/yum/transaction*}";
```
stdweird: deleting files and dirs without noaction protection?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    foreach my $dir (@dirs) {
+        if ( !rmtree $dir) {
+            $self->info("unable to remove directory $dir: $!");
+        }
+    }
+
+    # Test whether comps/groups are sane.
+    {
+	$yum_warned = 0;
+	my $groupinfo = $self->execute_yum_command( [ "yum groupinfo core " . YUM_PLUGIN_OPTS ], "testing comps/groups sanity", 1 );
```
stdweird: also regular yum spma

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+	    return 0;
+	}
+    }
+
+    # Query metadata for version locked packages including Epoch and write versionlock.list
+    my $repodata_rpms = $self->execute_yum_command( [REPO_AVAIL_PKGS], "fetching full package list", 1 );
+    return 0 if !defined($repodata_rpms);
+
+    # Test whether locked packages are present in the metadata
+    my $repoquery_list = Set::Scalar->new( reverse(split ( /\n/, $repodata_rpms)) );
```
stdweird: why the reverse?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+	$yum_warned = 0;
+	my $groupinfo = $self->execute_yum_command( [ "yum groupinfo core " . YUM_PLUGIN_OPTS ], "testing comps/groups sanity", 1 );
+	if ( $yum_warned ) {
+	    $self->error("Groups are not sane - core group missing. Will not continue.");
+	    return 0;
+	}
+    }
+
+    # Query metadata for version locked packages including Epoch and write versionlock.list
+    my $repodata_rpms = $self->execute_yum_command( [REPO_AVAIL_PKGS], "fetching full package list", 1 );
```
stdweird: why all packages? i just ran this on a compute node, this is 52k rpms (ok, we need to cleanup our repos a bit, but still)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    return 0 if !defined($repodata_rpms);
+
+    # Test whether locked packages are present in the metadata
+    my $repoquery_list = Set::Scalar->new( reverse(split ( /\n/, $repodata_rpms)) );
+    my $repoquery_list_noepoch = Set::Scalar->new();
+    my $locked_found           = Set::Scalar->new();
+    my $locked_found_noepoch   = Set::Scalar->new();
+    $self->info( "(" . $repoquery_list->size . " total packages)" );
```
stdweird: each info is a new entry in the log, the message shouldn't start with `(`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    while ( defined( my $p = $repoquery_list->each ) ) {
+        my $t = $p;
+        $t =~ s/^.*://;
+        if ( !$repoquery_list_noepoch->has($t) ) { $repoquery_list_noepoch->insert($t); }
+        if ( $wanted_pkgs_locked->has($t) ) {
```
stdweird: how does this code handle globs in package names and/or version?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+                $locked_found_noepoch->insert($t);
+            }
+        }
+    }
+    {
+        my $fh = CAF::FileWriter->new( YUM_PACKAGE_LIST, log => $self );
+        print $fh join ( "\n", @$locked_found );
+        $fh->close();
+        if ( $wanted_pkgs_locked->size != $locked_found->size ) {
+            $self->error( "Version-locked packages are missing from repositories - expected ", $wanted_pkgs_locked->size, ", available ", $locked_found->size );
```
stdweird: there should be only one error logged here (either on big message, or one error with the summary and e.g 1 or 2 warn with the rest of the data).
also the `missing` is for the case `$wanted_pkgs_locked->size > $locked_found->size`, not for the other inequality (although i don't know how that could occur)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+
+    # Test also whether version unlocked packages are present in repositories.
+    {
+	my $found = Set::Scalar->new();
+	while ( defined( my $r = $repoquery_list->each ) ) {
+	    my $t = $r;
+	    $t =~ s/^.*://;
+	    my $name = $t;
+	    my $arch = substr($name, rindex($t, '.'));
+	    while( (my $end = rindex($name, '-')) != -1) {
```
stdweird: a regexp might help (and make clear what this is). i also don't know what is so special about `-`. (the first `-` from the right indicate the start of the version?)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+		    last;
+		}
+		if ( $wanted_pkgs->has("$name") ) {
+		    $found->insert("$name");
+		    last;
+		}
+	    }
+	}
+	if ( $found->size != $wanted_pkgs->size ) {
+            $self->error("Requested packages are missing from repositories.");
```
stdweird: same remark wrt 2 errors

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        return 1 if !$t->{run};
+    }
+
+    # Run test transaction to get complete list of packages to be present on the system
+    my $groups	   = $config->getElement(GROUPS_TREE)->getTree();
+    my $clean_root = $self->execute_yum_command( [ "rm -rf " . YUM_TEST_CHROOT ], "cleaning YUM test chroot", 1 );
+    $clean_root = $self->execute_yum_command( [ "mkdir -p " . YUM_TEST_CHROOT . "/var/cache" ],                 "setting up YUM test chroot",    1 );
+    $clean_root = $self->execute_yum_command( [ "ln -s /var/cache/yum " . YUM_TEST_CHROOT . "/var/cache/yum" ], "setting YUM test chroot cache", 1 );
+    my $yum_install_test_command;
+    $yum_install_test_command = "yum install " . YUM_PLUGIN_OPTS . " -q -C --installroot=" . YUM_TEST_CHROOT;
```
stdweird: this should be an arrayref

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    if (@$groups)             { $yum_install_test_command .= " @" . join   ( " @",   sort @$groups ); }
```
stdweird: and you should just push these to the arrayref

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    if (@$wanted_pkgs_locked) { $yum_install_test_command .= " " . join    ( " ",    sort @$wanted_pkgs_locked ); }
+    if (@$wanted_pkgs)        { $yum_install_test_command .= " " . join    ( " ",    sort @$wanted_pkgs ); }
+    my $yum_install_test = $self->execute_yum_command( [$yum_install_test_command], "performing YUM chroot install test", 1, "/dev/null", "verbose", 1 );
+    $self->log($yum_install_test_command);
```
stdweird: `log`? no info?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+use constant REPOS_TREE          => "/software/repositories";
+use constant PKGS_TREE           => "/software/packages";
+use constant GROUPS_TREE         => "/software/groups/names";
+use constant CMP_TREE            => "/software/components/spma";
+use constant YUM_PACKAGE_LIST    => "/etc/yum/pluginconf.d/versionlock.list";
+use constant YUM_CONF_FILE       => "/etc/yum.conf";
+use constant RPM_QUERY_INSTALLED => qw(rpm -qa --nosignature --nodigest --qf %{EPOCH}:%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n);
+use constant REPO_AVAIL_PKGS     => qw(repoquery -C --show-duplicates --all --qf %{EPOCH}:%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH});
+use constant YUM_PLUGIN_OPTS	 => "--disableplugin=\* --enableplugin=fastestmirror --enableplugin=versionlock";
+use constant YUM_TEST_CHROOT     => qw(/tmp/spma_yum_testroot);
```
stdweird: should be a tempdir. how big does this become? and is there any IO during it's usage? (in particular, coudl /dev/shm instead of /tmp help?)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    if (@$wanted_pkgs)        { $yum_install_test_command .= " " . join    ( " ",    sort @$wanted_pkgs ); }
+    my $yum_install_test = $self->execute_yum_command( [$yum_install_test_command], "performing YUM chroot install test", 1, "/dev/null", "verbose", 1 );
+    $self->log($yum_install_test_command);
+    return 0 if !defined($yum_install_test);
+    $self->log($yum_install_test);
+    my $clean_root_after = $self->execute_yum_command( [ "rm -rf " . YUM_TEST_CHROOT ], "removing YUM test chroot", 1 );
+
+    # Parse YUM output to get full package list
+    my $to_install = Set::Scalar->new;
+    if ( !$legacy_os ) {
```
stdweird: you have the `else` block, use positive condition

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        # RHEL6+ falls here - we don't support anything older than RHEL5 (inclusive)
+        my $skipped = index( $yum_install_test, "Skipped (dependency problems):" );
```
stdweird: regexp might be better if syntax ever changes

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        if ( $skipped != -1 ) {
+	    $self->info($yum_install_test);
+            $self->error("Dependency problems in test transaction, see log.");
```
stdweird: which log?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+    $self->log($yum_install_test);
+    my $clean_root_after = $self->execute_yum_command( [ "rm -rf " . YUM_TEST_CHROOT ], "removing YUM test chroot", 1 );
+
+    # Parse YUM output to get full package list
+    my $to_install = Set::Scalar->new;
+    if ( !$legacy_os ) {
+        # RHEL6+ falls here - we don't support anything older than RHEL5 (inclusive)
+        my $skipped = index( $yum_install_test, "Skipped (dependency problems):" );
+        if ( $skipped != -1 ) {
+	    $self->info($yum_install_test);
```
stdweird: maybe a more specific message

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        }
+        my @tx_files = glob "/tmp/*.yumtx";
+        if ( scalar( grep { defined $_ } @tx_files ) != 1 ) {
+	    $self->info($yum_install_test);
+            $self->error( "Dependency problems or multiple yumtx files. See log." );
+            return 0;
+        }
+        my $tx_file = $tx_files[0];
+        my $fh;
+        if ( !open $fh, '<', $tx_file ) {
```
stdweird: `CAF::FileReader`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        my $skipped = index( $yum_install_test, "Skipped (dependency problems):" );
+        if ( $skipped != -1 ) {
+	    $self->info($yum_install_test);
+            $self->error("Dependency problems in test transaction, see log.");
+            if ( !defined( $t->{yumtolerant} ) || !( $t->{yumtolerant} eq 'yes' ) ) {
+                return 0;
+            }
+        }
+        my @tx_files = glob "/tmp/*.yumtx";
+        if ( scalar( grep { defined $_ } @tx_files ) != 1 ) {
```
stdweird: so any user creating a `/tmp/my.yumtx` will cause this to fail?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,810 @@
...
+        my $tx_file = $tx_files[0];
+        my $fh;
+        if ( !open $fh, '<', $tx_file ) {
+            $self->error("Unable to open transaction file $tx_file: $!");
+            return 0;
+        }
+        my @lines = grep /^mbr:/, <$fh>;
+        foreach (@lines) {
+            my ( $epoch, $name, $version, $release, $arch, $rest );
+            $_ =~ s/mbr: //;
```
stdweird: what's the format? maybe use a single regexp instead of this split block?

ncm-spma/pom.xml
jnovy: In fact is is 'yumng' without the dash - as the suffix is added from /software/components/spma/suffix which is 'ng' -> yumng.

ncm-spma/src/main/pan/components/spma/config.pan
jnovy: Good plan. We can stash it in the 'yum/ips/rpm' subdir.

ncm-spma/src/main/pan/components/spma/functions-ips.pan
jnovy: Yes please.

This PR isn't supposed to be even considered for inclusion yet. As it's just a first-shot review for now.

jnovy: WRT +# ${license-info} - yes, it came originally from the old 'functions' template.

ncm-spma/src/main/pan/components/spma/functions-ips.pan
jnovy: Same case as in the comment above.

ncm-spma/src/main/pan/components/spma/functions-rpm.pan
jnovy: The functions seem to be the same for the current 'yum' spma plugin as well as for other plugins as 'yumng'. So just value of packager should be sufficient IMO.

ncm-spma/src/main/pan/components/spma/schema.pan
jnovy: Actually I will resurrect the limitation as it makes sense - yum doesn't like '[' or ']' in repo name.

ncm-spma/src/main/pan/components/spma/schema.pan
jnovy: We (MS) are still using rpm packager settings for legacy builds. The 'yumng' is just a subplugin of 'yum' packager. In fact we are using PACKAGER == 'yum' conditionals to check for yum base OS builds and redefining PACKAGER to subplugin would introduce unwanted high granularity. I mean having huge and complicated conditions in templates.

ncm-spma/src/main/pan/components/spma/schema.pan
jnovy: Yes. We decided to not to support optional/default/mandatory group sets as RH tests only with the default set and deps may fail/have interesting consequences for other setups.

ncm-spma/src/main/pan/components/spma/yum/schema.pan
jnovy: Not sure. Is there any advantage in not sharing the schema with ordinary 'yum' spma plugin altogether? I'm not assuming we'd need to split these as they are not supposed to be conflicting. In the end we'll have richer schema for 'yum' packager - is that a problem?

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: In fact there seems to be only 'use Text::Glob qw(match_glob);' and mktree/mkpath new in yumng. Everything else should be same with default yum spma plugin.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Yep, we don't want any repo priorities. As the concept of it is quite badly maintainable. We follow the idea of - get latest available from all enabled repos. Once we need older version then we version-lock to it on template side.

So there is no need for priorities at all.

There is not much needed on fastestmirror side. In fact all is set in *.repo files generated.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Ok, makes sense.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Old code with improvements :)

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Old code with improvements - in fact it needs cleanup (the old code part).

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: The purpose of the recovery feature is that - do at least something if the yum transaction failed based on the log yum spat out.

That means it's on best effort basis - if it fails then OK - SPMA fails as well.

Regarding the rpm -e --nodeps - it's forced remove one package by package. In the yum log there could be more same packages listed - in that case one of such invocations could fail which is OK. In fact it is forced removal (package by package without deps).

Deps are fixed right after in install invocation. So no worries about single package removals.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: We had problems on slow serial consoles that the text output doesn't get flushed to it fully even though the underlying action already finished.

To illustrate - you make a new kickstart build watching it via serial console. spma fails printing a verbose error - before it;s fully printed via serial console the box reboots. So you lose info what actually happenned.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: We update the ncm-spma module shipping yumng plugin we use quite often so we are very interested in minor versions of rpm to be sure the planned fixes are in. To illustrate - we release minor releases with fixes/improvements approximately once per week which is way too more often that quattor releases.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: I've changed it to error, thanks.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: No problem with that. BTW the directory currently contains only zero-byte files with names of some of components. Could you please elaborate what is the real purpose of this directory and what is supposed to be in?

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: It's a workaround for RHEL5 where CAF:FileReader fails. Let's discuss this on Quattor workshop's hackfest.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: In fact we don't support RHEL4 and older any longer. I will adjust the comment.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: I followed the old userpkgs check from the old code. No problem to simplify that.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: In fact $self->prefix() doesn't work, so I will stick with the static path:

[INFO] Errors while configuring spma (1)
[ERROR] 1 errors, 0 warnings executing configure
[ERROR] Uncaught exception!
[ERROR] Calling stack is:
[ERROR] *** cannot resolve element /software/components/yumng

[ERROR] Uncaught exception!
[ERROR] Calling stack is:
        LC::Exception::throw_error called at /opt/edg/lib/perl/EDG/WP4/CCM/Configuration.pm line 249

        EDG::WP4::CCM::Configuration::getElement called at /opt/edg/lib/perl/NCM/Component/spma/yumng.pm line 285

        NCM::Component::spma::yumng::Configure called at /opt/edg/lib/perl/NCM/Component/spma.pm line 74

        NCM::Component::spma::call_entry_point called at /opt/edg/lib/perl/NCM/Component/spma.pm line 33

        NCM::Component::spma::Configure called at (eval 20) line 1

        (eval) called at /opt/edg/lib/perl/NCD/ComponentProxy.pm line 514

        NCD::ComponentProxy::_execute called at /opt/edg/lib/perl/NCD/ComponentProxy.pm line 94

        NCD::ComponentProxy::executeConfigure called at /opt/edg/lib/perl/NCD/ComponentProxyList.pm line 146

        NCD::ComponentProxyList::executeConfigComponents called at /usr/sbin/ncm-ncd line 465

[ERROR] *** EDG::WP4::CCM::Element->createElement (EDG::WP4::CCM::Configuration=HASH(0x2057deb0), EDG::WP4::CCM::Path=ARRAY(0x20dfb220)): Failed to read type of /software/components/yumng

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: It's an (logical) issue with functions.tpl with current spma.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Yes, added.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: It is intentional.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: We use GPG signed repositories and force YUM to verify signatures. These files are installed vya redhat-release for OS and our custom package shipping other GPG keys which we control.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Sometimes the result of YUM transaction is not exactly defined beforehand. E.g yum downgrade with list of packages. The invocation is to get list of packages installed before and after yum transaction to get info what actually changed (and log it). It is useful to also check yum misbehaviour which is obvious from no package changes for example.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: The reason for this is to build the initial cache so that subsequent yum/repoquery calls are fast and cached.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Yes. In fact noaction won't get here as noaction is handled via condition above and ends spma.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: repoquery sorts output version-wise in reverse order.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: spma matches the list with version-locked package requests. So all packages need to be queried. This is fast after yum cache is built.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: The '(' is used to signify a list of packages is going to be printed. This list might be long, hnece the ().

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: We don't support globs in package names - it's pure evil and it's in fact forbidden for our builds.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Yes, start of version.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Yes. The output is huge - several hundreds of lines.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: No packages are getting installed into the test chroot. Just metadata are symlinked there and yum/rpm creates basic rpmdb there. Just hundreds of kBs.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: spma log :)

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: and how many iops? if it is very smal in size, using a tempdir in /dev/shm can allieviate any iops load

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: default fastest mirror uses external mirrors by default, we don't want that :smile:

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: looks like you are making a `ncm-yumng` instead of a `yumng` plugin for spma. this should be fixed

PR #590, opened 2015-09-28
ncm-fstab: this component is NoAction ready but missing declaration

ajf8:


ajf8:
Actually, I missed that this uses LC::File::makedir, which isn't NoAction safe.

We will wait for this PR to be merged, which should provide NoAction support. https://github.com/quattor/configuration-modules-core/pull/562

So I'll close this, sorry for the noise.

Issue #586, opened 2015-09-25
ncm-ceph: ownership of known_hosts file when first talking to other monitors


GRyall:
We have just changed our `deploy_host`, this was relatively simple, but to make the component work we had to change the ownership of `/home/ceph/.ssh/known_hosts` from root to ceph. It wouldn't talk to the other monitors until we did this, is this something that the component could check before trying to add anything to the `known_hosts` file?

jrha:
Fixed by #595, but the fixes line was not written correctly.

PR #595, opened 2015-10-01
ncm-ceph: fix user for known_hosts and fix output for destroy and restart

### LINKED ISSUE #585: ncm-ceph: manual commands for removing monitors
When removing a monitor from my cluster, the component told me this:
```
[INFO] Commands to be run manually:
[INFO] su - ceph -c /usr/bin/ceph-deploy --cluster ceph fqdn destroy mon2
[INFO] su - ceph -c /usr/bin/ceph-deploy --cluster ceph mon destroy mon2
[INFO] su - ceph -c /usr/bin/ceph-deploy --cluster ceph config destroy mon2
```
What I actually needed to run was
```bash
su - ceph -c "/usr/bin/ceph-deploy --cluster ceph mon destroy mon2.example.org"
```

kwaegema:
fixes #585 and #586

### DIFF
--- a/ncm-ceph/src/main/perl/Ceph/commands.pm
+++ b/ncm-ceph/src/main/perl/Ceph/commands.pm
@@ -162,7 +162,10 @@ sub ssh_known_keys {
             $cmd = ['/usr/bin/ssh-keyscan', $host];
             my $key = $self->run_command_as_ceph($cmd);
             my $fh = CAF::FileEditor->open("$cephusr->{homeDir}/.ssh/known_hosts",
-                                           log => $self, owner => $cephusr->{uid}, group => $cephusr->{gid} );
+                                            log => $self,
+                                            owner => $cephusr->{uid},
+                                            group => $cephusr->{gid},
+                                        );
             $fh->head_print($key);
             $fh->close();
         }
--- a/ncm-ceph/src/main/perl/Ceph/commands.pm
+++ b/ncm-ceph/src/main/perl/Ceph/commands.pm
@@ -150,7 +150,7 @@ sub run_ceph_deploy_command {
 
 # Accept and add unknown keys if wanted
 sub ssh_known_keys {
-    my ($self, $host, $key_accept, $homedir) = @_; 
+    my ($self, $host, $key_accept, $cephusr) = @_;
     return 1 if !defined($key_accept);
     if ($key_accept eq 'first'){
         # If not in known_host, scan key and add; else do nothing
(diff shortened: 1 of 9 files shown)


ncm-ceph/src/main/perl/Ceph/commands.pm
```diff
@@ -161,8 +161,8 @@ sub ssh_known_keys {
         if (!$lines) {
```
piojo: Sorry for looking at something that is not in the diff, but whaty's the purpose of this?  It looks like you only care on whether the output is empty?

kwaegema: ($key_accept eq 'first') means: only add the ssh key blindy if it is not already there. If it is there, it will not replace it, it only does that when key_accept is 'always'

piojo: Sorry, github isn't showing you what I meant.  I meant this chunk (expand the context):

```perl  
my $cmd = ['/usr/bin/ssh-keygen', '-F', $host];
my $output = $self->run_command_as_ceph($cmd);
# Count the lines of the output
my $lines = $output =~ tr/\n//;
if (!$lines) {
```

Why do you need to count lines at all?  It looks to me like you're doing something if there is output and something if there isn't.

kwaegema: It needs to trim the newlines, because of the behaviour of ssh-keygen:

[kwaegema@daenerys ~]$ bla=`` `ssh-keygen -F gastlyf.ugent.be` ``
[kwaegema@daenerys ~]$ echo $bla

[kwaegema@daenerys ~]$

ncm-ceph/src/main/perl/Ceph/commands.pm
```diff
@@ -161,8 +161,8 @@ sub ssh_known_keys {
...
             $cmd = ['/usr/bin/ssh-keyscan', $host];
             my $key = $self->run_command_as_ceph($cmd);
-            my $fh = CAF::FileEditor->open("$homedir/.ssh/known_hosts",
-                                           log => $self);
+            my $fh = CAF::FileEditor->open("$cephusr->{homeDir}/.ssh/known_hosts",
+                                           log => $self, owner => $cephusr->{uid}, group => $cephusr->{gid} );
```
stdweird: one attribute per line, end with `,`

stdweird:
@kwaegema minor remark and update branch

stdweird:
rebase instead of update branch

kwaegema:
done

kwaegema:
@stdweird rebased again

jrha:
@gvasilak can you have a look over this?

jrha:
Also @GRyall...

PR #598, opened 2015-10-05
ncm-ccm: do not use stdin for tetsing with ccm-fetch

### LINKED ISSUE #597: ncm-ccm: ccm-fetch cannot handle stdin as configfile
the test of the new cofngifile in ncm-ccm uses ccm-fetch with configfile to test piped via stdin. but this pipe is supposed to be read twice: once by CAF::Application, once by CCM::CCfg, which is not possible

stdweird:
Fixes #597

### DIFF
--- a/ncm-ccm/src/main/perl/ccm.pm
+++ b/ncm-ccm/src/main/perl/ccm.pm
@@ -12,13 +12,17 @@ use CAF::Process;
 use CAF::FileWriter;
 use LC::Exception;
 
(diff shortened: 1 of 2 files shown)


stdweird:
The stdin solution is much more elegant, so if someone fixes the root cause of ccm-fetch, this might get reverted

jrha:
Open an issue to fix the root cause?

stdweird:
@jrha https://github.com/quattor/CCM/issues/74

PR #599, opened 2015-10-06
ncm-metaconfig: service httpd: initial apache 2.4 format support

stdweird:
Converted to versioned tests for 2.2 and 2.4, and support the new `mutex` instead of `sslmutex` keyword.
Requires https://github.com/quattor/maven-tools/pull/65, so new build-tools bump.


wdpypere:
I tried to use this but it fails on a configuration issue. The directive: ```include conf.modules.d/*.conf``` in ```/etc/httpd/conf/httpd.conf``` should come before any ```<ifmodule>``` statement or it will not see these parts of the configuration and use the defaults. 
As it is it comes after.

stdweird:
@wdpypere fixed in https://github.com/stdweird/configuration-modules-core/commit/32aa2246f3eafb50c22217339b0923735a0f883f

PR #600, opened 2015-10-07
ncm-fstab: Include ncm-filesystems functionality

kwaegema:
includes a boolean value 'manage_blockdevs' that adds the functionality of ncm-filesystems, because of duplicate and older code in ncm-filesystems. This deprecates ncm-filesystems .

This PR is build upon #562 !

### DIFF
--- a/ncm-fstab/src/main/perl/fstab.pm
+++ b/ncm-fstab/src/main/perl/fstab.pm
@@ -137,14 +137,14 @@ sub delete_outdated
     }
 
     foreach my $outdated (@rm) {
-    	$self->info("Removing line $outdated");
-    	$fstab->replace_lines (qr{$outdated}, qr{^$}, "");
-
         if ($remove) {
-            $self->info("Removing filesystem for line  $outdated");
             my $fsrm = NCM::Filesystem->new_from_fstab ($outdated);
+            $self->info("Removing filesystem for $fsrm->{mountpoint}");
             $fsrm->remove_if_needed==0 or return 0;
         }
+        $self->info("Removing line $outdated");
+        $fstab->replace_lines (qr{$outdated}, qr{^$}, "");
+
     }
     return 1;
 }
@@ -189,13 +189,13 @@ sub create_blockdevices
     }
     foreach (sort partition_compare @part) {
         if ($_->create != 0) {
-            throw_error ("Couldn't create partition: " . $_->devpath);
+            $self->error("Couldn't create partition: " . $_->devpath);
             return 0;
         }
     }
     foreach (@$fs) {
         if ($_->create_if_needed != 0) {
-            throw_error ("Failed to create filesystem:  $_->{mountpoint}");
+            $self->error("Failed to create filesystem:  $_->{mountpoint}");
             return 0;
         }
     }
@@ -208,12 +208,12 @@ sub Configure
     my ($self, $config) = @_;
 
     my $tree = $config->getTree($self->prefix());
-    my $create = $tree->{manage_blockdevs};
+    my $manage = $tree->{manage_blockdevs};
 
     my $fstab = CAF::FileEditor->new (NCM::Filesystem::FSTAB, log => $self,
 				      backup => '.old');
 
-    if ($create && $NoAction) {
+    if ($manage && $NoAction) {
         $self->warn ("--noaction not supported. Leaving."); # Should be checked if is still true
         return 1;
     }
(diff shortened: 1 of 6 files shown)


ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -75,26 +119,34 @@ sub mount_from_entry
...
 
     while (my $f = <$fstab>) {
     	my $mount = $self->mount_from_entry($f) or next;
-    	if (!exists ($mounts{$mount})) {
+    	if (!exists ($mounts{$mount}) && !exists($mounts{abs_path($mount)})) {
     	    CAF::Process->new ([UMOUNT, $mount],
     			       log => $self)->run();
     	    push (@rm, $f);
-    	    $self->verbose ("Scheduling for removal: $mount");
+            $self->verbose ("Scheduling for removal of fstab: $mount");
```
ned21: Are these to log messages not a bit redundant and confusing? Could they be combined into a single message? It would also be great to log when the umount was successful or not according to $?

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -123,25 +175,76 @@ sub remount_everything
     return $rt;
 }
 
+sub create_blockdevices
```
ned21: This sub is undocumented.

ncm-fstab/src/main/perl/fstab.pm
```diff
@@ -123,25 +175,76 @@ sub remount_everything
...
+{
+    my ($self, $config, $fs) = @_;
+    # Partitions must be created first, see bug #26137
```
ned21: Is that a legacy sf bug number?

ned21:
I would prefer that ncm-filesystems not be deprecated. The intent of the filesystems/fstab split was to allow ncm-filesystems to contain all the "dangerous" code that might destroy data (by e.g. overwriting a partition) and keep ncm-fstab as only managing /etc/fstab. "Do one thing and do it well". The addition of the "manage_blockdevs" attribute is a replacement for ```/software/components/filesystems/active = false```

jrha:
What is the motivation for bringing the functionality of ncm-filesystems into ncm-fstab?
In general we only deploy ncm-fstab because of the safety concerns @ned21 raised.

This should not be merged without a larger discussion of the risks/benefits.

kwaegema:
The motivation is that there is a lot of code in ncm-filesystems which does the same as ncm-fstab, so when certain changes are done to fstab, it should also be done for filesystems. And even then, it mostly does the same, but with extra. that is why it can easily be added with one boolean. Furthermore the code of filesystems is somewhat older, without tests

ned21:
Could you be more specific? The preference would be to remove the duplicate code from ncm-filesystems and then use a post-depedency to take care of /etc/fstab, but without a more detailed example/use case, it's not clear if what I am suggesting is possible.

kwaegema:
@ned21  the duplicate code is not that easily removed, because a lot of things has to be done for both functionalities:eg. protected hash has to be build, for having the 'valid mounts', those that should not be deleted. Then both fstab and filesystems components are iterating over the filesystems and removing old entries: fstab only from fstabfile, filesystems the filesystems. This is where I added the boolean, so it is just one iteration. I think it is more consistent code, but I can't argue that the component gains power

jrha:
I'm not comfortable with this as it is without a longer discussion, so I'm going to postpone this until we can discuss it at the next workshop.

ncm-fstab/src/main/perl/fstab.pm
jrha: removal **from** fstab

brunocanning:
I'm not against this if making changes to the _filesystems_ is a boolean and false by default.

However, our recent experience at RAL is that 'protected_mounts' listed in ncm-fstab configuration are not completely protected. Typically, all our filesystems are created at installation time and don't change until the host is reinstalled or repurposed. If we change the filesystem layout in quattor, e.g. add new partitions or change the filesystem on the partition, the component will make changes to the line for this mount point in /etc/fstab. We would like the entry for each protected mount point to be completely protected from changes.

stdweird:
@brunocanning can you open an issue for the additional protection feature you want?

kwaegema:
@brunocanning are you sure you did use the last ncm-fstab. in the last version, there are finer options:
Of the documentation in the fstab schema:
keep entries are always kept, but can be changed
static entries can not be changed, but can be deleted
protected_mounts is still here for backwards compability, and is the same as keep/mounts

so protected_mounts was only for not deleting entries once they were in there. What you need is static/mounts or static/fstypes, which will do the trick

kwaegema:
@brunocanning we have this in or fstab service file:
'keep/mounts' = FSTAB_PROTECTED_MOUNTS;
'static/mounts'= FSTAB_PROTECTED_MOUNTS;
'keep/fs_types' = FSTAB_PROTECTED_FSTYPES;
'static/fs_types' = FSTAB_PROTECTED_FSTYPES;

piojo:
`filesystems` used to embed `fstab` and I split it for safety concerns.  These components share a library (that's what libraries are for), but why do you want this removed?

piojo:
Agreement from the workshop: refactoring the duplicated code into `ncm-lib-blockdevices` or somewhere else is a better and safer way.  RAL and Morgan Stanley still want to keep the dangerous code isolated and if possible not even installed.

PR #601, opened 2015-10-08
ncm-interactivelimits: merge config-{rpm,common} into config.pan.

ajf8:


ncm-interactivelimits/src/main/pan/components/interactivelimits/config.pan
```diff
@@ -5,5 +5,18 @@
 
 unique template components/${project.artifactId}/config;
 
-include { 'components/${project.artifactId}/config-common' };
-include { 'components/${project.artifactId}/config-rpm' };
+include { 'components/${project.artifactId}/schema' };
```
stdweird: remove the `{}`, they are not required to include pure strings, only when handling dml

ncm-interactivelimits/src/main/pan/components/interactivelimits/config.pan
```diff
@@ -5,5 +5,18 @@
...
+include { 'pan/functions' };
+
+# Package to install.
+"/software/packages" = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
+
+# Set prefix to root of component configuration.
+prefix '/software/components/${project.artifactId}';
+
+#'version' = '${project.version}';
+#'package' = 'NCM::Component';
```
ned21: Same as comment on #602 - these lines should either be removed a part of clean up or uncommented because they are required.

stdweird:
LGTM, needs rebase

PR #603, opened 2015-10-08
ncm-resolver: merge config-{rpm,common} into config.pan

ajf8:


ncm-resolver/src/main/pan/components/resolver/config.pan
```diff
@@ -5,5 +5,12 @@
...
-include { 'components/${project.artifactId}/config-rpm' };
+include 'components/${project.artifactId}/schema';
+include 'pan/functions';
+
+"/software/packages" = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
+
+prefix '/software/components/${project.artifactId}';
+
+'active' ?= true;
+'dispatch' ?= true;
```
stdweird: pre dependency on spma?

ajf8: There wasn't one originally, I'll add it in.

stdweird: @ajf8 i was not suggesting you forgot it, but while you are doing some (much appreciated) cleanup, why not fix it.

stdweird:
LGTM, awaiting a rebase

PR #605, opened 2015-10-09
ncm-network: change type of nisdomain to something closer to reality

### LINKED ISSUE #604: ncm-network: nisdomain should not be type_fqdn
See discussion in quattor/release#115

Replaces quattor/template-library-core#92

jrha:
A valid NIS domain is (as far as anyone can tell) a string of up to sixty-four non-whitespace characters.

Fixes #604


stdweird:
LGTM. i'll merge when jenkins gives green light

PR #606, opened 2015-10-12
ncm-ssh: fix some issues introduced during rewrite.

ajf8:
* Add GSSAPIKeyExchange, GSSAPIStrictAcceptorCheck,
   KerberosTicketAuthentication back into schema.

 * /dev/stdin because /proc/self does not exist on Solaris and
   possibly other UNIX.

 * Add support for multiline options back in, determined by
   a list in the code.

 * Don't hardcode the sshd/sshd_config/ssh_config (/usr/sbin/sshd
   doesn't exist for us), make it possible to set through profile.

Backwards incompatible schema change - AcceptEnv, ListenAddress, HostKey changed to arrays.

### DIFF
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -6,6 +6,7 @@
 declaration template components/ssh/schema;
 
 include 'quattor/types/component';
+include 'pan/types';
 
 type ssh_yesnostring = string with match(SELF, "^(yes|no)$");
 
@@ -18,6 +19,7 @@ type ssh_core_options_type = {
     "Compression"                       ? string with match (SELF, '^(yes|delayed|no)$')
     "GSSAPIAuthentication"              ? ssh_yesnostring
     "GSSAPICleanupCredentials"          ? ssh_yesnostring
+    "GSSAPIKeyExchange"                 ? ssh_yesnostring
     "GatewayPorts"                      ? ssh_yesnostring
     "HostbasedAuthentication"           ? ssh_yesnostring
     "LogLevel"                          ? string with match (SELF, '^(QUIET|FATAL|ERROR|INFO|VERBOSE|DEBUG[123]?)$')
@@ -35,7 +37,8 @@ type ssh_core_options_type = {
 type ssh_daemon_options_type = {
     include ssh_core_options_type
     "AFSTokenPassing"                   ? ssh_yesnostring
-    "AcceptEnv"                         ? ssh_yesnostring
+    @{AcceptEnv, one per line}
+    "AcceptEnv"                         ? string[]
     "AllowAgentForwarding"              ? ssh_yesnostring
     "AllowGroups"                       ? string
     "AllowTcpForwarding"                ? ssh_yesnostring
@@ -48,7 +51,9 @@ type ssh_daemon_options_type = {
     "ClientAliveInterval"               ? long
     "DenyGroups"                        ? string
     "DenyUsers"                         ? string
-    "HostKey"                           ? string
+    "GSSAPIStrictAcceptorCheck"         ? ssh_yesnostring
+    @{HostKey, one per line}
+    "HostKey"                           ? string[]
     "HPNDisabled"                       ? ssh_yesnostring
     "HPNBufferSize"                     ? long
     "IgnoreRhosts"                      ? ssh_yesnostring
@@ -58,9 +63,11 @@ type ssh_daemon_options_type = {
     "KerberosGetAFSToken"               ? ssh_yesnostring
     "KerberosOrLocalPasswd"             ? ssh_yesnostring
     "KerberosTgtPassing"                ? ssh_yesnostring
+    "KerberosTicketAuthentication"      ? ssh_yesnostring
     "KerberosTicketCleanup"             ? ssh_yesnostring
     "KeyRegenerationInterval"           ? long
-    "ListenAddress"                     ? string
+    @{ListenAddress, one per line}
+    "ListenAddress"                     ? type_hostport[]
     "LoginGraceTime"                    ? long
     "MaxAuthTries"                      ? long
     "MaxStartups"                       ? long
@@ -109,11 +116,14 @@ type ssh_client_options_type = {
 type ssh_daemon_type = {
     "options" ? ssh_daemon_options_type
     "comment_options" ? ssh_daemon_options_type
+    "sshd_path" ? string
+    "config_path" ? string
 };
 
 type ssh_client_type = {
     "options" ? ssh_client_options_type
     "comment_options" ? ssh_client_options_type
+    "config_path" ? string
 };
 
 type component_ssh_type = {
--- a/ncm-ssh/src/main/perl/ssh.pm
+++ b/ncm-ssh/src/main/perl/ssh.pm
@@ -20,20 +20,25 @@ our $NoActionSupported = 1;
 
 use CAF::Process;
 use CAF::FileEditor;
-use constant SSHD_CONFIG  => "/etc/ssh/sshd_config";
-use constant SSH_CONFIG   => "/etc/ssh/ssh_config";
-use constant SSH_VALIDATE => qw(/usr/sbin/sshd -t -f /proc/self/fd/0);
+use constant DEFAULT_SSHD_CONFIG  => "/etc/ssh/sshd_config";
+use constant DEFAULT_SSHD_PATH    => "/usr/sbin/sshd";
+use constant DEFAULT_SSH_CONFIG   => "/etc/ssh/ssh_config";
+use constant SSHD_CONFIG_MULTILINE => qw(HostKey AcceptEnv ListenAddress);
 
 # Returns true if $file is a valid SSHD configuration file.
 sub valid_sshd_file
 {
-    my ($self, $file) = @_;
+    my ($self, $file, $cfg) = @_;
+
+    # Use /dev/stdin, instead of /proc/self/fd/0 (which is used in some other components).
+    # This is because /proc/self/fd/0 does not exist on Solaris. 
+    my $cmdline = [ $cfg->{sshd_path} || DEFAULT_SSHD_PATH, '-t', '-f', '/dev/stdin' ];
 
     my $cmd = CAF::Process->new(
-        [SSH_VALIDATE],
+        $cmdline,
         log         => $self,
-        stdin       => "$file",
         stderr      => \my $err,
+        stdin       => "$file",
         keeps_state => 1
     );
 
@@ -45,6 +50,15 @@ sub valid_sshd_file
     }
     $self->warn("Non-fatal issues in the configuration file: $err") if $err;
     return 1;
+}
+
+# escape values for the ssh config format.
+# adds \ before ?{}.()[]
+sub ssh_escape
+{
+    my ($self, $val) = @_;
+    $val =~ s{([?{}.()\[\]])}{\\$1}g;
+    return $val;
 }
 
 #
@@ -74,16 +88,23 @@ sub handle_config_file
             my $ref = ref($val);
             if($ref) {
                 if($ref eq 'ARRAY') {
+                    if (grep {$_ eq $option} SSHD_CONFIG_MULTILINE) {
+                        # remove any existing line for this option
+                        # then add the options back in at the end of the file
+                        $fh->remove_lines(qr{(?i)^\W*$option(?:\s+\S+)+}, qr{^#});
+                        foreach my $multival (@$val) {
+                            print $fh "$option " . $self->ssh_escape($multival) . "\n";
+                        }
+                        next;        
+                    }
                     $val = join(',', @$val);
                 } else {
                     $self->error("Unsupported value reference $ref for option $option.",
                                  " (Possibly bug in profile/schema).");
                     next;
                 }
-            };
-
-            my $escaped_val = $val;
-            $escaped_val =~ s{([?{}.()\[\]])}{\\$1}g;
+            }
+            my $escaped_val = $self->ssh_escape($val);
             $fh->add_or_replace_lines(
                 qr{(?i)^\W*$option(?:\s+\S+)+}, qr{^\s*$comment\s*$option\s+$escaped_val\s*$},
                 "$comment$option $val\n",
@@ -92,7 +113,7 @@ sub handle_config_file
         }
     }
 
-    if ($validate && !$validate->($self, $fh)) {
+    if ($validate && !$validate->($self, $fh, $cfg)) {
         $fh->cancel();
     }
 
@@ -110,7 +131,10 @@ sub Configure
     my $ok         = 1;
 
     if ($ssh_config->{daemon}) {
-        if ($self->handle_config_file(SSHD_CONFIG, 0600, $ssh_config->{daemon}, \&valid_sshd_file))
+        if ($self->handle_config_file($ssh_config->{daemon}->{config_path} || DEFAULT_SSHD_CONFIG,
+                                      0600,
+                                      $ssh_config->{daemon},
+                                      \&valid_sshd_file))
         {
             CAF::Process->new([qw(/sbin/service sshd condrestart)], log => $self)->run();
             if ($?) {
@@ -121,7 +145,9 @@ sub Configure
     }
 
     if ($ssh_config->{client}) {
-        $self->handle_config_file(SSH_CONFIG, 0644, $ssh_config->{client});
+        $self->handle_config_file($ssh_config->{client}->{config_path} || DEFAULT_SSH_CONFIG,
+                                  0644,
+                                  $ssh_config->{client});
     }
     return $ok;
 }
--- a/ncm-ssh/src/test/perl/valid.t
+++ b/ncm-ssh/src/test/perl/valid.t
@@ -7,7 +7,7 @@ use CAF::Object;
 use NCM::Component::ssh;
 use Readonly;
 
-Readonly my $CMD => join(" ", NCM::Component::ssh::SSH_VALIDATE);
+my $cmd = join(" ", (NCM::Component::ssh::DEFAULT_SSHD_PATH, '-t', '-f', '/dev/stdin'));
 
 my $cmp = NCM::Component::ssh->new("ssh");
 
@@ -19,11 +19,11 @@ Test for the C<valid_ssh_file> predicate.
 
 =cut
 
-set_command_status($CMD, 0);
+set_command_status($cmd, 0);
 ok($cmp->valid_sshd_file("foo"), "Success upon valid file");
 
-set_command_status($CMD, 1);
-set_desired_err($CMD, "Error");
+set_command_status($cmd, 1);
+set_desired_err($cmd, "Error");
 ok(!$cmp->valid_sshd_file("foo"), "Invalid file is detected");
 
 done_testing();


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -5,7 +5,7 @@
 
 declaration template components/ssh/schema;
 
-include 'quattor/types/component';
+include 'quattor/schema';
```
stdweird: what do you need that is not in the component template? we should try to keep this minimal.

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -48,7 +49,8 @@ type ssh_daemon_options_type = {
     "ClientAliveInterval"               ? long
     "DenyGroups"                        ? string
     "DenyUsers"                         ? string
-    "HostKey"                           ? string
+    "GSSAPIStrictAcceptorCheck"         ? ssh_yesnostring
+    "HostKey"                           ? string[]
```
stdweird: this is backwards incompatible. is it really needed?

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -20,20 +20,22 @@ our $NoActionSupported = 1;
...
+use constant DEFAULT_SSHD_PATH    => "/usr/sbin/sshd";
+use constant DEFAULT_SSH_CONFIG   => "/etc/ssh/ssh_config";
 
 # Returns true if $file is a valid SSHD configuration file.
 sub valid_sshd_file
 {
-    my ($self, $file) = @_;
+    my ($self, $file, $cfg) = @_;
+
+    my @cmdline = ( $cfg->{sshd_path} || DEFAULT_SSHD_PATH, '-t', '-f', '/dev/stdin' );
```
stdweird: `/dev/stdin`, how does that work? (e.g. who can use it)

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -47,6 +49,13 @@ sub valid_sshd_file
     return 1;
 }
 
+sub ssh_escape
```
stdweird: a (small) comment what this is for

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -74,16 +83,21 @@ sub handle_config_file
             my $ref = ref($val);
             if($ref) {
                 if($ref eq 'ARRAY') {
+                    if ($option_set eq 'multiline_options') {
+                        $fh->remove_lines(qr{(?i)^\W*$option(?:\s+\S+)+}, qr{^#});
```
stdweird: this will always remove content that you add below. why does it it need to be done this way? add a small comment why this is done

stdweird:
@ajf8 you made a remark about `Temp::File` that i do not understand (nor see anything in the diff).
everything looks ok, but don't be afraid to add comments in the code :smile:

ncm-ssh/src/main/perl/ssh.pm
ajf8: I used a File::Temp initially, then someone here suggested using /dev/stdin so I updated the commit.

On Red Hat Linux, this is a symlink to /proc/self/fd/0, so it works the same as before. On Solaris there is a symlink to fd/0.

$ ls -l /dev/stdin
lrwxrwxrwx   1 root            6 May 23 05:23 /dev/stdin -> ./fd/0
$ ls -l /dev/fd/0
crw-rw-rw-   1 root     404,   0 Oct 12 16:49 /dev/fd/0
$ echo test | cat /dev/stdin
test

stdweird: very nice. can you add this as a comment?

ncm-ssh/src/main/pan/components/ssh/schema.pan
gombasg: Well, the biggest problem with the current schema is the handling of multi-valued options. The sshd config file has 3 different syntaxes for specifying multiple values:

* space-separated, e.g. `AcceptEnv`, `AllowUsers`, `AuthorizedKeysFile`, `PermitOpen`
* comma-separated, e.g. `AuthenticationMethods`, `Ciphers`
* repeating the same key multiple times, e.g. `HostKey`, `AcceptEnv`, `ListenAddress`.

The current Perl code assumes comma-separated representation if the value is an array, and there's no direct support for the other two syntaxes. Arguably, it would be better to use arrays for multi-line options, and use simple strings (plus string operations on the Pan template side) to handle comma-separated and space-separated values. However, that would break the existing array-typed options - I think there's only one, so it might be possible to hard-code the difference for that specific option in the Perl code.

Latest sshd versions also allow using the "Match ..." syntax on the server-side, to allow overriding the defaults set outside of such `Match` blocks based on various criteria; we don't have any immediate needs to support that, but it would be nice if the schema would eventually allow that as well.

stdweird: we use the following for per user ssh (client) config:
https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/metaconfig/ssh/client_attrs.tt

the boolean handling in the TT can be cleaned up (there's a https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/metaconfig/ssh/pan/schema.pan which uses boolean instead of `yesno_string`), but the remaining is indeed hardcoded.

i have nothing against the hardcoding in the perl code, it's unavoidable due to crappy config formats. i was just wondering if the backwards incompatibility in the schema was really needed).
i'm in favour of handling multi valued options with `string[]` rather than some pan string joining.

ajf8: We could have a separate schema for the multiline_options, instead of using the same one, I think this makes the most sense actually. This way we can leave HostKey as a string in options to keep compatibility, and put a string[] in multiline_options.

The Match blocks shouldn't be too difficult to support, hopefully we can just embed the existing structures inside another one.

ncm-ssh/src/main/pan/components/ssh/schema.pan
ajf8: There's nothing extra needed, but we don't have quattor/types/component, presumably because our template tree is based on an old template-library-core and it changed at some point.

I'm not against changing things on our end, but the reason I changed it is because include quattor/schema seems to be the much more common preamble across configuration-modules-core.

$ grep -r 'include.*quattor/schema' */src/main/pan | wc -l
63
$ grep -r 'include.*quattor/types/component' */src/main/pan | wc -l
9

stdweird: @ajf8 yes, it got refactored completly, because we want to unittest templates at some point, and minimal includes are a lot better.
in the same effort, the bind statement at the bottom should be moved to the config file (but you aren't touching that, so it's ok for now).

ncm-ssh/src/main/pan/components/ssh/schema.pan
stdweird: for multiple `Hostkey` you could use a 
```pan
type sshd_hostkeys = {
    'rsa' ? string
    'dsa' ? string
...
};
```
and use a new attribute for the multiline
```pan
Hostkeys ? sshd_hostkeys
```
 so as key the protocol and value the filename. the key does not appear in the config file, but that's what makes it multiple. it also makes no sesne to have more keys than supported protocols, or more than one file for the same protocol

for match, check the https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/metaconfig/ssh/pan/schema.pan

ajf8: I've changed multiline options to have its own type, and options/HostKey back to string. So hopefully this is completely backwards compatible.

I could do an sshd_hostkeys type, but I guess it would require something specifically for handling HostKey. Would it go under component/ssh/mutliline_options or component/ssh? We could do something slightly more generic when it encounters a hash, but I'm not sure discarding the keys (rsa, dsa etc) is particularly generic. I think I would prefer keeping things simple and consistent (like they are now), but I'm happy to do it this way too.

ncm-ssh/src/main/perl/ssh.pm
ajf8: What do you mean by remove content that you add below?

The intention (and it appeared to work) was to remove all existing instances of an option in the file, and then add them back in. This matches the behaviour of regular options, if you set an option then you expect an existing one (or multiple in this case) of that name to be replaced.

I looked at CAF::FileWriter and couldn't figure anything that would fit multiline options, like add_or_replace does for the regular ones, other than doing this.

ajf8: The other nice thing about this is that it provides a nice upgrade path for us to update the templates before the code is upgraded.

We can leave stuff in options (temporarily) to keep compatability with the old code, then if something goes into multiline_options it removes what was set in options.

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -20,20 +20,24 @@ our $NoActionSupported = 1;
...
 
 # Returns true if $file is a valid SSHD configuration file.
 sub valid_sshd_file
 {
-    my ($self, $file) = @_;
+    my ($self, $file, $cfg) = @_;
+
+    # Use /dev/stdin, instead of /proc/self/fd/0 (which is used in some other components).
+    # This is because /proc/self/fd/0 does not exist on Solaris. 
+    my @cmdline = ( $cfg->{sshd_path} || DEFAULT_SSHD_PATH, '-t', '-f', '/dev/stdin' );
```
stdweird: use arayref directly `$cmdline = [...]` and then `CAF::Process->new($cmdline, ...)`

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -108,12 +115,16 @@ type ssh_client_options_type = {
 
 type ssh_daemon_type = {
     "options" ? ssh_daemon_options_type
+    "multiline_options" ? ssh_daemon_multiline_options_type
```
stdweird: i'm not a fan of refelcting the format in this way, `HostKey` should be configured under options like the other options. we typically use a static list of keys that should be converted in certain format. 
(and i liked my previous proposla of making `HostKey` a dict with valid key types :smile: )

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -20,20 +20,24 @@ our $NoActionSupported = 1;
...
+use constant DEFAULT_SSHD_PATH    => "/usr/sbin/sshd";
+use constant DEFAULT_SSH_CONFIG   => "/etc/ssh/ssh_config";
 
 # Returns true if $file is a valid SSHD configuration file.
 sub valid_sshd_file
 {
-    my ($self, $file) = @_;
+    my ($self, $file, $cfg) = @_;
+
+    # Use /dev/stdin, instead of /proc/self/fd/0 (which is used in some other components).
```
stdweird: pleaser open an issue in this repository that we scan other code for any `/proc/self/...` and replace it with more OS neutral code, like this one.

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -47,6 +51,15 @@ sub valid_sshd_file
     return 1;
 }
 
+# escape values for the ssh config format.
+# adds \ before ?{}.()[]
+sub ssh_escape
```
stdweird: this is potentially backwards incompatible. current profiles might have the escaped values as the actual value, this will make it worse. your pattern should not match any already escaped characters. adding negative lookbehind like 
```
$val =~ s{(?<!\\)([?{}.()\[\]])}{\\$1}g;
```
should avoid this, but this needs to be documented and is still potentially backwards incompatible.

stdweird: nvm, this used to be the case already, i see you just factored out the code below.

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -74,16 +87,22 @@ sub handle_config_file
             my $ref = ref($val);
             if($ref) {
                 if($ref eq 'ARRAY') {
+                    if ($option_set eq 'multiline_options') {
```
stdweird: if you use a `multiline_options` whitelist, you can do something like
```perl
use constant SSHD_CONFIG_MULTILINE => qw(HostKeys);
...
if($ref eq 'ARRAY') {
    if (grep {$_ eq $option} SSHD_CONFIG_MULTILINE) {
        ... do something
    } else {
        $val = join(',', @$val);
    }
}
```

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -74,16 +87,22 @@ sub handle_config_file
...
+                        # remove any existing line for this option, then add the options back in
+                        $fh->remove_lines(qr{(?i)^\W*$option(?:\s+\S+)+}, qr{^#});
```
stdweird: this will always readd  all the keys (well, all the mutline opts) at the end of the file. not sure this is so ideal...

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -64,7 +77,7 @@ sub handle_config_file
         backup => '.old'
     );
 
-    foreach my $option_set (qw(comment_options options)) {
+    foreach my $option_set (qw(comment_options options multiline_options)) {
```
stdweird: what if we build a list of key/value first (adding the HostKey as key/value entries with same key and different), and then using that list to `add_or_replace_lines`?
```perl
my @todo;
foreach my $option_set (qw(comment_options options)) {
    next if (! $cfg->{$option_set});
    my $comment = $option_set eq 'comment_options' ? '#' : '';
    foreach my $option (sort keys %{$cfg->{$option_set}}) {
             my $val = $cfg->{$option_set}->{$option};
             if (grep {$_ eq $option} SSHD_CONFIG_MULTILINE) {
                 foreach my $ml_value (@$val) {
                     push(@todo, [$option, $ml_value, $comment]);
                 }
             } else {
                 push(@todo, [$option, $val, $comment]);
             } 
    }
}
foreach my $data (@todo) {
     my ($option, $val, $comment) = @$data;
     ... 
}
```

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -110,7 +129,9 @@ sub Configure
     my $ok         = 1;
 
     if ($ssh_config->{daemon}) {
-        if ($self->handle_config_file(SSHD_CONFIG, 0600, $ssh_config->{daemon}, \&valid_sshd_file))
+        if ($self->handle_config_file($ssh_config->{daemon}->{config_path}
+                                          || DEFAULT_SSHD_CONFIG,
+                                      0600, $ssh_config->{daemon}, \&valid_sshd_file))
```
stdweird: either all on one line, or one per line

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -110,7 +129,9 @@ sub Configure
...
         {
             CAF::Process->new([qw(/sbin/service sshd condrestart)], log => $self)->run();
```
stdweird: this is not EL7 safe, systemd only has `try-restart`, better to switch to `CAF::Service` here (does this even work on solaris?)

stdweird:
@ajf8 i assume you are not using the `Temp::File` anymore? maybe edit the description a bit to reflect this.

ncm-ssh/src/main/perl/ssh.pm
ajf8: My recollection is that add_or_replace_lines looked great for single line options, but I couldn't see a way for it to easily work for multiline options. Maybe I'm missing something though.

You provide add_or_replace_lines with a regex of lines to match, and what to replace it with. So if you provide the option name as the match, this is going to match every instance of that option, and replace each line with the same value right? That's why I went for the remove and readd approach - I know the multiline options will always end up at the end of the file, but as we're doing it in sorted key order, the output should be consistent at least so I didn't see a problem.

stdweird: yes, it will work as you describe, and as you say, it will be consistent. i guess that e.g. `HostKey` are never set by the distributed configuration file, so at least in the case of `HostKey`, it will not trigger an unnecessary restart.

ncm-ssh/src/main/perl/ssh.pm
ajf8: Yes, this will not work on Solaris currently. We have a few components for Solaris which have been modified, this being one of them, but they are based on older version. I have an outstanding task to bring these up-to-date and refactor+submit our Solaris patches :)

ajf8:
Good point, I've updated the commit message to mention /dev/stdin.

ncm-ssh/src/main/perl/ssh.pm
stdweird: actually, `CAF::Service` has no `condrestart`, maybe we should add one so we can do `CAF::Service->new(['sshd'], log => $self)->condrestart()` here and worry about solaris/sysvinit/systemd in `CAF`. @ned21 do you remember why we didn't add `condrestart` to `CAF::Service`?

and it looks like `condrestart` is an undocumented alias for `try-restart` in systemd. so this might work on EL7 afterall...

ncm-ssh/src/main/perl/ssh.pm
ajf8: I'll change this to use a whitelist instead tomorrow.

ncm-ssh/src/main/perl/ssh.pm
ned21: No recollection, sorry. Sounds like it would be useful to provide cross-platform compatibility.

ajf8:
Hopefully all the points have been addressed, and I've just done a rebase.

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -20,20 +20,25 @@ our $NoActionSupported = 1;
 
 use CAF::Process;
 use CAF::FileEditor;
-use constant SSHD_CONFIG  => "/etc/ssh/sshd_config";
-use constant SSH_CONFIG   => "/etc/ssh/ssh_config";
-use constant SSH_VALIDATE => qw(/usr/sbin/sshd -t -f /proc/self/fd/0);
+use constant DEFAULT_SSHD_CONFIG  => "/etc/ssh/sshd_config";
+use constant DEFAULT_SSHD_PATH    => "/usr/sbin/sshd";
+use constant DEFAULT_SSH_CONFIG   => "/etc/ssh/ssh_config";
+use constant SSHD_CONFIG_MULTILINE => qw(HostKey AcceptEnv ListenAddress);
```
stdweird: @ajf8 this still requires support in the schema. these have to be in the `ssh_daemon_options_type` and be lists of things.

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -58,6 +60,7 @@ type ssh_daemon_options_type = {
     "KerberosGetAFSToken"               ? ssh_yesnostring
     "KerberosOrLocalPasswd"             ? ssh_yesnostring
     "KerberosTgtPassing"                ? ssh_yesnostring
+    "KerberosTicketAuthentication"      ? ssh_yesnostring
     "KerberosTicketCleanup"             ? ssh_yesnostring
     "KeyRegenerationInterval"           ? long
     "ListenAddress"                     ? string
```
stdweird: as an example, this has to beocme 
```pan
@{ListenAddress, one per line}
ListenAddress ? type_hostport[]
```
the docstring `@{}` is optional, but since this change in partiticular is backwards incompatible,  and you need to add a 
```pan
include 'pan/types';
```
to make sure the `type_hostport` is usable

ncm-ssh/src/main/perl/ssh.pm
stdweird: `AccetpEnv` is currenlty a boolean, which makes no sense at all. nice catch making this work.

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -74,16 +88,22 @@ sub handle_config_file
             my $ref = ref($val);
             if($ref) {
                 if($ref eq 'ARRAY') {
+                    if (grep {$_ eq $option} SSHD_CONFIG_MULTILINE) {
+                        # remove any existing line for this option, then add the options back in
```
stdweird: `back in` -> `back in at the end of the file`

stdweird:
@ajf8 you rproposal with new `AcceptEnv` and `ListenAddress` makes this PR backwards incompatible (for good reasons). can you update the description mentioning this and set the `backwards incompatible` label?

ajf8:
Hmm I don't think it is backwards incompatible.

If people are using single value options (ie. not a list), which they must be because we didn't support it earlier for these options (a list would have created a CSV which isn't valid for them), then we don't enter the "if ($ref)" code block which checks if an option is multiline and it should work as before.

ajf8:
Sorry, I was reading comments backwards :) And was only thinking about the code, now I see you point about the schema needing changing. I'll do that now.

ncm-ssh/src/main/perl/ssh.pm
stdweird: i created #643

stdweird:
LGTM 
@ajf8 only thing missing is adding the backwards incompatible attributs to the description

ajf8:
Which description, the commit message or is there some tag you add into the schema? Thanks

stdweird:
the description of the PR (i.e. the first comment)

ajf8:
Ah ok, done!

stdweird:
LGTM
@ajf8 thx!

PR #611, opened 2015-10-15
ncm-hostsfile: merge config-{common,rpm} into config.pan

ajf8:
Another one of these, the last one for now hopefully :)


ncm-hostsfile/src/main/pan/components/hostsfile/config.pan
```diff
@@ -5,5 +5,13 @@
...
+include 'components/${project.artifactId}/schema';
+
+# Package to install
+"/software/packages" = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
+
+# Set prefix to root of component configuration.
+prefix '/software/components/${project.artifactId}';
+
+'active' ?= false;
+'dispatch' ?= false;
```
ned21: Previously this was true (and for active). version has been dropped too: this is what causes ncm-cdispd to re-run the component when the rpm is updated.

jrha:
LGTM

stdweird:
also LGTM. 
@ajf8 thanks for all the effort!

PR #616, opened 2015-10-22
ncm-dirperm: checkmount feature.

ajf8:
This checks that a directory is within a mount, before changing
anything. Useful to ensure directories don't unintentionally end up
on the root filesystem.


ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +35,24 @@ sub Configure($$@) {
     # Define paths for convenience.
     my $base = "/software/components/dirperm";
 
+    my $filesystems = "/system/filesystems";
```
stdweird: @kwaegema what code is available in `ncm-lib-blockdevices` that does
* check mounted filesystems status (which ones are mounted and should be mounted)
* check if path is below supposed-t-be-mounted filesystem

kwaegema: fs->mounted returns if the fs is mounted. 
There is nothing in ncm-lib-blockdevices for what should be mounted. this is done in ncm-fstab.
Checking if path is below it is not there afaik

ncm-dirperm/src/main/pan/components/dirperm/config.pan
```diff
@@ -5,5 +5,15 @@
 
 unique template components/${project.artifactId}/config;
 
-include { 'components/${project.artifactId}/config-common' };
-include { 'components/${project.artifactId}/config-rpm' };
+include { "components/dirperm/schema" };
```
stdweird: no `{}`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +35,24 @@ sub Configure($$@) {
...
+    if ($config->elementExists($filesystems)) {
```
stdweird: hmmm, ugly `Element` code and `$_` usage...
```perl
my $fs_tree = $config->getTree($filesystems);
push(@configured_mounts, map {$_->{mountpoint}} 
                                               sort { $a->{mountpoint} cmp $b->{mountpoint}}
                                               @{$fs_tree});
```
requires some recent CCM (15.8 i think) and is completely untested :wink:

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +35,24 @@ sub Configure($$@) {
...
+        foreach (sort { $a->{mountpoint} cmp $b->{mountpoint} } @{$config->getElement($filesystems)->getTree()}) {
+            push @configured_mounts, $_->{mountpoint};
+        }
+    }
+
+    # Get actual mountpoints available on the server into a hash.
+    if (open(my $mtab, "<", MTAB_PATH)) {
```
stdweird: use `CAF::FileReader`
```perl
my $fh = CAF::FileReader->new(MTAB_PATH, log => $self);
foreach my line (split ("\n", "$fh")) {
...
}
```

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -189,6 +211,40 @@ sub process_path {
...
 			}
 		} elsif ($type eq "d") {
+            # Should we check if the needed mount is actually mounted?
+            # (i.e. don't build path on / (root) if the mount point is not available)
+            if ($pathentry->{checkmount}) {
+                # Which mountpoint is path built on?
+                my $path_mountpoint = undef;
+
+                foreach my $mntpt (@configured_mounts) {
+                    if ($path =~ /^$mntpt\//) {
```
stdweird: how certain are we that the entries in `configured_mounts` have no trailing slash?

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -189,6 +211,40 @@ sub process_path {
...
+                        $path_mountpoint = $mntpt;
+                        last;
+                    }
+                }
+
+                if (!defined($path_mountpoint)) {
+                    $self->error("couldn't find matching mount in profile for $path. skipping.");
+                    return 0
```
stdweird: add `;`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -189,6 +211,40 @@ sub process_path {
...
+
+                foreach my $mntpt (@configured_mounts) {
+                    if ($path =~ /^$mntpt\//) {
+                        $path_mountpoint = $mntpt;
+                        last;
+                    }
+                }
+
+                if (!defined($path_mountpoint)) {
+                    $self->error("couldn't find matching mount in profile for $path. skipping.");
```
stdweird: use capitals in messages?

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -189,6 +211,40 @@ sub process_path {
 				return 0;
 			}
 		} elsif ($type eq "d") {
+            # Should we check if the needed mount is actually mounted?
+            # (i.e. don't build path on / (root) if the mount point is not available)
+            if ($pathentry->{checkmount}) {
+                # Which mountpoint is path built on?
+                my $path_mountpoint = undef;
```
stdweird: because you don not actually care about the vale of the moutnpotin, you can do
```
my @path_mountpoints = grep {$path =~ m/^$_\//} @configured_mounts;
if (@path_mountpoints) {
  # found mountpoint $path_mountpoints[0] is the one you are looking for 
} else {
  error()...
}
```

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -189,6 +211,40 @@ sub process_path {
...
+                if (!defined($path_mountpoint)) {
+                    $self->error("couldn't find matching mount in profile for $path. skipping.");
+                    return 0
+                } elsif (not exists($available_mounts{$path_mountpoint})) {
+                    $self->error("filesytem mount $path_mountpoint not available for $path. skipping.");
+                    return 0;
+                }
+            }
+            if ($pathentry->{notroot}) {
+                my $path_mountpoint = undef;
```
stdweird: same remarks as above

stdweird:
@ajf8 ping?

jrha:
@ajf8 ping?

stdweird:
@ajf8 ping?

ajf8:
Sorry, not sure how I missed these pings, I will look at getting this PR finished.

ncm-dirperm/src/main/perl/dirperm.pm
ajf8: Unfortunately we do not have a recent CCM for $config->getTree(), but I did the other things.

ajf8:
Thanks a lot for the comments @stdweird, all addressed hopefully, and sorry for the delay picking them up.

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
```diff
@@ -30,11 +30,13 @@ function dirperm_permissions_valid = {
...
 type structure_dirperm_entry = {
-    'path'    : string with match(SELF, '^/')
-    'perm'    : string with match(SELF,'[0-7]{3,4}')
-    'owner'   : string with match(SELF, '\w+(:\w+)?')
-    'type'    : string with match(SELF, 'd|f')
-    'initdir' ? string[]
+    'path'       : string with match(SELF, '^/')
+    'perm'       : string with match(SELF,'[0-7]{3,4}')
+    'owner'      : string with match(SELF, '\w+(:\w+)?')
+    'type'       : string with match(SELF, 'd|f')
```
stdweird: `^[df]$`

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
```diff
@@ -30,11 +30,13 @@ function dirperm_permissions_valid = {
...
+    'initdir'    ? string[]
+    'checkmount' : boolean = false
+    'notroot'    : boolean = false
```
stdweird: i am not a big fan of negated option names? can we invert the meaning of the name (and logic of the code)?
so here
```pan
'root' : boolean = true
```

ncm-dirperm/src/main/perl/dirperm.pm
stdweird: ok.

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +36,22 @@ sub Configure($$@) {
...
+    my $filesystems = "/system/filesystems";
+    if ($config->elementExists($filesystems)) {
+        my $fs_tree = $config->getElement($filesystems)->getTree();
+        push(@configured_mounts, map {$_->{mountpoint}}
+                                 sort { $a->{mountpoint} cmp $b->{mountpoint} }
+                                 @{$fs_tree});
+    }
+
+    # Get actual mountpoints available on the server into a hash.
+    my $fh = CAF::FileEditor->new(MTAB_PATH, log => $self);
```
stdweird: if you can't use `FileReader, can you do a `$fh->cancel();` right after the new (and add a TODO to replace it ast some point in future). this is just in case, i also have no clue if you can edit mtab or what would happen.

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -52,14 +73,14 @@ sub process_path {
     # Pull out the values and check that they really are defined.
     my $path = $pathentry->{path};
     if (!defined($path)) {
-        $self->error("entry with undefined path");
+        $self->error("Entry with undefined path");
```
stdweird: add the undefined `$path` to the message? (it looks very useless without it imho).

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -52,14 +73,14 @@ sub process_path {
...
         return 0;
     }
 
     # Get the owner.
     my $owner = $pathentry->{owner};
     if (!defined($owner)) {
-        $self->error("entry with undefined owner");
+        $self->error("Entry with undefined owner");
```
stdweird: add `$owner` to the message

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -82,25 +103,25 @@ sub process_path {
         $gid = (getgrnam($group))[2] if ($group);
 
         if (!defined($uid) or !defined($gid)) {
-            $self->error("bad owner or group ($owner)");
+            $self->error("Bad owner or group ($owner)");
```
stdweird: add group to message like
```perl
$self->error("Bad owner or group ($owner/".($group ? $group : 'default user group').")");
```

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -82,25 +103,25 @@ sub process_path {
...
-        $self->error("owner field with bad format ($owner)");
+        $self->error("Owner field with bad format ($owner)");
         return 0;
     }
 
     # Permissions.
     my $perm = oct($pathentry->{perm});
     if (!defined($perm)) {
-        $self->error("entry with undefined permissions");
+        $self->error("Entry with undefined permissions");
```
stdweird:
```perl
$self->error("Entry with undefined permissions ".(defined($pathentry->{perm}) ? $pathentry->{perm} : '<undef>'));
```

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -82,25 +103,25 @@ sub process_path {
...
         return 0;
     }
 
     # Type: f=file, d=directory.
     my $type = $pathentry->{type};
     if (!defined($type)) {
-        $self->error("entry with undefined type");
+        $self->error("Entry with undefined type");
```
stdweird: same as with `perm`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -185,14 +206,35 @@ sub process_path {
...
+
+                if (!@path_mountpoints) {
+                    $self->error("Couldn't find matching mount in profile for $path. Skipping.");
+                    return 0;
+                } elsif (not exists($available_mounts{$path_mountpoints[0]})) {
+                    $self->error("Filesytem mount $path_mountpoints[0] not available for $path. Skipping.");
+                    return 0;
+                }
+            }
+            if ($pathentry->{notroot}) {
```
stdweird: with inverted logic `if(! $pathentry->{root}) {`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -185,14 +206,35 @@ sub process_path {
...
+            # Should we check if the needed mount is actually mounted?
+            # (i.e. don't build path on / (root) if the mount point is not available)
+            if ($pathentry->{checkmount}) {
+                # Which mountpoint is path built on?
+                my @path_mountpoints = grep {$path =~ m/^$_\//} @configured_mounts;
+
+                if (!@path_mountpoints) {
+                    $self->error("Couldn't find matching mount in profile for $path. Skipping.");
+                    return 0;
+                } elsif (not exists($available_mounts{$path_mountpoints[0]})) {
```
stdweird: why only check the first element? shouldn't it be the "longest" path? suppoes there is 3 partitions mounted under
`/`, `/var` and `/var/tmp`, then if you are using dirperm on `/var/tmp/x`, the `/var/tmp mountpoint` is the one to check, no? (or all of them?)

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -185,14 +206,35 @@ sub process_path {
...
+                    $self->error("Filesytem mount $path_mountpoints[0] not available for $path. Skipping.");
+                    return 0;
+                }
+            }
+            if ($pathentry->{notroot}) {
+                my @path_mountpoints = grep {$path =~ m/^$_\//} keys %available_mounts;
+                if (!@path_mountpoints) {
+                    $self->error("$path would be created on the root filesystem (notroot enabled).");
```
stdweird: `notroot enabled` -> `root disabled`

ncm-dirperm/src/main/perl/dirperm.pm
stdweird: ah no, the grep tales care of this. can you add a small comment why `0` is sufficient?

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +36,22 @@ sub Configure($$@) {
     # Define paths for convenience.
     my $base = "/software/components/dirperm";
 
+    my $filesystems = "/system/filesystems";
+    if ($config->elementExists($filesystems)) {
+        my $fs_tree = $config->getElement($filesystems)->getTree();
+        push(@configured_mounts, map {$_->{mountpoint}}
```
stdweird: the code belows relies on the fact that each path ends with a `/`. can you check and/or force this?

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -185,14 +206,35 @@ sub process_path {
...
+                if (!@path_mountpoints) {
+                    $self->error("$path would be created on the root filesystem (notroot enabled).");
+                    return 0;
+                }
+            }
 			# Make the directory and any parent directories.
 			eval { mkpath($path,0,$perm) };
 			if ($@) {
-				$self->error("error making directory: $@");
+				$self->error("Error making directory: $@");
```
stdweird: add `$path` and `$perm` to the message

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -201,7 +243,7 @@ sub process_path {
 	if(not $correct_ownership) {
 		my $cnt = chown $uid, $gid, $path;
 		if ($cnt != 1) {
-			$self->error("can't change owner on $path");
+			$self->error("Can't change owner on $path");
```
stdweird: add `uid/gid` to message

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -209,7 +251,7 @@ sub process_path {
 	if(not $correct_permissions) {
 		my $cnt = chmod $perm, $path;
 		if ($cnt != 1) {
-			$self->error("can't change permissions on $path");
+			$self->error("Can't change permissions on $path");
```
stdweird: add perm to message

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -240,11 +282,11 @@ sub process_path {
                     my $dstfile = "$path/" . basename($_);
                     if (! -f $dstfile) {
                         copy($_,$dstfile);
-                        $self->error("error copying file $_") if ($?);
+                        $self->error("Error copying file $_") if ($?);
```
stdweird: add `to $dstfile` to message, and maybe also `$?` and `$!`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -240,11 +282,11 @@ sub process_path {
...
 
                         my $cnt = chown $uid, $gid, $dstfile;
                         if ($cnt != 1) {
-                            $self->error("can't change owner on $dstfile");
+                            $self->error("Can't change owner on $dstfile");
```
stdweird: add uid/gid

stdweird:
even more remarks to code you didn't write :smile: (but thanks for the cleanup!)

ncm-dirperm/src/main/perl/dirperm.pm
ajf8: It's undefined so there is nothing to add to the message ;)

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
ajf8: I'm not sure we can, because there is no default that won't be breaking for others. It would mean "check this directory is on the root filesystem". If we set the default to true, it would break things for people that were using dirperm to create directories within mounts, and if we set it to false then it's going to break for everything on the root fs.

ncm-dirperm/src/main/perl/dirperm.pm
stdweird: ok, indeed this makes no sense :smile:

ncm-dirperm/src/main/perl/dirperm.pm
stdweird: hah, indeed. you could add the path to the message, but than that could be said about all messages. so feel free to do it or not, i'll stopping makeing remarks about code you didn't write

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
stdweird: i don't understand. this is a new attribute, so it won't affect anyone how it is named or defined. `notroot=false` and `root=true` are the same thing. (and i meant to change it here and invert the logic in the code as well ofcourse).

ncm-dirperm/src/main/perl/dirperm.pm
jrha: You always say that :stuck_out_tongue:

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
ajf8: We discussed this here the other day. 

I agree notroot=false and root=true are the same thing, but I disagree that as a new attribute it won't affect anyone. a root=true default will break the creation of directories in mounts, where people are already doing this.

I've changed it to "within_mount", I think this was suggested by @ned21, to avoid the double negation. 

"notroot" or "within_mount" only lets us check for one possible configuration (instead of two - non-root and root), so avoids the problem. Hope this explains it better :)

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
```diff
@@ -30,11 +30,13 @@ function dirperm_permissions_valid = {
...
-    'owner'   : string with match(SELF, '\w+(:\w+)?')
-    'type'    : string with match(SELF, 'd|f')
-    'initdir' ? string[]
+    'path'         : string with match(SELF, '^/')
+    'perm'         : string with match(SELF,'[0-7]{3,4}')
+    'owner'        : string with match(SELF, '\w+(:\w+)?')
+    'type'         : string with match(SELF, '^[df]$')
+    'initdir'      ? string[]
+    'checkmount'   : boolean = false
+    'within_mount' : boolean = false
```
stdweird: much better!

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -21,6 +22,10 @@ use Fcntl ':mode';
 
 local(*DTA);
```
stdweird: you can remove this, we also don't know what it means :smile:

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -21,6 +22,10 @@ use Fcntl ':mode';
...
 
+use constant MTAB_PATH => "/etc/mtab";
+
+my @configured_mounts;
+my %available_mounts;
 
 ##########################################################################
 sub Configure($$@) {
```
stdweird: remove the `($$@)`

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -185,14 +207,35 @@ sub process_path {
 			open TMP, ">>$path";
 			close TMP;
 			if (! -f $path) {
-				$self->error("error creating file $path");
+				$self->error("Error creating file $path");
 				return 0;
 			}
 		} elsif ($type eq "d") {
+            # Should we check if the needed mount is actually mounted?
```
stdweird: the is something wrong with the indentation of this block

ajf8: It was quite a mix of tabs and spaces, so I've replaced all the tabs with 4 spaces.

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -82,33 +102,33 @@ sub process_path {
...
+        $self->error("Entry with undefined type: $path");
         return 0;
     }
 
     $self->debug(1,"Configuring $path : Type=$type, Owner=$user, Group=" . ($group ? $group : "undefined") . ", Permissions=oct($perm)");
 
     if ((-l $path) && (! -e $path)) {
-	$self->error("$path is a broken symlink");
-	return 0;
+    $self->error("$path is a broken symlink");
```
stdweird: wrong indentation

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -31,6 +34,23 @@ sub Configure($$@) {
...
+    }
+
+    # Get actual mountpoints available on the server into a hash.
+    # TODO: replace with FileReader
+    my $fh = CAF::FileEditor->new(MTAB_PATH, log => $self);
+    for my $line (split("\n", "$fh")) {
+        my ($device, $mnt_point, $fstype, $options) = split /\s+/, $line;
+        $available_mounts{$mnt_point}++;
+    }
+    $fh->cancel();
```
stdweird: `cancel` immediately after new like
```perl
my $fh = CAF::FileEditor->new(MTAB_PATH, log => $self);
$fh->cancel();
```

ncm-dirperm/src/main/perl/dirperm.pm
```diff
@@ -155,64 +175,85 @@ sub process_path {
...
-		my $cnt = chmod $perm, $path;
-		if ($cnt != 1) {
-			$self->error("can't change permissions on $path");
-			return 0;
-		}
-	}
+    if(not $exists) {
+        if ($type eq "f") {
+            # Make the file.
+            open TMP, ">>$path";
```
stdweird: what are you trying to achieve here? it creates a file when none exists, but of one exists, you don't actully do anything (except opening it). If that is ok, you can use FileEditor for this.
```perl
CAF::FileEditor->new($path, log=>$self)->close();                                                                                                        
if($EC->error()) {                                                                                                                                    
    $self->error("Error creating file $path:".$EC->error()->reason());
    $EC->ignore_error();
    return 0;
}
```

ncm-dirperm/src/main/pan/components/dirperm/schema.pan
```diff
@@ -30,11 +30,13 @@ function dirperm_permissions_valid = {
...
-    'perm'    : string with match(SELF,'[0-7]{3,4}')
-    'owner'   : string with match(SELF, '\w+(:\w+)?')
-    'type'    : string with match(SELF, 'd|f')
-    'initdir' ? string[]
+    'path'         : string with match(SELF, '^/')
+    'perm'         : string with match(SELF,'[0-7]{3,4}')
+    'owner'        : string with match(SELF, '\w+(:\w+)?')
+    'type'         : string with match(SELF, '^[df]$')
+    'initdir'      ? string[]
+    'checkmount'   : boolean = false
```
stdweird: can you add some documentation as follows
```pan
    @{does X}
    'checkmount'   : boolean = false
    @{behaves like Y}
    'within_mount' : boolean = false
```

(the `@{}` has to come before the attribute that it refers to)

stdweird:
@ajf8 apologies for not noticing the `open TMP` earlier.

ajf8:
I've addressed the remaining comments I hope, sorry for the delay. I tested the new error checking for new files you suggested, by trying to create something into /proc, and it reported the error correctly.

stdweird:
@ajf8 ok. i'm happy  with the current work done. much appreciated

LGTM

PR #628, opened 2015-11-06
ncm-filecopy, ncm-metaconfig: validate file paths

### LINKED ISSUE #627: ncm-filecopy, ncm-metaconfig: Filenames are not validated
We recently had an issue where a filename had a trailing slash causing the component to barf, this should ideally be picked up by schema validation (patch coming soon).

Also affects other components that allow user defined filenames (e.g. ncm-metaconfig).

jrha:
Checks for trailing slashes and throws a helpful error.

Fixes #627, requires quattor/template-library-core#96.

### DIFF
--- a/ncm-filecopy/src/main/pan/components/filecopy/schema.pan
+++ b/ncm-filecopy/src/main/pan/components/filecopy/schema.pan
@@ -38,7 +38,7 @@ type structure_filecopy = {
 
 type component_filecopy = {
     include structure_component
-    'services'    ? structure_filecopy{}
+    'services' ? structure_filecopy{} with valid_absolute_file_paths(SELF)
     'forceRestart' : boolean = false
 };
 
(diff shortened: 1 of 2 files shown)


jrha:
Looks like metaconfig will suffer from the same issue, so perhaps a more general dict key path check would be useful.

jrha:
Should we also insist upon a leading slash?

```
match(SELF, '^/[^\0]+[^/]$')
```

stdweird:
:+1: on test trailing slash, and maybe rename it to something without filecopy, move it to template_library_core so we can use it for metaconfig too

jrha:
@stdweird - moved function to template-library-core and updated

stdweird:
@jrha should i be worried that the tests pass?

jrha:
Nope... :worried:

stdweird:
@ned21 are you ok with this?

PR #629, opened 2015-11-06
The user was hardcoded to be root,but this doesnt work for el7

almumontiel:
I'd like to be able to change the owner of nrpe config file.

### DIFF
--- a/ncm-nrpe/src/main/perl/nrpe.pm
+++ b/ncm-nrpe/src/main/perl/nrpe.pm
@@ -25,7 +25,7 @@ sub Configure {
     my $st = $config->getElement (PATH)->getTree;
 
     my $mode = $config->getElement('/software/components/nrpe/mode')->getValue();
-    my $owner = "root";
+    my $owner = $st->{nrpe_user};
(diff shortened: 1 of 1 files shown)


jrha:
please test this

stdweird:
@almumontiel can you elaborate a bit more on the problem you see on EL7? the group permissions should be enough (we run this on EL7)
(not that i oppose this PR, i can be merged)

almumontiel:
The reason I point el7 is because I only saw this for the first time in el7, _not_ because I found a problem specific of el7. The nrpe daemon, by default, is run by nrpe in el7, whereas in sl6 and sl5 was run by nagios. nrpe does not belong to nagios group by default. For all the machines i have defined NRPE_GROUP=nagios, that is why for el7 started to fail the nrpe daemon start.
Of course, group permissions are enough. Only adding nrpe to nagios group is fine. 
Maybe my title is confusing, it seems like there is something wrong with root owner for nrpe in el7, sorry!
I still find better to be able to be able to define the owner of nrpe.cfg.
Thank you for the quick action.

PR #630, opened 2015-11-10
ncm-icinga: adapt schema for recent icinga 1.X versions

stdweird:
`log_external_commands_user` and `event_profiling_enabled` are not supported in recent icinga 1.X, create a very clear deprecation message (but still allowed)


jrha:
Can you add a description to the PR please.
I presume this is backwards incompatible?

stdweird:
@jrha not really, the schema still allows the values, but icinga will fail (and panc will produce clear messages why).

jrha:
Cool

PR #634, opened 2015-11-18
ncm-altlogrotate: Remove timestamp at top of file

ajf8:
The timestamp at the top of the file isn't especially useful and it causes problems for file monitoring, so remove it.


jrha:
Nice, thanks.

PR #639, opened 2015-11-24
Point all 'ncm' links at corresponding 'readthedocs' pages

GooBall:
Partially resolves quattor/quattor.github.com#128

### DIFF
--- a/ncm-accounts/src/main/perl/accounts.pod
+++ b/ncm-accounts/src/main/perl/accounts.pod
@@ -11,9 +11,9 @@ ncm-accounts: NCM component to manage the local accounts on the machine.
 
 The I<accounts> component manages the local accounts on a machine. LDAP
 authentication depends on the LDAP configuration, which is handled by
-L<ncm-authconfig>.
+L<authconfig|../authconfig/>.
 
-Shadowing of passwords is also controlled by L<ncm-authconfig>.
+Shadowing of passwords is also controlled by L<authconfig|../authconfig/>.
 
 =head1 FUNCTIONS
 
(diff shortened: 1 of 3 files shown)


jrha:
test this please

ncm-accounts/src/main/perl/accounts.pod
```diff
@@ -11,9 +11,9 @@ ncm-accounts: NCM component to manage the local accounts on the machine.
 
 The I<accounts> component manages the local accounts on a machine. LDAP
 authentication depends on the LDAP configuration, which is handled by
-L<ncm-authconfig>.
+L<authconfig|../authconfig/>.
```
stdweird: so how does this work with `man` as regular pod files? `authconfig` actually exists.
shouldn't we do this conversion in https://github.com/quattor/release/pull/105?

jrha:
A fair point. Are the component man pages actually usable or useful?

stdweird:
useful, i'd say yes, since they contain information how the component behaves with having to read most of the code. it certainly helps when modifying methods to read what the method was supposed to do in the first place.
so for developers, it's usable, since they are reading the code anyway (assuming the pods are inlined), but to be really usable, we need https://github.com/quattor/release/pull/105

if the pod format ideal for the intended usage: probably not, but it's at least a standard and it can be converted to other formats.

jrha:
What I mean is, are they more useful as man pages or html on readthedocs, I for one have never looked at the generated man pages and I'm not sure `man` can even find them on our systems.

stdweird:
@jrha they are more useful as readthedocs for sure. but that needs conversion anyway, so i would keep the pods as pods, and handle the conversion elsewhere.
if you are volunteering to rewrite all pods in whatever endformat you prefer, that's also ok. but something in between is not needed i think.

PR #640, opened 2015-11-24
Change 'aii' links to render as code instead of links

GooBall:
Currently the `aii` links are dead ends and there is no correct link to point them at.

The following changes render them as code, rather than links. They can become links again, once they have have something to link to. 

Fixes quattor/quattor.github.com#128.


jrha:
test this please

PR #644, opened 2015-12-03
ncm-hostsfile: fix removal of old entries which were managed by NCM.

ajf8:
The logging declares its intent to remove some entries which were
previously managed by NCM, but then adds them into the file output
instead of skipping them.

### DIFF
--- a/ncm-hostsfile/src/main/perl/hostsfile.pm
+++ b/ncm-hostsfile/src/main/perl/hostsfile.pm
@@ -218,7 +218,7 @@ sub Configure {
             if (exists $newncmhosts{$h}) {
                 $le = delete $newncmhosts{$h};
             } elsif (exists $ncmhosts{$h}) {
-                $le = delete $ncmhosts{$h};
+                next;
             } else {
                 $le = delete $non_ncm_hosts{$h};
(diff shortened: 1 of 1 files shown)


ncm-hostsfile/src/main/perl/hostsfile.pm
```diff
@@ -218,7 +218,7 @@ sub Configure {
             if (exists $newncmhosts{$h}) {
                 $le = delete $newncmhosts{$h};
             } elsif (exists $ncmhosts{$h}) {
-                $le = delete $ncmhosts{$h};
+                next;
```
stdweird: shouldn't it also delete the entry from `$ncmhosts` to indicate it was processed? (i tried to read the code a bit, i'm not that familiar)

ajf8:
No, I don't think that is needed, at this point the purpose of %ncmhosts is just to know what has been removed and doesn't get used again.

%ncmhosts is populated with existing entries (that have been tagged with the NCM comment), loop over the profile removes these, leaving only entries to be deleted.

stdweird:
@ajf8 ok
@jrha LGTM

PR #645, opened 2015-12-04
ncm-ssh: make config checks configurable.

ajf8:
small oversight on my part in the previous PR regarding this component during build time for us. sshd is on a network FS which isn't available at build time, so I've added a new param where you can make this check skip if sshd isn't existing.

existing behaviour, to always check, is still the default.

### DIFF
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -117,6 +117,8 @@ type ssh_daemon_type = {
     "options" ? ssh_daemon_options_type
     "comment_options" ? ssh_daemon_options_type
     "sshd_path" ? string
+    @{if false and sshd doesn't exist, skip config validation}
+    "always_validate" : boolean = true
     "config_path" ? string
 };
 
--- a/ncm-ssh/src/main/perl/ssh.pm
+++ b/ncm-ssh/src/main/perl/ssh.pm
@@ -30,9 +30,16 @@ sub valid_sshd_file
 {
     my ($self, $file, $cfg) = @_;
 
+    my $sshd_bin = $cfg->{sshd_path} || DEFAULT_SSHD_PATH;
+    
+    if (defined($cfg->{always_validate}) && !$cfg->{always_validate} && ! -x $sshd_bin) {
+        $self->info("$sshd_bin doesn't exist with always_validate=0, skipping sshd config test.");
+        return 1;
+    }
+
     # Use /dev/stdin, instead of /proc/self/fd/0 (which is used in some other components).
     # This is because /proc/self/fd/0 does not exist on Solaris. 
-    my $cmdline = [ $cfg->{sshd_path} || DEFAULT_SSHD_PATH, '-t', '-f', '/dev/stdin' ];
+    my $cmdline = [ $sshd_bin, '-t', '-f', '/dev/stdin' ];
 
     my $cmd = CAF::Process->new(
         $cmdline,


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -117,6 +117,7 @@ type ssh_daemon_type = {
     "options" ? ssh_daemon_options_type
     "comment_options" ? ssh_daemon_options_type
     "sshd_path" ? string
+    "validate_config" ? string with match (SELF, '^(yes|no|optional)$')
```
stdweird: please use a boolean to validate or not. 
and use another boolean to allow to skip the validation if sshd is missing (but if `sshd` is missing ,you cannot validate anyway, so this feels like the boolean means something else) 

but why would one ever choose not to validate? and why would we allow that?
i think one boolean to indicate that the test can be skipped if sshd is missing is enough.
btw, what sort of system is this that there is no sshd available? (maybe not the one that you will run, but can't you use a system sshd to validate the config if the one from the network FS is missing? imho it's better than just skipping the config test)

ajf8: OK, I've changed it to a boolean called validate_optional. Disabling the checks completely has been removed.

Our sshd is a component which happens to be managed in the network filesystem, installing another through a package isn't something we would want to do. The config validation is very nice to have after build, but we've done without it previously and it would be easiest to just allow the check to skip at build time.

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -117,6 +117,8 @@ type ssh_daemon_type = {
     "options" ? ssh_daemon_options_type
     "comment_options" ? ssh_daemon_options_type
     "sshd_path" ? string
+    # if true and sshd doesn't exist, will skip config validation
```
ned21: Can you add this to the pod too please? I think "validation_optional" would be a more correct name too? Or we could reverse the sense and say ```"always_validate" : boolean = true``` ?

stdweird: or put the comment in `@{}`, so it will at least be picked up by the annotations

@ned21 but what is the inverse/opposite of always? never or optional

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -30,9 +30,17 @@ sub valid_sshd_file
 {
     my ($self, $file, $cfg) = @_;
 
+    my $sshd_bin = $cfg->{sshd_path} || DEFAULT_SSHD_PATH;
+    my $validate_optional = $cfg->{validate_optional};
+    
+    if (defined($validate_optional) && $validate_optional && ! -x $sshd_bin) {
```
stdweird: there'no need for the `defined()` check (and than neither for the new variable). `if ($cfg->{validate_optional} && ...)` does the same thing

ncm-ssh/src/main/perl/ssh.pm
```diff
@@ -30,9 +30,17 @@ sub valid_sshd_file
...
+        $self->info("$sshd_bin doesn't exist, skipping sshd config test.");
```
stdweird: maybe add `with validate_optional=1` (the message best expresses why this is skipped, and it's not only because sshd is missing)

ncm-ssh/src/main/pan/components/ssh/schema.pan
ned21: The opposite of always is never but "never != (not always)".  Or to put it another way: always = false does not imply never.

ajf8:
Update with the latest feedback addressed hopefully

jrha:
LGTM

stdweird:
LGTM

PR #649, opened 2015-12-14
ncm-afsclt: remove out of place code for managing PAM and iptables.

ajf8:
It is better serviced in the proper dedicated components.

Our local version has diverged a lot from the upstream, which hasn't
been updated in many years. This patch adds NoAction support and a few
other fixes/features, but it also throws a lot of code for managing PAM
and iptables rules.

I think those features may only have used by one organisation which no
longer uses quattor.


ned21:
"Sync to our local version" isn't very descriptive. Perhaps: "Remove out of place code for managing PAM and iptables that is better serviced in the proper dedicated components" ?

ncm-afsclt/src/main/pan/components/afsclt/schema.pan
```diff
@@ -1,19 +1,17 @@
...
 # ${developer-info}
 # ${author-info}
-
-# Coding style: emulate <TAB> characters with 4 spaces, thanks!
-################################################################################
+# ${build-info}
 
 declaration template components/afsclt/schema;
 
 include { 'quattor/schema' };
```
stdweird: no `{}`, only include the parts you need

ncm-afsclt/src/main/pan/components/afsclt/schema.pan
```diff
@@ -1,19 +1,17 @@
...
 
 type component_afsclt_entry = {
-#   include component_type
-    include structure_component
-
+   include structure_component
+   
    "thiscell"   : string         # "AFS home cell"
+   "thesecells" ? list           # "Cell list to authenticate to"
```
stdweird: list of what?

ncm-afsclt/src/main/pan/components/afsclt/schema.pan
```diff
@@ -1,19 +1,17 @@
...
    "suidcells"  ? list           # "List of suid cells"
```
stdweird: list of what?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -15,74 +13,40 @@ use strict;
...
 $NCM::Component::afsclt::NoActionSupported = 1;
 
-# prevent authconfig from trying to launch in X11 mode
-delete($ENV{"DISPLAY"});
-
 #*** The following gives the OS name using the LC perl modules.
 #*** In example, $OSname can be 'Linux' or 'Solaris'
 #*** It's equivalent to perl's variable $^O (it gives the OS in lower-case)
 use LC::Sysinfo;
 my $OSname=LC::Sysinfo::os()->name;
```
stdweird: is this used somewhere? maybe you can use `CAF::Service::os_flavour`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -15,74 +13,40 @@ use strict;
...
 use vars qw(@ISA $EC);
 @ISA = qw(NCM::Component);
 $EC=LC::Exception::Context->new->will_store_all;
-
 $NCM::Component::afsclt::NoActionSupported = 1;
 
-# prevent authconfig from trying to launch in X11 mode
-delete($ENV{"DISPLAY"});
-
 #*** The following gives the OS name using the LC perl modules.
```
stdweird: don't hesitate to cleanup the funky comment `#***` -> `#` :smile:

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -103,7 +67,7 @@ sub Configure_Cell {
     close(THISCELL);
```
stdweird: FileReader?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
-      $self->warn("Cannot find $libpam (=$libpamtmp), will not configure PAM");
-      return 0;
-    }
-
-    my $pam_systemauth = "/etc/pam.d/system-auth";
-    my $pam_screensaver = '/etc/pam.d/screensaver-auth'; #optional
-
-    if ( ! -e "$pam_systemauth") {
-      $self->warn("Cannot find $pam_systemauth - not touching it");
+  my $thesecells = "/usr/vice/etc/TheseCells";
```
stdweird: 4 space indent

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
+
+  if ($config->elementExists($mypath."/thesecells")) {
+    if( ! $NoAction ) {
```
stdweird: FileWriter.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
+      my $cells = $config->getElement($mypath."/thesecells")->getTree();
+      unless (open(THESECELLS, ">", $thesecells)) {
+        $self->error("cannot write $thesecells: $!");
+        return;
+      }
+      foreach my $cell ( @$cells ) {
+        print THESECELLS $cell." ";
```
stdweird: like `join(" ", @$cells)`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
+      }
+      print THESECELLS "\n";
+      close(THESECELLS);
+      $self->OK("Configured cell list for authentication $thesecells");
```
stdweird: `OK`? no `info`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
+      $self->info("Would configure $thesecells");
     }
-
-    # a special PAM file for screensavers isn't always required
-    # silently ignore it if missing and we have no special options to set
-    if (! -e "$pam_screensaver") {
-      if (! $libpam_options_refresh ) {
-         $self->debug(1, "No $pam_screensaver and no special options for it, ignoring");
+  } elsif ( -f $thesecells ) {
+    if( ! $NoAction ) {
```
stdweird: no negative conditions in if/else, just do 
```perl
if ($NoAction) {
  would remove
} else {
  unlink
}
```

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -128,361 +92,46 @@ sub Configure_Cell {
...
+      if ( unlink($thesecells) ) {
+        $self->OK("Removed cell list for authentication $thesecells");
```
stdweird: info?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -503,30 +152,20 @@ sub Configure_Cache {
...
-  my $automatic_cache = 0;
+  my $file_afsmount = '';
 
   my $ret = 1;
 
   if ($config->elementExists($mypath."/cachesize")) {
     $new_cache = $config->getValue($mypath."/cachesize");  #new cache size
   } else {
-    $self->warn("cannot get CDB $mypath/cachesize, giving up");
+    $self->warn("cannot get CDB $mypath/cachesize - not setting cache size");
```
stdweird: what does CDB mean?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -503,30 +152,20 @@ sub Configure_Cache {
...
-  if ($config->elementExists($mypath."/cachemount")) {
-    my $cdb_cachemount=$config->getValue($mypath."/cachemount"); # new cache mount, if any
-    if ( $cdb_cachemount =~ m;^(/[.\w\/-]+)$; ) {
-      $new_cachemount=$1;
-    } else {
-      $self->warn("CDB new AFS cache mount point $cdb_cachemount looks weird");
-    }
-  }
-
   my $fs_run_cache_string =`fs getcacheparms 2>/dev/null`;
```
stdweird: CAF::Process (with keeps_state?) (and can you check the code for more backticks or `qx//` or `system`? there's no real point in making NoAction effort if you don't use the CAF tools

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -536,112 +175,64 @@ sub Configure_Cache {
...
-  close(CI);
-  if ($t[0] =~ m;^([^:]+):([^:]+):(\d+)$; ) {
-    my $afsmount = $1;
-    $file_cachemount = $2;
-    $file_cache = $3;
-    if($afsmount ne "/afs") {
-      $self->error("cannot handle non-/afs mount point $afsmount in $afs_cacheinfo: ".$t[0]);
-      return 0;
-    }
+  if (open(CI, '<'.$afs_cacheinfo)) {
```
stdweird: FileReader

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -536,112 +175,64 @@ sub Configure_Cache {
...
-  } else {
-    $self->warn("cannot run 'df' on the AFS cache?");
-  }
-
-  # OpenAFS-1.4/Linux takes the info from /etc/sysconfig/afs
-  if( -r $sysconfig_afs) {
-    if($automatic_cache) {
-      $new_cache = "AUTOMATIC";
+  # sanity check - don't allow cachesize bigger than 95% of a partition size
+  my @df=`df -k $file_cachemount 2>/dev/null`;
```
stdweird: backticks.....

stdweird: @kwaegema might be nice thing to add to lib-blockdevices? (determine size from mounted FS via `df -k`)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -536,112 +175,64 @@ sub Configure_Cache {
...
   }
 
   # adjust stored cache size (gets overwritten on restart for OpenAFS-1.4)
-  if ($new_cache ne "AUTOMATIC") {
-    if ($new_cache != $file_cache) {
-      if(! open (CI, '>'.$afs_cacheinfo)) {
-	$self->error("cannot open $afs_cacheinfo for writing: $!");
-	return 0;
+  if ($new_cache != $file_cache) {
+      if (! $NoAction) {
```
stdweird: FileWriter

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -536,112 +175,64 @@ sub Configure_Cache {
...
-    my $fscommand = "fs setcachesize $new_cache";
-    my $fsout=`$fscommand 2>&1`;
-    if ($? >> 8) {
-      $self->warn("problem changing running AFS cache size via \"$fscommand\":\n$fsout");
-    } else {
-      $self->OK("changed running AFS cache $run_cache -> $new_cache (1K blocks)");
-    }
+      if (! $NoAction) {
+	  my $fscommand = "fs setcachesize $new_cache";
+	  my $fsout=`$fscommand 2>&1`;
```
stdweird: grrrrrr.....

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -654,50 +245,22 @@ sub Configure_Cache {
...
-	return 0;
-      }
-    } elsif ( $master_cellservdb_cdb =~ m;^((ftp|http|file)://.+);i) { # known URIs - could contain other stuff i.e. for programmatic access
-      $master_cellservdb = $1;
-    } else {
-      $self->warn("weird/unsafe cellservdb URL/PATH: $master_cellservdb, giving up");
+    $master_cellservdb = $config->getValue($mypath."/cellservdb");
+    if ( $master_cellservdb =~ m/^\/\w/i) { # non-URI -> file
+      $master_cellservdb = "file://".$master_cellservdb;
+    } elsif ( $master_cellservdb !~ m/^(ftp|http|file)/i) { # known URIs
```
stdweird: `https?`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -754,7 +317,7 @@ sub update_afs_cells ( $$ ) {
     @todo = ();
     my @current = `fs listcell -numeric 2>&1`;
```
stdweird: Process with keeps_state

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -781,18 +344,22 @@ sub update_afs_cells ( $$ ) {
...
-        $error = 0;
-        foreach $cell (@todo) {
-            my $out=`fs newcell $cell @{$ipaddrs{$cell}} 2>&1`;
-	    if($? >> 8) {
-	      $self->Debug("error while updating AFS cell info for $cell: $out");
-	      $error++;
+	if (! $NoAction) {
+	    $error = 0;
+	    foreach $cell (@todo) {
+		my $out=`fs newcell $cell @{$ipaddrs{$cell}} 2>&1`;
```
stdweird: guess what...

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
-      return;
-    }
-    $self->OK("Unconfigured AFS client");
+  my $path = "$mypath/afsd_args";
+  if ($config->elementExists($path)) {
+      if (! $NoAction) {
+	  my $args = $config->getElement($path)->getTree();
+	  my $content;
+	  $content .= "$_:" . $args->{$_} . "\n" for sort keys %$args;
+	  my $fh = CAF::FileWriter->new("/etc/afsd.args", log => $self, backup => ".old");
```
stdweird: `CAF::FileWriter` is NoAction safe

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
-    $self->info("iptables will not be unconfigured: error opening file $iptables: $!");
-    return 0;
-  }
 
-  my @iptables = <FD>;
-  close(FD);
-  my @newtables;
-  my $return = 1;
-  my $found = 0;
+=head1 NAME
```
stdweird: move this to the top of the module and it's respective methods. it is much better to keep the pod close to the code

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
+=over
 
-##########################################################################
-sub Unconfigure_Cache {
-##########################################################################
-  my ($self,$config)=@_;
-  $self->info("AFS cache config not changed");
-  return 1;
-}
+=item /software/components/afsclt/active : boolean
```
stdweird: this is a component, no need to document this for each component

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
-  my $libpam;
-  my $libpam_options_auth;
-  my $libpam_options_refresh;
-  my $ret = 0;
+activates/deactivates the component.
 
-  if ($OSname =~ /Solaris/) {
-    $self->warn("Somebody needs to write something to unconfigure PAM on Solaris"); #FIXME
-    return 0;
+=item /software/components/afsclt/thiscell : string
```
stdweird: documentation of the schema is best done via annotations in the schema. so this could be
```pan
type ... {
    ...
    @{local AFS cell for this machine}
     "thiscell" ? string
```
put the `@{}` block before the attribute

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
+
+=item /software/components/afsclt/afsd_args : nlist
+
+nlist of arguments to stash in /etc/afsd.args used by rc.afs.
+
+=back
+
+=head1 BUGS
+
+The "debug" and "suidcell" parameters are silently ignored for now.
```
stdweird: add deprecation warnings in the schema
```pan
    "suidcell" ? string with { deprecated(0, "suidcell is silently ignored"); true;}
```

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
+
+A regularly-updated AFS CellServDB URL or filename (e.g. from AFS)
+that this component will copy to local disk. The local AFS client will
+get notified of any additions or changes within a cell.
+
+=item /software/components/afsclt/afsd_args : nlist
+
+nlist of arguments to stash in /etc/afsd.args used by rc.afs.
+
+=back
```
stdweird: delete this stuff, just make sure that jaroslav is in the authors in the pom.xml (and maybe add yourself as maintainer?)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
+      }
   }
-  return; 
 }
 
-##########################################################################
-sub Unconfigure_firewall {
-##########################################################################
-  my ($self,$config)=@_;
+1; #required for Perl modules
```
stdweird: this should be the last (empty) line in the component.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
-      $self->error("can't run $s, no changes made");
-      return;
-    }
-    $self->OK("Unconfigured AFS client");
+  my $path = "$mypath/afsd_args";
+  if ($config->elementExists($path)) {
+      if (! $NoAction) {
+	  my $args = $config->getElement($path)->getTree();
+	  my $content;
+	  $content .= "$_:" . $args->{$_} . "\n" for sort keys %$args;
```
stdweird: why no `print $fh "$_:" . $args->{$_} . "\n" for sort keys %$args;`, the extra variable makes no sense

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -801,197 +368,113 @@ sub update_afs_cells ( $$ ) {
...
+	  my $fh = CAF::FileWriter->new("/etc/afsd.args", log => $self, backup => ".old");
+	  print $fh $content;
+	  $fh->close();
+	  $self->OK("Configured afsd.args with $content");
```
stdweird: info?

stdweird:
@ajf8 some advise in cleaning up: the CAF::Filewriter/Editor/Reader and Process are NoAction safe (won't do anything if NoAction is set)
for `CAF::Process` there exists a `keeps_state => 1` option to add during init, that will run a command even if NoAction is set (this is usefull to run harmless stuff like list something, so the component can proceed as much as allowed during NoAction)
i also do not know which version of CCM this is targeting, but as of 15.6, you can do `$config->getTree(/some/path)`, which returns undef if paths doesn't exists, so no more $config->elemenetexists / config->getvalue constructions

stdweird:
@ajf8  and thanks for the effort!

ajf8:
Thanks a lot for the comments @stdweird, seems this module needed some attention, so I've tried to go through it all and tidy things up in general as well as addressing your comments.

ncm-afsclt/src/main/perl/afsclt.pm
ajf8: I decided to just delete these unimplemented options instead.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
 my $afs_cacheinfo = '/usr/vice/etc/cacheinfo';
-my $localcelldb = "/usr/vice/etc/CellServDB";
-
-my $mypath='/software/components/afsclt'; # this could be rather in /system/filesystems/remote ?
-my $compname = "NCM-afsclt";
-
-#***    In Solaris the fs executable at /usr/afsws/bin
-#*** so we put that directory first in the path.
-if ($OSname=~ /Solaris/) {$ENV{'PATH'}="/usr/afsws/bin:$ENV{'PATH'}"}
+my $localcelldb   = '/usr/vice/etc/CellServDB';
```
stdweird: if these are constants, use uppercase (and even better, readonly)
```perl
use Readonly;
Readonly my $LOCALCELLDB => '/usr/vice/etc/CellServDB';
```

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
-
-  if($config->elementExists("$mypath/enabled")) {
-    $node = $config->getElement("$mypath/enabled");
-    my $enabled = $node->getValue(); #syntax needs to be on/off, not yes/no
-    if ($enabled =~ /^off$|^no$/i) {
-      $enabled = 'off';
-    } else {
-      $enabled = 'on'; #default
+    my $proc = CAF::Process->new( [ "fs", "getcacheparms" ], log => $self, keeps_state => 1 );
+    if ( !( $? >> 8 ) && $proc->output() =~ /AFS using \d+ of the cache's available (\d+) (\w+) byte blocks/ ) {
```
stdweird: can't check the exitcode before running something
```perl
my $output = CAF::Process->new( [ "fs", "getcacheparms" ], log => $self, keeps_state => 1 )->output();
if (!( $? >> 8 ) && $output && $output =~ /AFS using \d+ of the cache's available (\d+) (\w+) byte blocks/ ) {
...
```

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+        if ( $2 ne "1K" ) {
+            $self->error("Cannot handle $2 (non-1K) AFS cache block sizes");
+            return 0;
+        }
+        $run_cache = $1;
+    }
+    else {
+        $self->info("Cannot determine current AFS cache size, changing only config file");
```
stdweird: isn't this at least `warn`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+    }
+
+    my $afs_cacheinfo_fh = CAF::FileEditor->new( $afs_cacheinfo, log => $self );
```
stdweird: use `CAF::FileReader` if you are not going to modify anything

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+    else {
+        $self->error("Cannot parse stored AFS cache mount or size from $afs_cacheinfo");
+        return 0;
+    }
+    $afs_cacheinfo_fh->close();
+
+    # sanity check - don't allow cachesize bigger than 95% of a partition size
+    $proc = CAF::Process->new( [ "df", "-k", $file_cachemount ], log => $self, keeps_state => 1 );
+    for ( split ( "\n", $proc->output() ) ) {
+        if (m{^(.*?)\s+(\d+)\s+(\d+)\s+(\d+)\s+(\d+%)\s+(.*)}) {
```
stdweird: don't group the patterns you don' need. it makes it much harder to read/interpret

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+    }
+    else {
+        $self->error("Cannot parse stored AFS cache mount or size from $afs_cacheinfo");
+        return 0;
+    }
+    $afs_cacheinfo_fh->close();
+
+    # sanity check - don't allow cachesize bigger than 95% of a partition size
+    $proc = CAF::Process->new( [ "df", "-k", $file_cachemount ], log => $self, keeps_state => 1 );
+    for ( split ( "\n", $proc->output() ) ) {
```
stdweird: better not use `$_` in perl, it really does not behave as you expect. this better becomes
```perl
foreach my $line split ( "\n", $proc->output() ) ) {
   if ($line =~ m{...}) {
...
```
(`$_` is good for oneliners like `grep {$_ eq 'something} @array`, but for eveything else, keep it readable and use a variable)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+    if ( $run_cache && ( $new_cache ne "AUTOMATIC" ) && ( $run_cache != $new_cache ) ) {
+        if ($NoAction) {
+            $self->info("Would need to update running cache size");
+        }
+        else {
+            my $fscommand = [ "fs", "setcachesize", $new_cache ];
+            $proc = CAF::Process->new( $fscommand, log => $self );
+            $proc->run();
+            if ( $? >> 8 ) {
+                $self->warn( "Problem changing running AFS cache size via \"" . join ( " ", @$fscommand ) . "\":\n" + $proc->output() );
```
stdweird: `CAF::Process` has stringification, so instead of `join ( " ", @$fscommand )`, you can just `"$proc"`
`proc->output()` will run the command a second time; so better to `$output = $proc->output()` instead of `->run` (i also think `run` dumps output to stdout, which noone wants)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
-  # adjust online (in-kernel) value
-  if($run_cache && ($new_cache ne "AUTOMATIC") && ($run_cache != $new_cache)) {
-    my $fscommand = "fs setcachesize $new_cache";
-    my $fsout=`$fscommand 2>&1`;
-    if ($? >> 8) {
-      $self->warn("problem changing running AFS cache size via \"$fscommand\":\n$fsout");
-    } else {
-      $self->OK("changed running AFS cache $run_cache -> $new_cache (1K blocks)");
+    else {
+        LC::Check::file(
```
stdweird: use `FileWriter`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
-    # take it from the real file
-    if( ! open(*FH, $localcelldb)) {
-      $self->warn("cannot read local $localcelldb, giving up");
-      return 0;
-    }
+    my ($self) = @_;
+    my ( $cell, %seen, %ipaddrs, @hosts, $todo, @todo, $error, $afsutil );
 
-    while (<FH>) {
+    my $localcelldb_fh = CAF::FileEditor->new( $localcelldb, log => $self );
```
stdweird: `FileReader` if you are only reading

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
         }
     }
-    close(FH);
-    foreach $cell (keys(%ipaddrs)) {
-        @{$ipaddrs{$cell}} = sort(@{$ipaddrs{$cell}});
-        $self->Debug("CellServDB $cell -> @{$ipaddrs{$cell}}");
+    $localcelldb_fh->close();
+    foreach $cell ( keys(%ipaddrs) ) {
+        @{ $ipaddrs{$cell} } = sort( @{ $ipaddrs{$cell} } );
+        $self->debug( 1, "CellServDB $cell -> @{$ipaddrs{$cell}}" );
```
stdweird: does this print anythin sensible?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+
     # read known cells
     @todo = ();
-    my @current = `fs listcell -numeric 2>&1`;
+    my $proc = CAF::Process->new( [ "fs", "listcell", "-numeric" ], log => $self, keeps_state => 1 );
 
-    if($? >> 8 ) {
-      $self->info("cannot read current AFS cell info via fs:\n".join('',@current));
-      return 1;
+    if ( $? >> 8 ) {
```
stdweird: don't chekc exitcode before output. creating the `CAF::Process` instance does not run anything.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+        $self->error( "Cannot read current AFS cell info via fs:\n" . $proc->output() );
+        return 1;
     }
-    foreach (@current) {
+
+    foreach ( split ( /\n/, $proc->output() ) ) {
```
stdweird: use `my $line` instead of relying on `$_`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+        @hosts = split ( /\s+/, $2 );
         $seen{$cell} = 1;
-        $self->Debug("fs listcell: $cell -> @hosts");
-        next unless $ipaddrs{$cell} && @{$ipaddrs{$cell}};
+        next unless $ipaddrs{$cell} && @{ $ipaddrs{$cell} };
         @hosts = sort(@hosts);
-        if ("@hosts" ne "@{$ipaddrs{$cell}}") {
-            push(@todo, $cell);
-            $self->Debug("cell info for $cell changed from @hosts to @{$ipaddrs{$cell}}");
+        if ( "@hosts" ne "@{$ipaddrs{$cell}}" ) {
```
stdweird: does this work? what does it compare?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
+        if ($NoAction) {
+            $self->info("Would need to configure new cells @todo");
+        }
+        else {
+            $error = 0;
+            foreach $cell (@todo) {
+                $proc = CAF::Process->new( [ "fs", "newcell", $cell, @{ $ipaddrs{$cell} } ], log => $self );
+                $proc->run();
+                if ( $? >> 8 ) {
+                    $self->error( "Error while updating AFS cell info for $cell: " . $proc->output() );
```
stdweird: will run the command a second time. there is also no need for the `NoAction` block, `CAF::Process` is safe to use

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,314 @@
...
-    } elsif ($line =~ /^AFS_CLIENT=/ ) {
-      $self->debug(2, "resetting AFS_CLIENT enable");
-      $line = "AFS_CLIENT=on\n";
-      $found =1;
-    } else {
-      $self->debug(2, "copying existing line");
+    my $path = "$mypath/afsd_args";
+    if ( $config->elementExists($path) ) {
+        my $args = $config->getElement($path)->getTree();
+        my $fh = CAF::FileWriter->new( "/etc/afsd.args", log => $self, backup => ".old" );
```
stdweird: the `"/etc/afsd.args"` should be a Readonly / constant

stdweird:
@ajf8 creating `CAF::Process` instances does not do anything, you need to `->output()` or `->execute()`. (i think `->run()` does execute with everything  redirected to stdout (and you don't want command output to stdout)

but almost there :smile:

ncm-afsclt/src/main/perl/afsclt.pm
ajf8: Yeah, it is comparing the IP addresses in the current CellServDB and the new one to determine if IP addresses have changed.

I guess this is a simple way of comparing two arrays by using their string representation, it does seem to work OK.

From 5.10 onwards the ~~ "smartmatch operator" is probably better. As we have 5.10 on RHEL4, I suppose this is a possibility.

ncm-afsclt/src/main/perl/afsclt.pm
ajf8: It prints what was parsed out of the existing cell DB, I don't mind dropping it.

ncm-afsclt/src/main/perl/afsclt.pm
stdweird: no it seems to do space separated join, so for IP addresses this is safe.
```
# perl -e '@hosts=qw(a b c);$b={x => [qw(abc)]}; print "@hosts @{$b->{x}}\n"'
a b c abc
```
learned something new :wink:

ncm-afsclt/src/main/perl/afsclt.pm
stdweird: leave it, i was curioes what the stringification of an array did.

ncm-afsclt/src/main/pan/components/afsclt/schema.pan
```diff
@@ -24,18 +21,7 @@ type component_afsclt_entry = {
...
-   "libpam_options_auth_auth" ? string # "PAM options for authentication, PAM-auth"
-   "libpam_options_auth_session" ? string # "PAM options for authentication, PAM-session"
-   "libpam_options_auth_passwd" ? string # "PAM options for authentication, PAM-passwd"
-   "libpam_options_auth" ? string # "PAM options for authentication, PAM-session etc"
-   "libpam_options_refresh" ? string # "PAM options for refresh"	
-
-   "verbose"    ? string
-   "debug"      ? boolean
-   
+   "afsd_args"              ? nlist # /etc/afsd.args values for rc.afs
```
stdweird: can you be a bit more specific? like `string{}`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
+Readonly my $AFS_CACHEINFO => '/usr/vice/etc/cacheinfo';
+Readonly my $LOCALCELLDB   => '/usr/vice/etc/CellServDB';
+Readonly my $THISCELL      => '/usr/vice/etc/ThisCell';
+Readonly my $THESECELLS    => '/usr/vice/etc/TheseCells';
+Readonly my $AFSD_ARGS     => '/etc/afsd.args';
 
-#***    In Solaris the fs executable at /usr/afsws/bin
-#*** so we put that directory first in the path.
-if ($OSname=~ /Solaris/) {$ENV{'PATH'}="/usr/afsws/bin:$ENV{'PATH'}"}
+my $mypath = '/software/components/afsclt';
```
stdweird: also a readonly? (in principal, you can use `$self->prefix()` for this value, but it's used quite a lot, so variable is fine)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
 sub Unconfigure {
-##########################################################################
-  my ($self,$config)=@_;
-  $self->Unconfigure_Cache($config);
-  if ($OSname=~ /Linux/) {
-      $self->Unconfigure_Config($config);
-      $self->Unconfigure_firewall($config);
-      $self->Unconfigure_Cell($config);
-  }
+    my ( $self, $config ) = @_;
```
stdweird: remove the `Unconfigure` completely if it is not supported

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-  } elsif (! $NoAction) {
-    if ($current_cell) {
-      $self->info("replacing '$current_cell' with '$afscell' in $thiscell");
-    }
-    unless (open(THISCELL, ">", $thiscell)) {
-      $self->error("cannot write $thiscell: $!");
-      return;
+    my ( $self, $config ) = @_;
+    unless ( $config->elementExists("$mypath/thiscell") ) {
+        $self->error("Cannot get $mypath/thiscell");
```
stdweird: make the message clearer that this is a problem in the profile, not woth the component or the system itself (as this is mandatory in the schema)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-    $self->OK("Configured AFS client for cell $afscell in $thiscell");
-  } else {
-    $self->info("would need to update $thiscell\n");
-  }
+    my $afscell = $config->getValue("$mypath/thiscell");
 
-  return;
-}
+    my $thiscell_fh = CAF::FileWriter->new( $THISCELL, log => $self );
+    print $thiscell_fh $afscell . "\n";
```
stdweird: maybe `"$afscell\n"`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-				    return $expected;
-				   }
-			  );
+sub Configure_Config {
+    my ( $self, $config ) = @_;
+
+    # cache size is configured in a separate function
+    if ( $config->elementExists("$mypath/afsmount") ) {
+        my $new_afsmount = $config->getValue("$mypath/afsmount");
+        if ( $new_afsmount ne '/afs' ) {
```
stdweird: if only one value is allowed, also enforce it in the schema?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
+            $self->error("Cannot handle non-/afs AFS mount point $new_afsmount");
```
stdweird: this is a profile error, make it clear in the message.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-      my $newtables = join('',@newtables);
-
-      $self->debug(3, "new $iptables:\n".$newtables);
-      $return = (LC::Check::file($iptables,
-			     "contents"    => $newtables
-			    )
-	    );
+    else {
+        $self->info("Cannot get CDB $mypath/cachesize - not setting cache size");
+        return;
```
stdweird: if the exitcodes are menaingful, this is success and maybe `return 1`?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
+            $self->error("Cannot handle $2 (non-1K) AFS cache block sizes");
+            return 0;
+        }
+        $run_cache = $1;
+    }
+    else {
+        $self->warn("Cannot determine current AFS cache size, changing only config file");
+    }
+
+    my $afs_cacheinfo_fh = CAF::FileEditor->new( $AFS_CACHEINFO, log => $self );
```
stdweird: `CAF::FileReader`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
+    $afs_cacheinfo_fh->close();
+
+    # sanity check - don't allow cachesize bigger than 95% of a partition size
+    $proc = CAF::Process->new( [ "df", "-k", $file_cachemount ], log => $self, keeps_state => 1 );
+    foreach my $line ( split ( "\n", $proc->output() ) ) {
+        if ($line =~ m{^.*?\s+(\d+)\s+\d+\s+\d+\s+\d+%\s+(.*)}) {
+            my $disk_cachesize = $1;
+            my $mount          = $2;
+            if ( $mount eq $file_cachemount && $new_cache > 0.95 * $disk_cachesize ) {
+                $self->error("Cache size cannot exceed 95% of its partition size. Not changing.");
```
stdweird: add mount and disk_cachesize  to the message?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-	return 0;
-      }
-      print CI "/afs:$file_cachemount:$new_cache\n";
-      close(CI) or
-	$self->warn("cannot close $file_cachemount: $!");
-      $self->OK("changed AFS cache config file $afs_cacheinfo: $file_cachemount $run_cache -> $new_cache (1K blocks)");
-    } else {
-      $self->info("AFS cache size is unchanged ($new_cache K)");
+
+    if ($NoAction) {
```
stdweird: the `NoAction` is only needed for the `update_afs_cells` (or is that method alread `NoAction` safe on itself?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-      $self->warn("cannot read local $localcelldb, giving up");
-      return 0;
-    }
+    my ($self) = @_;
+    my ( $cell, %seen, %ipaddrs, @hosts, $todo, @todo, $error, $afsutil );
 
-    while (<FH>) {
-        if (/^>(.\S+)/) {
-            # new cell
+    my $localcelldb_fh = CAF::FileEditor->new( $LOCALCELLDB, log => $self );
```
stdweird: `FileReader`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-    local(*FH, $_);
-    # init
-
-    # read CellServDB
-    # take it from the real file
-    if( ! open(*FH, $localcelldb)) {
-      $self->warn("cannot read local $localcelldb, giving up");
-      return 0;
-    }
+    my ($self) = @_;
```
stdweird: i think this method is `NoAction` safe on it's own. no need for the `if` construct above

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-        if (/^>(.\S+)/) {
-            # new cell
+    my $localcelldb_fh = CAF::FileEditor->new( $LOCALCELLDB, log => $self );
+    foreach my $line ( split ( /\n/, "$localcelldb_fh" ) ) {
+        if ($line =~ /^>(.\S+)/) {
             $cell = $1;
             $ipaddrs{$cell} = [];
-        } elsif (/^(\S+)\s+\#\s*(\S+)/) {
+        }
+        elsif ($line =~ /^(\S+)\s+\#\s*(\S+)/) {
```
stdweird: no need for the second group

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-      return 1;
+    if ( $? >> 8 ) {
+        $self->error( "Cannot read current AFS cell info via fs: $output" );
+        return 1;
     }
-    foreach (@current) {
-        chomp($_);
-        next unless /^Cell\s+(\S+)\s+on hosts\s+(\S.+)\.$/;
+
+    foreach my $line ( split ( /\n/, $proc->output() ) ) {
```
stdweird: just `$output`

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,306 @@
...
-            $self->Debug("cell info for $cell changed from @hosts to @{$ipaddrs{$cell}}");
+        if ( "@hosts" ne "@{$ipaddrs{$cell}}" ) {
+            push ( @todo, $cell );
+            $self->debug( 1, "Cell info for $cell changed from @hosts to @{$ipaddrs{$cell}}" );
         }
     }
+
     # check new cells
-    foreach $cell (keys(%ipaddrs)) {
+    foreach $cell ( keys(%ipaddrs) ) {
```
stdweird: `sort keys` to make it reproducable

stdweird:
@ajf8 we're at the minor remarks :+1: :smile:

ncm-afsclt/src/main/perl/afsclt.pm
ajf8: Unfortunately we are still using CAF 13.1.0 on our older builds, so FileReader is not available.

stdweird: ok, can you add the  following:
1. an item with future milestaone that says to use FileReader in afsctl
2. add a `# TODO use FileReader`
3. do a `$fs->cancel()` after each `->new` (it's basically what filereader does)

ajf8:
Yep nearly there! It's a much better component now. Thanks again.

ncm-afsclt/src/main/perl/afsclt.pm
ajf8: I had a closer look at this, it seems totally useless. This parameter is not used anywhere.

it wouldn't work for us anyway because our afs is not mounted to /afs.

stdweird:
@ajf8 one minor remark left (i was suprised the noaction supported was there :smile: )
wrt using this with old `CAF`, i'm unaware if there are any NoAction-related bugs in previous releases, so you better test the noaction part

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -2,997 +2,281 @@
...
+use CAF::FileWriter;
+use CAF::FileEditor;
+use CAF::Process;
 
 use vars qw(@ISA $EC);
 @ISA = qw(NCM::Component);
-$EC=LC::Exception::Context->new->will_store_all;
-
+$EC  = LC::Exception::Context->new->will_store_all;
 $NCM::Component::afsclt::NoActionSupported = 1;
```
stdweird: use
```perl
our $NoActionSupported = 1;
```

(i totally missed that before, i also find it "intersting" this was set before)

stdweird:
LGTM
@ajf8 can you confirm that this code actually works. nice work!

ajf8:
Thanks :+1:  Yeah, I have been testing it. In NoAction mode and not, checking the files come out the same, and editing/deleting them to make sure updates are tested.

PR #652, opened 2015-12-16
ncm-network: Fix ovs_extra network script output

### LINKED ISSUE #651: ncm-network: ovs_extra is not set correctly
If you set ovs_extra conf from quattor the component encloses the text into single quotes instead of double quotes, for example if you set in quattor:
> "ovs_extra", 'set Interface $DEVICE type=internal',

you get from the network conf file:
> OVS_EXTRA='set Interface $DEVICE type=internal'

and the variable is not expanded, from OVS logs:
> ovs-vsctl: ovs|00001|vsctl|INFO|Called as ovs-vsctl -t 10 -- --may-exist add-br br100 -- set Interface $DEVICE type=internal
> ovs-vsctl: ovs|00002|vsctl|ERR|no row "$DEVICE" in table Interface

It should be:
> OVS_EXTRA="set Interface $DEVICE type=internal"

It could be fixed from network.pm just including double quotes:
> 294                     $text .= "OVS_EXTRA=\"$net{$iface}{ovs_extra}\"\n";

alvarosimon:
* Fix #651 : ovs_extra is not set correctly

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -279,7 +279,7 @@ sub Configure
                     $text .= "OVS_OPTIONS='$net{$iface}{ovs_opts}'\n";
                 }
                 if ($net{$iface}{'ovs_extra'}) {
-                    $text .= "OVS_EXTRA='$net{$iface}{ovs_extra}'\n";
+                    $text .= "OVS_EXTRA=\"$net{$iface}{ovs_extra}\"\n";
                 }
                 if ($net{$iface}{'bond_ifaces'}) {
                     $text .= "BOND_IFACES='".join(' ',@{$net{$iface}{bond_ifaces}})."'\n";


ncm-network/src/main/perl/network.pm
```diff
@@ -279,7 +279,7 @@ sub Configure
                     $text .= "OVS_OPTIONS='$net{$iface}{ovs_opts}'\n";
```
stdweird: also here? if we would do this for the whole component, we better have a TT file create this. but this will trigger a mass network restart

alvarosimon: I think that is only required by `OVS_EXTRA` the rest of OVS options do not include variables. 
https://github.com/asteven/openvswitch/blob/master/rhel/README.RHEL

stdweird: ok

stdweird:
LGTM

PR #654, opened 2015-12-17
Bump build tools to 1.47

stdweird:


stdweird:
the useraccess commit is due to another encounter of #432

ned21:
LGTM (are we merging 16.2 stuff yet?)

Issue #658, opened 2016-01-04
ncm-accounts: add the possibility to add or replace group members


jouvin:
I encountered a use case that `ncm-accounts` cannot handle currently. Our web servers run `httpd` with a non default user (defined by Apache `User` directive) and a non default group (defined by Apache `Group` directive). Unfortunately this is causing troubles with the new `cgid` used to run CGIs when using a multithreaded MPM (Apache 2.4): in this case the httpd child must access a named socket (file) located under `/var/run/httpd` which is readable only by `root` and the group `apache` (whatever Apache `Group` directive is set to). As `/var/run/httpd` is recreated at each boot, using Quattor `ncm-dirperm` to change directory group is not working well on machines (like VMs) where you don't necessarily start `ncm-cdispd`.
An easier approach would be just to add the user used to run Apache in the local `apache` group. Unfortunately, the current `ncm-accounts` only allows to replace the group member list (which should not be a problem in this particular case as the group has no explicit members): it would help to address the general case to be able to say that group members should be added to existing ones instead. Looking quickly at `ncm-accounts`, I have the feeling that this would be a minor modification. Is there an agreement on doing it?

jouvin:
In fact, looking again at `ncm-accounts`, I realize that currently this configuration module doesn't allow to define the list of group members but only the group gid. Thus I changed the title.
I am not sure why this restriction, may be this is only that we didn't implement it... The change would really be trivial: just add to `apply_profile_groups()` the processing of `$cfg->{members}` after adding this property to the schema.
Any objection?

stdweird:
what is wrong with adding the apache group to the list of groups of this user? (or is this a user created by rpm/something else?) i guess the main issue is that there would be a second way to add users to the group, possibly leading to unexpected results. maybe by adding a a restriction to the schema that only `kept_users` can be added this way?

jouvin:
Yes, in my use case, the account is in LDAP and there is no reason for it to have this group by default... I need to double check how members are added to groups. Clearly it should not break what exists today!

stdweird:
hmm, interesting. so `ncm-accounts` requires a working ldap config.
what creates the socket?
also, if httpd is not running as apache user, have you tried also changing the user in the unitfile (it will probbaly create /var/run/httpd with correct ownership)

jouvin:
I don't see any documented way of customizing the group for the `/var/run/httpd` directory. Are you aware of something?

stdweird:
[...]

PR #659, opened 2016-01-05
ncm-accounts: add the possility to define group required members

jouvin:
- Unit tests added
- Documentation updated (requiredMembers)

Contribution to #658.

This currently allows to replace explicitly defined group members on the target machines by what is defined in the configuration. I'm still willing to add the possibility to add rather than replace the existing members. But, after reviewing, it is probably worth merging before the next release (15.12).


jrha:
Please do not add anything else to 15.12, I wanted to get a final out ASAP, with the build tool PR merged I already have to build at least one more RC, I don't want it delayed any further.

jouvin:
I was not sure about the exact schedule for 15.12, no problem if it is postponed to next one.

jouvin:
Complete, ready for merging after 15.12 is released.

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -24,6 +25,7 @@ our @ISA = qw(NCM::Component);
 our $EC=LC::Exception::Context->new->will_store_all;
 
 our $NoActionSupported = 1;
+$CAF::Object::NoAction = 1;
```
stdweird: euhm, what?

jouvin:
Oooppppsss... Thanks for noticiing the mistake, obviously was intended to go into the unit test, not the component code!

jouvin:
BTW, deployed successfully at LAL with the last fix.

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -393,11 +393,23 @@ sub apply_profile_groups
       if (!exists($system->{groups}->{$group})) {
         $self->debug(2, "Scheduling addition of group $group");
         $system->{groups}->{$group} = { name => $group,
-                                        members => {},
+                                        members => {map(($_ => 1), @{$cfg->{requiredMembers}})},
```
stdweird: with use strict/warings, this should fail when there's no `requiredMembers`

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -393,11 +393,23 @@ sub apply_profile_groups
...
-      } elsif ($system->{groups}->{$group}->{gid} != $cfg->{gid}) {
-        $self->debug(2, "Changing gid of group $group to $cfg->{gid}");
-        $system->{groups}->{$group}->{gid} = $cfg->{gid};
+      } else {
+        if ($system->{groups}->{$group}->{gid} != $cfg->{gid}) {
+          $self->debug(2, "Changing gid of group $group to $cfg->{gid}");
+          $system->{groups}->{$group}->{gid} = $cfg->{gid};
+        }
+        if ( $cfg->{replaceMembers} ) {
+          $self->debug(2,"Group $group: replacing existing member list");
```
stdweird: space after comma

jouvin:
Should be ok now.

ned21:
edcf330: this commit log mentions a different component. Since it, and the previous one, are just fixup commits, perhaps the branch could be rebased and these squashed into the prior commits to keep the commit history more readable?

 d02207f: do we still want config-common and config-rpm for this component? These should just all be merged into config.pan?

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -390,14 +390,28 @@ sub apply_profile_groups
...
-      } elsif ($system->{groups}->{$group}->{gid} != $cfg->{gid}) {
-        $self->debug(2, "Changing gid of group $group to $cfg->{gid}");
-        $system->{groups}->{$group}->{gid} = $cfg->{gid};
+      } else {
+        if ($system->{groups}->{$group}->{gid} != $cfg->{gid}) {
+          $self->debug(2, "Changing gid of group $group to $cfg->{gid}");
+          $system->{groups}->{$group}->{gid} = $cfg->{gid};
+        }
+        if ( $cfg->{replaceMembers} ) {
+          $self->debug(2, "Group $group: replacing existing member list");
```
ned21: Should be verbose or info not debug?

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -390,14 +390,28 @@ sub apply_profile_groups
...
+          $system->{groups}->{$group}->{members} = {};
+        }
+        if ( @$required_members ) {
+          $self->debug(3, "Group $group: adding required members (",join(',',@$required_members),")");
```
ned21: verbose or info level message.

jouvin: Really? I don't think it is info, it could be verbose. But I'm afraid we should review what is displayed in verbose mode as it looks quite a detail compared to what is displayed at verbose level by this component (`grep verbose account.pm`).

ncm-accounts/src/main/perl/accounts.pm
jouvin: Same as the following comment... but if there is a consensus on this, I'm happy to change debug to verbose.

jouvin:
@ned21 thanks for spotting the typo in the component name in the commit log. As for merging all the `config*` templates I was about to do it but I hesitated as this config module was one used by Solaris and I was not sure if we should keep the ability to support several packager there...

ned21:
Good point. OTOH config.pan doesn't actually implement Solaris packaging, it unconditionally includes config-common and config-rpm so I think we might as well tidy this up for now and think about Solaris support in the future?

ncm-accounts/src/main/perl/accounts.pm
ned21: I like to see components summarise their effect at info level ("Changed N groups", "Added Y users") and then report individual actions at verbose level ("adding members X,Y,Z", "replacing users with ...") so I would definitely like to see these as verbose. I agree info is probably too much but too many components don't give any indication of what they have changed unless run in debug mode!

jouvin:
@ned21 I tend to agree with you that `ncm-accounts` is one of the configuration module that is no verbose enough about what has been modified, in particular accounts/groups removed from the existing password/group files. I reviewed logging done: https://github.com/quattor/configuration-modules-core/commit/39b326273d52b334a50bb79c08e99069bf1ef0d5 contains the proposed modification. The basic idea is to log every removal from the passwd/group file that is difficult to try otherwise at the info level (as it cannot be reproduced at next run, even at verbose level). It should remain a moderate level of information displayed and if nothing is changed nothing is displayed at info level.
Feedback welcome!

jouvin:
@ned21 and you are totally right about pan templates! I mixed my mind with `ncm-spma`! No a good start for the year!

jouvin:
FYI, this version has been deployed successfully on all LAL machines (EL5 to EL7).

ned21:
LGTM - thanks for the clean ups

jouvin:
Another missed issue with required members fixed (due to a unit test not covering all the possibilities, namely a required member not defined in the configuration).

ncm-accounts/src/main/perl/accounts.pm
```diff
@@ -390,14 +395,55 @@ sub apply_profile_groups
...
+      $required_members = [] unless $required_members;
+      my @initial_members;
+      if ( exists($system->{groups}->{$group}) ) {
+        if ($system->{groups}->{$group}->{gid} != $cfg->{gid}) {
+          $self->info("Changing gid of group $group to $cfg->{gid}");
+          $system->{groups}->{$group}->{gid} = $cfg->{gid};
+        }
+        @initial_members = keys(%{$system->{groups}->{$group}->{members}});
+        if ( $cfg->{replaceMembers} ) {
+          $self->verbose("Group $group: replacing existing member list");
```
stdweird: change the message a bit to `resetting existing member list, to be replaced`

ncm-accounts/src/main/pan/components/accounts/config.pan
```diff
@@ -1,12 +1,27 @@
...
+prefix '/software/components/${project.artifactId}';
+'active' ?= true;
+'dispatch' ?= true;
+'dependencies/pre' ?= list('spma');
+'version' = '${no-snapshot-version}';
+
+# Include system users and groups which shouldn't be removed
+# by default.  The machine configuration can still modify or
+# remove them manually.
+include { 'components/accounts/sysgroups' };
```
stdweird: no `{}`

ncm-accounts/src/main/perl/accounts.pm
stdweird: i would actaully add a comment here too like
```
# replaceMembers is achieved by initial reset of the members of the group
```
i'm confused a bit by the `replace`, maybe `forceMembers` is better? it's clear in the pod, but still confusing in the code

jouvin: I'll add a comment to clarify it.

jrha:
@stdweird - happy?

stdweird:
LGTM

PR #660, opened 2016-01-06
ncm-metaconfig: add service beats and logstash plugin

stdweird:
Support for [beats](https://www.elastic.co/products/beats) (and beats logstash input)

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/beats/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/beats/pan/schema.pan
@@ -1,0 +1,159 @@
+declaration template metaconfig/beats/schema;
+
+include 'pan/types';
+
+type beats_output_logstash_tls = {
+    'certificate_authorities' ? string[]
+    'certificate' ? string
+    'certificate_key' ? string
+    'insecure' ? boolean
+    'cipher_suites' ? string[]
+    'curve_types' ? string[]
+};
+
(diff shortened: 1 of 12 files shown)


ncm-metaconfig/src/main/metaconfig/beats/pan/schema.pan
```diff
@@ -0,0 +1,159 @@
+declaration template metaconfig/beats/schema;
+
+include 'pan/types';
+
```
kwaegema: documentation strings for the types in this schema?

jrha:
Please improve the description, I have provided a template :stuck_out_tongue:

stdweird:
@jrha is that enough?

jrha:
Lovely.

jrha:
@alahiff might like this one.

alahiff:
Yes, this will definitely be useful.

ncm-metaconfig/src/main/metaconfig/beats/pan/schema.pan
jrha: @stdweird could you respond?

stdweird: ok, i'll add them

stdweird:
@jrha @kwaegema is that sufficient for now?

PR #663, opened 2016-01-07
ncm-metaconfig: publish component version

jouvin:
Not having the version published makes `metaconfig` usage complicated in templates that must still support Quattor 13.1. It'd be great to have it fixed in 15.12 but I don't dare to ask for it...!!!

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/config.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/config.pan
@@ -17,3 +17,4 @@ prefix '/software/components/${project.artifactId}';
 'active' ?= true;
 'dispatch' ?= true;
 'dependencies/pre' ?= list('spma');
+'version' = '${no-snapshot-version}';


ncm-metaconfig/src/main/pan/components/metaconfig/config.pan
```diff
@@ -9,6 +9,8 @@ include 'components/${project.artifactId}/schema';
 
 bind '/software/components/${project.artifactId}' = ${project.artifactId}_component;
 
+'version' ?= '${no-snapshot-version}';
```
ned21: Why ```?=```  and not ```=``` ?

jouvin: An habit! I almost never use `=`. For me `?=` tends to be the default assignment operator. And this is defined like this in all configuration modules. In this particular case, I agree that it doesn't make sense to allow somebody to define another value and it may make sense to change for `= `. A topic for next workshop?!

jrha: Using `?=` may actually be harmful for the version. Imagine you have two templates that both `include 'components/metaconfig/config'` but for some reason are name-spaced to different releases at different points in the compilation, `pkg_repl` would update the version installed, but `/software/components/metaconfig/version` would always reflect the first version included which would be confusing and not helpful if you are trying to use the version for logic.

jrha:
Can be merged when the `?=` is switched to `=`.

jrha:
@jouvin ping

PR #667, opened 2016-01-08
ncm-nfs: fix incorrect exports hosts example in documentation 

jrha:
The schema and perl code indicate that hosts is a dict of host-option pairs, but the example in the pod shows a bogus list. Also tidy up the example code block while we are at it.

Spotted by @apdibbo.

### DIFF
--- a/ncm-nfs/src/main/perl/nfs.pod
+++ b/ncm-nfs/src/main/perl/nfs.pod
@@ -44,17 +44,21 @@ change, then the volume will be remounted.
 
 =head1 NFS
 
-"/software/components/nfs/exports" = list(
-  nlist("path","/shared/path/",
-        "hosts",dict("server*.example.org","no_root_squash"))
-);
+    "/software/components/nfs/exports" ?= list();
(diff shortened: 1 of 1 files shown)


ncm-nfs/src/main/perl/nfs.pod
```diff
@@ -44,17 +44,23 @@ change, then the volume will be remounted.
...
-);
-
-# If the SE is exporting its disk, mount it on the worker nodes.
-"/software/components/nfs/mounts" = list(
-  nlist("device","foreign.example.org:/shared/path/",
-        "mountpoint","/mnt/foreign",
-        "fstype","nfs",
-        "options","rw")
-);
+    "/software/components/nfs/exports" = list(
```
stdweird: i consider assigning list like pure expert mode (as it resets any previously defined exports elsewhere). can we use `append(` here instead? it's an example afterall...

jrha: Sure, I didn't feel like messing around with the examples too much.

jrha:
@stdweird are you happy with this now?

PR #669, opened 2016-01-08
ncm-metaconfig: support convert option for predefined conversions

### LINKED ISSUE #665: ncm-metaconfig: support TextRender conversion options
* add https://github.com/quattor/CCM/blob/master/src/main/perl/TextRender.pm#L225 support to https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/perl/metaconfig.pm#L101
* add predefined convert options to schema
* add this also to the unittest framework (so it can be unittested by the packaged metaconfig services)

stdweird:
Fixes #665

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -11,9 +11,10 @@ include 'quattor/functions/validation';
 type ${project.artifactId}_extension = extensible {};
 
 @documentation{
-    Set CCM::TextRender element options
+    Convert value of certain types (e.g. boolean to string yes/no)
+    (using the CCM::TextRender element options)
 }
-type ${project.artifactId}_textrender_element_options = {
+type ${project.artifactId}_textrender_convert = {
     @{Convert boolean to (lowercase) 'yes' and 'no'.}
     'yesno' ? boolean
     @{Convert boolean to (uppercase) 'YES' and 'NO'.}
@@ -30,7 +31,7 @@ type ${project.artifactId}_textrender_element_options = {
     # Only one boolean conversion can be true
     boolean_conversion = list('yesno', 'YESNO', 'truefalse', 'TRUEFALSE');
     found = false;
-    foreach(idx;opt;boolean_conversion) {
+    foreach (idx; name; boolean_conversion) {
         if(exists(SELF[name]) && SELF[name]) {
             if(found) {
                 error('metaconfig element can only have one boolean conversion enabled, got '+to_string(SELF));
@@ -42,7 +43,7 @@ type ${project.artifactId}_textrender_element_options = {
     # Only one string conversion can be true
     string_conversion = list('singlequote', 'doublequote');
     found = false;
-    foreach(idx;opt;string_conversion) {
+    foreach (idx; name; string_conversion) {
         if(exists(SELF[name]) && SELF[name]) {
             if(found) {
                 error('metaconfig element can only have one string conversion enabled, got '+to_string(SELF));
@@ -64,7 +65,7 @@ type ${project.artifactId}_config =  {
     'backup' ? string
     'preamble' ? string
     'contents' : ${project.artifactId}_extension
-    'element' ? ${project.artifactId}_textrender_element_options
+    'convert' ? ${project.artifactId}_textrender_convert
 } = nlist();
 
 type ${project.artifactId}_component = {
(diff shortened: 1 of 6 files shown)


stdweird:
@jouvin see the unittest how it works with e.g. tiny

jrha:
Ooh, déjà vu. Haven't I seen this change before?

jrha:
Maybe I'm thinking of some of the code from quattor/CAF#85.

stdweird:
@jrha yes, this is the code to expose the element options to ncm-metaconfig, it was previously only available to `TextRender`

jouvin:
LGTM. But I don't like `element` as the resource name. What about `format` or `formatting` instead?

stdweird:
they are not formatters either, and i would like to keep the `CCM::TextRender` option the same as this one. an element is a predefined conversion function, they actually convert the value of the element (and not the type).

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -5,21 +5,66 @@
...
-     'contents' : ${project.artifactId}_extension
+    'mode' : long = 0644
+    'owner' : string = 'root'
+    'group' : string = 'root'
+    'daemons' ? caf_service_action{}
+    'module' : string
+    'backup' ? string
+    'preamble' ? string
+    'contents' : ${project.artifactId}_extension
+    'element' ? ${project.artifactId}_textrender_element_options
```
jrha: How about annotations for all of these options?

ncm-metaconfig/src/test/resources/element.pan
```diff
@@ -0,0 +1,18 @@
...
+
+prefix "/software/components/metaconfig/services/{/foo/bar}";
+"mode" = 0644;
+"owner" = 'root';
+"group" = 'root';
+"module" = "tiny";
+"contents" = dict(
+    "string", "mystring",
+    "boolean", true,
+    );
```
jrha: stray indent

jouvin:
@stdweird I understand your point about using the same wording in `TextRender` and `metaconfig` but I am still not convinced by `element`. For me `element` tend to refer to the contents where here it is about conversion during rendering. `formatting` may not be the right word too but this is a typical formatter operation for me! What others think? I guess that changing in `TextRender` too is not necessary a big deal, isn'it? Or is it heavily used?

stdweird:
the current `element` is mainly used in CCM itself, and would also have to be renamed in the build-scripts, but other than that, i'm unaware of any usage. 
wrt to changing the name, a `format` in `CCM::TExtRender` is something that applies to the whole profile, not to a single element type. (`json` is both an element option (it casts all values to their type and converts booleans to something `JSON::XS` sees as a boolean; none of this should be considered "formatting" imho) and a format (it renders valid json string))

but i could name the option `formatting` or `convert` (and clarify in the pod the relation with the TextRender element option)

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -5,21 +5,66 @@
...
+    'TRUEFALSE' ? boolean
+    @{Convert string to doublequoted string.}
+    'doublequote' ? boolean
+    @{Convert string to singlequoted string.}
+    'singlequote' ? boolean
+} with {
+    # Only one boolean conversion can be true
+    boolean_conversion = list('yesno', 'YESNO', 'truefalse', 'TRUEFALSE');
+    found = false;
+    foreach(idx;opt;boolean_conversion) {
```
ned21: Isn't our convention for spaces after ; ? ```foreach(idx; opt; boolean_conversion) {```

ned21:
I agree that calling the attribute "element" on its own is misleading. ```element_conversions``` or ```element_options``` would be more accurate. (Are underscores conventional? Or should we use camel case?)

stdweird:
@jouvin @ned21 i renamed it to convert, also clarified the annotation and fixed a bug in the schema (the unittest didn't validate anything before)

ned21:
LGTM. You make it sound like it's not too hard to rename it in TextRender, but is it worth the effort?

jrha:
@stdweird tests failing...

stdweird:
@jrha seems like the template-library-core templates are not updated. i'll see if we start using the master repo for those.

stdweird:
test this please

stdweird:
@jrha should be fine in a bit

jrha:
:grinning:

PR #674, opened 2016-01-12
ncm-ceph: ceph uid for infernalis

kwaegema:
Infernalis wants a specific uid for the ceph user. To be backwards compatible with running cluster, by default nothing changes, but a deprecated message is added. When installing new clusters, or going to infernalis, you should set CEPH_OLD_UID to false

### DIFF
--- a/ncm-ceph/src/main/pan/components/ceph/ceph-user.pan
+++ b/ncm-ceph/src/main/pan/components/ceph/ceph-user.pan
@@ -12,7 +12,7 @@ include 'components/accounts/config';
 variable CEPH_OLD_UID ?= true; # set to false for new clusters, especially infernalis and above
 variable CEPH_USER_ID = {
     if (CEPH_OLD_UID) {
-        deprecated(0, 'ceph user should get a new uid');
+        deprecated(0, 'ceph user should get a new uid. set final CEPH_OLD_UID = false');
         111;
     } else {
         167;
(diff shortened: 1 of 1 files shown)


ncm-ceph/src/main/pan/components/ceph/ceph-user.pan
```diff
@@ -6,14 +6,25 @@
...
 unique template components/${project.artifactId}/ceph-user;
 
-include { 'components/accounts/config' };
+include 'components/accounts/config';
+
+# do not change uids of existing cluster
+variable CEPH_OLD_UID ?= true; # set to false for new clusters, especially infernalis and above
+variable CEPH_USER_ID = {
+    if (CEPH_OLD_UID) {
+        deprecated(0, 'ceph user should get a new uid');
```
stdweird: add a reference to `(set final CEPH_OLD_UID = false)` somehow in the derecation message

jrha:
@kwaegema ping?

jrha:
Also could @gvasilak take a look at this?

kwaegema:
@jrha done

jrha:
You could implement the default value based on the CEPH version number, but maybe it's best left to site admins.

kwaegema:
@jrha Indeed, we've set  variable `CEPH_OLD_UID ?= { if (CEPH_MAJOR_VERSION >= 9) { false; }};` in our site templates, because I know the update procedure and so rather just install infernalis from scratch :smile: 
But when trying to (it's a PITA..) update to the newest ceph, one should now this also involves chowning all the files to the new user, so maybe it is more clear if the user not *magically* changes uid, leaving all his files behind

PR #675, opened 2016-01-12
ncm-network: Don't warn if hwaddr is missing from bridge interfaces

### LINKED ISSUE #609: ncm-network: bridge interface missing hwaddr should be info, not warn
it's "normal" that there is no hwaddr
```
[WARN] No value found for the hwaddr of interface br100. Setting set_hwaddr to false.
[WARN] No value found for the hwaddr of interface br101. Setting set_hwaddr to false.
```

jrha:
This is a normal situation, so treat them the same as other meta interfaces.

Fixes #609.

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -200,10 +200,7 @@ sub Configure
                 } elsif ($hr_ifacename =~ m/^br/) {
                     $hr_ifacename = "bridging";
                 }
-                $self->info("$msg. As it appears to be a $hr_ifacename ",
-                            "interface, this very very probably not a problem. ",
-                            "(Please contact the developers in case you ",
-                            "think it is.)");
+                $self->info("$msg. As it appears to be a $hr_ifacename interface, this is considered normal.");
             } else {
                 $self->warn($msg);
             };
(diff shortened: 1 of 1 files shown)


ncm-network/src/main/perl/network.pm
```diff
@@ -191,12 +191,14 @@ sub Configure
...
                       "$ifacename. Setting set_hwaddr to false.";
-            if ($ifacename =~ m/^(bond|vlan|ib)/) {
+            if ($ifacename =~ m/^(bond|vlan|ib|br)/) {
                 my $hr_ifacename = $1;
                 if ($hr_ifacename eq "bond") {
                     $hr_ifacename = "bonding";
                 } elsif ($hr_ifacename eq "ib") {
                     $hr_ifacename = "IPoIB";
+                } elsif ($hr_ifacename ~= m/^br/) {
+                    $hr_ifacename = "bridging";
```
stdweird: bridge instead of bridging?

jrha: I was being consistent with "bonding"

stdweird: ok.

stdweird:
LGTM

PR #676, opened 2016-01-12
ncm-download: CAF::Process execute is a method, not a function

stdweird:


stdweird:
This PR looks odd, but we have made some changes to `CAF`, and now this `use` fails. it certainly does work in the releases (incl 15.12), but it isn't/can't be used in this way nor do we understand why it worked in the first place.

Issue #681, opened 2016-01-14
ncm-accounts must keep some more user for CentOS 7


guillaume-philippon:
Hi, 
During upgrading CentOS 7.1 to CentOS 7.2 we find some users that shoud be kept on CentOS machine that was not protected by default. We must add it as default protected user.
```pan
'kept_groups/ssh_keys' = '';
'kept_users/ssh_keys' = '';
'kept_groups/polkitd' = '';
'kept_users/polkitd' = '';
'kept_groups/libstoragemgmt' = '';
'kept_users/libstoragemgmt' = '';
'kept_groups/unbound' = '';
'kept_users/unbound' = '';
'kept_groups/chrony' = '';
'kept_users/chrony' = '';
```
Regards,

jrha:
Closed by #682.

PR #682, opened 2016-01-14
ncm-accounts: fix some useful accounts on CentOS 7.2

guillaume-philippon:
Fix issue #681.


jouvin:
For an easier maintenance of this file in the future, I'd suggest to sort the list of kept users/groups....

jouvin:
BTW, the list is not complete, see https://github.com/quattor/template-library-os/pull/74.

jouvin:
Note that if you installed a CentOS 7 box without this or the relevant fix for `ncm-accounts`, you may need to recreate all users/groups that may have been removed by `ncm-accounts` and created by those 2 RPMS (`rpm -q xxx --scripts` to see them). You may be affected by the problem if you run `ncm-accounts` with `remove_unknown=true` (recommended). Whether you are impacted or not depends on the setting of `preserved_accounts` and the actual value of `SYS_UID_MIN/MAX`  and `UID_MIN/MAX`.

jrha:
@guillaume-philippon, the correct syntax is "Fixes #681"

Issue #683, opened 2016-01-15
ncm-symlink: full rewrite needed


jouvin:
`ncm-symlink` deserves a full rewrite based on `getTree()` API and addressing all the [ncm-symlink issues](https://github.com/quattor/configuration-modules-core/issues?utf8=✓&q=is%3Aopen+is%3Aissue+symlink+in%3Atitle) currently open, in particular #288 about ensuring that a symlink definition cannot be added twice to the configuration, leading to unpredictable results.

ned21:
@jouvin Are you planning a new component so people can run them side-by-side during the transition? cf. #648 I think it would be worth considering having a single component that combines both symlink and dirperm since at present it's perfectly possible to get yourself into a situation where dirperm and symlink try to replace the same path, and the ordering of dirperm and symlink is dependent on how a site configures their depends.

jouvin:
Currently I was more thinking at rewriting symlink in such a way that it is based on a dict and still remain backward compatible with the old schema (I have not tested my ideas but I think it should be doable).
But I have also been thinking that it could be good/useful to merge dirperm and symlink.We could use the `ncm-system` approach with `ncm-chkconfig`: a new configuration modules with a clean schema able to digest the schema of the legacy configuration modules. What do you think about it?

jouvin:
One of the question regarding the schema is the need, if any, of operation ordering. I think the reason for using a list rather than a dict in both configuration module was to be able to address this need. If moving to a dict, it would be painful to invent a complex dependency description... Can we consider that processing the dict entries in the key alphabetical order (the key being the path) is enough? I tend to think that yes as I think the only important thing is to ensure you process the parents before the children...

stdweird:
for the rewrite, a few remarks/requests
- a big :+1: on merging dirperm and symlink. this is just basic filesystem operations
- any code actually making the modification should be part of CAF (see https://github.com/quattor/CAF/issues/69); or at least be able to factor out if you already want to start coding (like the current code in `ncm-postgresql`) :smile: 
- ideally, the new component can use the old schema, but there is a small difference between this component and `ncm-systemd`: ncm-systemd does not work on EL6, so ncm-chkconfig is not replaced by ncm-systemd entirely. this component should be able to actually replace it or run in parallel. so it would certainly be nice wrt migration from old ncm-symlink/ncm-dirperm, i don't consider it a requirement. (it's certainly a lot more work, including understanding what and how the old components worked in the first place)
[...]

PR #684, opened 2016-01-15
ncm-symlink: fix message for duplicated links

jouvin:
This is a quick fix for a component that deserves a full rewrite (#683) using the `getTree()` API...

### DIFF
--- a/ncm-symlink/src/main/perl/symlink.pm
+++ b/ncm-symlink/src/main/perl/symlink.pm
@@ -81,7 +81,8 @@ sub Configure($$@) {
                 my $link_name = $link_options{name}->getValue();
 
                 if ( exists($links{$link_name}) ) {
-                    $self->warn("Duplicate link $link_name defined in profile. Replacing previous definition target ".$links{$link_name}{target});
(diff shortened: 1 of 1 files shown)


ncm-symlink/src/main/perl/symlink.pm
```diff
@@ -81,7 +81,8 @@ sub Configure($$@) {
                 my $link_name = $link_options{name}->getValue();
 
                 if ( exists($links{$link_name}) ) {
-                    $self->warn("Duplicate link $link_name defined in profile. Replacing previous definition target ".$links{$link_name}{target});
+                    $self->warn("Duplicate link $link_name defined in profile. Replacing
```
stdweird: remove newline (if you think line is too long, use a `,` to split the message

jouvin: Done, sorry for the mistake.

stdweird: forgot to push?

jouvin:
Sorry, mistake again! Should be ok.

stdweird:
LGTM

PR #688, opened 2016-01-18
ncm-iptables: Rewrite, add unit tests and add support for per-rule comments

### LINKED ISSUE #576: ncm-iptables: Add support for comment option
iptables has support for adding comments to rules, this would be useful to have in environments where quattor is not the only thing that may change the configuration on the host. See below for the relevant section from the iptables man page on RHEL 6.

    comment
       Allows you to add comments (up to 256 characters) to any rule.

       --comment comment

       Example:
              iptables -A INPUT -s 192.168.0.0/16 -m comment --comment "A privatized IP block"

jrha:
Fixes #576.

This includes a large amount of clean-up and re-factoring, because the source code made myself and others sad.

Review with [white-space changes hidden](https://github.com/quattor/configuration-modules-core/pull/688/files?w=1) for sanity.

### DIFF
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -348,6 +348,8 @@ sub GetPathEntries {
     my ($content, $entry, $name, $value);
     my $entries = {};
 
+    $self->debug(5, "Entering method GetPathEntries");
+
     # Check the input parameters.
     if (!defined $path || $path   eq "") {
         $? = 1;
@@ -421,6 +423,8 @@ sub GetResource {
     my ($self, $path, $config) = @_;
     my ($entries, $table, $target, $rule, $name, $command, $key, $aux, $i);
     my $target_exists = 0;
+
+    $self->debug(5, "Entering method GetResource");
 
     $entries = $self->GetPathEntries( $path, $config );
     return $entries if $?;
@@ -538,6 +542,8 @@ sub sort_keys {
     my ($i, $m, $purge, $swap, $reg);
     my (@keys, @ord);
 
+    $self->debug(5, "Entering method sort_keys");
+
     # Check parameters.
     if ($rule !~ /^HASH/) {
         $? = 1;
@@ -598,6 +604,8 @@ sub rule_options_translate {
     my ($self, $rule) = @_;
     my $key;
 
+    $self->debug(5, "Entering method rule_options_translate");
+
     # Check parameters.
     if ($rule !~ /^HASH/) {
         $? = 1;
@@ -635,6 +643,8 @@ sub WriteFile {
     my ($table, $chain, $target, $rule, $name, $field, $line);
     my (@names);
 
+    $self->debug(5, "Entering method WriteFile");
+
     # Check input parameters.
     if (!defined $filename || $filename eq "") {
         $? = 1;
@@ -651,17 +661,22 @@ sub WriteFile {
     # write our "tag" into it. Assist some poor admin in debugging..
     print FILE "# Firewall configuration written by ncm-iptables\n";
     print FILE "# Manual modifications will be overwritten on the next NCM run.\n";
+    $self->debug(5, "Wrote header tag to file");
 
     # Write new content to file.
     if ( defined $iptables && ref($iptables) =~ /^HASH/ ) {
+        $self->debug(5, "iterating over tables");
         foreach $table (keys %iptables_totality) {
+            $self->debug(5, "processing table $table");
             next if (!defined $iptables->{$table} || $iptables->{$table} eq "" || ref($iptables->{$table}) !~ /^HASH/);
             print FILE "*$table\n";
 
             if (defined $iptables->{$table}->{preamble} && ref($iptables->{$table}->{preamble}) =~ /^HASH/ ) {
                 my $preamble = $iptables->{$table}->{preamble};
+                $self->debug(5, "table has preamble $preamble");
 
                 foreach $chain (@{$iptables_totality{$table}{chains}}) {
+                    $self->debug(5, "processing chain $chain");
                     next if (!defined $preamble->{$chain} || $preamble->{$chain} eq "");
                     my $g = $chain;
                     $g =~ tr/a-z/A-Z/;
@@ -672,16 +687,18 @@ sub WriteFile {
             }
 
             foreach $target (sort keys %{$iptables_totality{$table}{user_targets}}){
+                $self->debug(5, "defining target $target");
                 print FILE "-N $target\n";
             }
 
             foreach $target (@{$iptables_totality{$table}{targets}}) {
-
+                $self->debug(5, "processing rules for target $target");
                 next if ( ! defined $iptables->{$table}->{rules}->{$target}         );
                 next if (   ref($iptables->{$table}->{rules}->{$target}) !~ /^HASH/ );
                 next if ( ! scalar(%{$iptables->{$table}->{rules}->{$target}})      );
 
                 foreach $name (sort { $a <=> $b; } keys %{$iptables->{$table}->{rules}->{$target}}) {
+                    $self->debug(5, "processing rule $name for target $target");
                     next if ($name !~ /^\d+$/);
                     $rule = $iptables->{$table}->{rules}->{$target}->{$name};
                     $line = '';
@@ -720,6 +737,8 @@ sub cmp_rules {
     my ($self, $rule1, $rule2) = @_;
     my ($field);
     my (@fields1, @fields2);
+
+    $self->debug(5, "Entering method cmp_rules");
 
     # Check parameters.
     if (!defined $rule1 || ref($rule1) !~ /^HASH/) {
@@ -770,6 +789,8 @@ sub find_rule {
     my ($self, $rule, $hash) = @_;
     my ($name);
 
+    $self->debug(5, "Entering method find_rule");
+
     # Check parameters.
     if (!defined $rule || ref($rule) !~ /^HASH/) {
         $? = 1;
@@ -805,6 +826,8 @@ sub Configure {
     my ($self, $config) = @_;
     my $iptables;
     local $@;
+
+    $self->debug(5, "Entering method Configure");
 
     # Get global components parameters
     $iptables = $self->GetResource($path_iptables, $config);
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -54,7 +54,7 @@ sub regExp {
 }
 
 # Right order for iptables options.
-Readonly::Hash my %options_ord => (
+Readonly::Hash my %OPTION_SORT_ORDER => (
     '-N'                 => 0,
     '-A'                 => 0,
     '-D'                 => 0,
@@ -113,7 +113,7 @@ Readonly::Hash my %options_ord => (
 );
 
 # Translate resource names to iptables options.
-Readonly::Hash my %options_tra => (
+Readonly::Hash my %OPTION_MAPPINGS => (
     'new_chain'          => '-N',
     'append'             => '-A',
     'delete'             => '-D',
@@ -179,7 +179,7 @@ Readonly::Hash my %options_tra => (
 );
 
 # Preliminary test on the resource and sysconfig file options.
-Readonly::Hash my %options_arg => (
+Readonly::Hash my %OPTION_VALIDATORS => (
     '-A'                 => "", #defined as "($regexp_chains)" on a table by table basis
     '-D'                 => "",
     '-I'                 => "",
@@ -239,7 +239,7 @@ Readonly::Hash my %options_arg => (
 );
 
 # Operations to perform on the resource options when read for the first time.
-Readonly::Hash my %options_op  => (
+Readonly::Hash my %OPTION_MODIFIERS  => (
     '-A' => \&uppercase,
     '-D' => \&uppercase,
     '-I' => \&uppercase,
@@ -386,7 +386,7 @@ sub GetPathEntries {
         #this is a fix for the issue of values with the "command" key
         #no translation => iptables screwing up
         if ($name eq "command") {
-            $value = $options_tra{$value} if defined $options_tra{$value};
+            $value = $OPTION_MAPPINGS{$value} if defined $OPTION_MAPPINGS{$value};
         }
 
         $entries->{$name} = $value;
@@ -425,8 +425,8 @@ sub GetResource {
 
         #define the regular expressions for -N, -A etc based on the specific targets for table
         my $tmp = &uppercase($self, &regExp(@{$iptables_totality{$table}{chains}}));
-        $options_arg{$_} = $tmp foreach (@{$iptables_totality{$table}{commands}});
-        $options_arg{'-j'} = &uppercase($self, &regExp(@{$iptables_totality{$table}{targets}}));
+        $OPTION_VALIDATORS{$_} = $tmp foreach (@{$iptables_totality{$table}{commands}});
+        $OPTION_VALIDATORS{'-j'} = &uppercase($self, &regExp(@{$iptables_totality{$table}{targets}}));
 
         $entries->{$table} = &GetPathEntries("$path/$table", $config);
         next if $?;
@@ -455,7 +455,7 @@ sub GetResource {
 
             if (defined $rule->{-j}) {
                 #check if exists
-                if (&uppercase($self, $rule->{-j}) !~ /$options_arg{'-j'}/) {
+                if (&uppercase($self, $rule->{-j}) !~ /$OPTION_VALIDATORS{'-j'}/) {
                     $iptables_totality{$table}{user_targets}{&uppercase($self, $rule->{-j})} = 1;
                 }
             }
@@ -468,9 +468,9 @@ sub GetResource {
             my $val = &regExp(@{$iptables_totality{$table}{commands}});
 
             foreach $key (keys %{$rule}) {
-                if (defined $options_op{$key} && $options_op{$key} ne "") {
+                if (defined $OPTION_MODIFIERS{$key} && $OPTION_MODIFIERS{$key} ne "") {
                     my $opresult;
-                    $opresult = &{$options_op{$key}}($self, $rule->{$key});
+                    $opresult = &{$OPTION_MODIFIERS{$key}}($self, $rule->{$key});
                     if (!$opresult) {
                         $self->warn("failed to convert $key : ".$rule->{$key}." - IGNORING THIS RULE");
                         next RULE;
@@ -480,8 +480,8 @@ sub GetResource {
                     }
                 }
 
-                if (defined $options_arg{$key} && $options_arg{$key} ne "") {
-                    $aux = $options_arg{$key};
+                if (defined $OPTION_VALIDATORS{$key} && $OPTION_VALIDATORS{$key} ne "") {
+                    $aux = $OPTION_VALIDATORS{$key};
                     if ($rule->{$key} !~ /^$aux$/ && $key =~ /^$val$/) {
                         my $skip = 0;
                         foreach (@{$iptables_totality{$table}{targets}}) {
@@ -525,7 +525,7 @@ sub GetResource {
 #   OUTPUT: @keys     - list of keys in the right order,
 #           $?        - 0 keys sorted,
 #                     - 1 error.
-#      USE: %options_ord
+#      USE: %OPTION_SORT_ORDER
 #   ASSUME: If rule is not empty then is well formed.
 sub sort_keys {
     my ($self,$rule) = @_;
@@ -555,16 +555,16 @@ sub sort_keys {
     while ($swap) {
         for($m=0, $swap=0; $m<$#keys; $m++) {
             for($i=$m+1; $i<=$#keys; $i++) {
-                $self->error("$keys[$i] is not a valid option\n") if ! exists $options_ord{$keys[$i]};
-                $self->error("$keys[$m] is not a valid option\n") if ! exists $options_ord{$keys[$m]};
+                $self->error("$keys[$i] is not a valid option\n") if ! exists $OPTION_SORT_ORDER{$keys[$i]};
+                $self->error("$keys[$m] is not a valid option\n") if ! exists $OPTION_SORT_ORDER{$keys[$m]};
 
                 #next
-                if (!exists $options_ord{$keys[$i]} || !exists $options_ord{$keys[$m]}) {
+                if (!exists $OPTION_SORT_ORDER{$keys[$i]} || !exists $OPTION_SORT_ORDER{$keys[$m]}) {
                     $? = 1;
                     $@ = "keys unsorted";
                     return @keys;
                 }
-                next if ($options_ord{$keys[$i]} >= $options_ord{$keys[$m]});
+                next if ($OPTION_SORT_ORDER{$keys[$i]} >= $OPTION_SORT_ORDER{$keys[$m]});
                 $reg = $keys[$i];
                 $keys[$i] = $keys[$m];
                 $keys[$m] = $reg;
@@ -586,7 +586,7 @@ sub sort_keys {
 #    INPUT: $rule     - pointer to an hash table describing the rule;
 #   OUTPUT: $?        - 0 options translated,
 #                     - 1 error.
-#      USE: %options_tra
+#      USE: %OPTION_MAPPINGS
 #   ASSUME: If rule is not empty then is well formed.
 sub rule_options_translate {
     my ($rule) = @_;
@@ -600,9 +600,9 @@ sub rule_options_translate {
     }
 
     foreach $key (keys %{$rule}) {
-        next if (! defined $options_tra{$key} || $options_tra{$key} eq "");
-        next if (defined $rule->{$options_tra{$key}});
-        $rule->{$options_tra{$key}} = $rule->{$key};
+        next if (! defined $OPTION_MAPPINGS{$key} || $OPTION_MAPPINGS{$key} eq "");
+        next if (defined $rule->{$OPTION_MAPPINGS{$key}});
+        $rule->{$OPTION_MAPPINGS{$key}} = $rule->{$key};
         delete $rule->{$key};
     }
 
@@ -681,7 +681,7 @@ sub WriteFile {
                     $line = '';
                     foreach $field (&sort_keys($self,$rule)) {
                         $line .= ($line) ? " $field" : $field;
-                        $line .= " $rule->{$field}" if $options_arg{$field};
+                        $line .= " $rule->{$field}" if $OPTION_VALIDATORS{$field};
                     }
                     print FILE "$line\n" if $line and $line !~ /^-N/;
                 }
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -48,7 +48,7 @@ my %iptables_totality => {
 };
 
 sub regExp {
-    my $reg = "@_";
+    my ($self, $reg) = @_;
     $reg =~ s/\s/\|/g;
     return $reg;
 }
@@ -344,7 +344,7 @@ sub uppercase {
 #                     - 5 $path doesn't exist as a resource path,
 #                     - 6 $path has no entries.
 sub GetPathEntries {
-    my ($path, $config) = @_;
+    my ($self, $path, $config) = @_;
     my ($content, $entry, $name, $value);
     my $entries = {};
 
@@ -422,35 +422,35 @@ sub GetResource {
     my ($entries, $table, $target, $rule, $name, $command, $key, $aux, $i);
     my $target_exists = 0;
 
-    $entries = &GetPathEntries( $path, $config );
+    $entries = $self->GetPathEntries( $path, $config );
     return $entries if $?;
 
     foreach $table (keys %iptables_totality) {
         next if (!defined $entries->{$table});
 
         #define the regular expressions for -N, -A etc based on the specific targets for table
-        my $tmp = &uppercase($self, &regExp(@{$iptables_totality{$table}{chains}}));
+        my $tmp = $self->uppercase($self->regExp(@{$iptables_totality{$table}{chains}}));
         $OPTION_VALIDATORS{$_} = $tmp foreach (@{$iptables_totality{$table}{commands}});
-        $OPTION_VALIDATORS{'-j'} = &uppercase($self, &regExp(@{$iptables_totality{$table}{targets}}));
+        $OPTION_VALIDATORS{'-j'} = $self->uppercase($self->regExp(@{$iptables_totality{$table}{targets}}));
 
-        $entries->{$table} = &GetPathEntries("$path/$table", $config);
+        $entries->{$table} = $self->GetPathEntries("$path/$table", $config);
         next if $?;
 
-        $entries->{$table}->{preamble} = &GetPathEntries("$path/$table/preamble", $config);
+        $entries->{$table}->{preamble} = $self->GetPathEntries("$path/$table/preamble", $config);
 
         my $cnt = {};
         foreach $target (@{$iptables_totality{$table}{targets}}) {
             $cnt->{$target} = 0;
         }
 
-        $entries->{$table}->{rules} = &GetPathEntries("$path/$table/rules", $config);
+        $entries->{$table}->{rules} = $self->GetPathEntries("$path/$table/rules", $config);
         next if $?;
 
         RULE: foreach $name (sort { $a <=> $b } keys %{$entries->{$table}->{rules}}) {
             next if ($name !~ /^\d+$/);
-            $rule = &GetPathEntries( "$path/$table/rules/$name", $config );
+            $rule = $self->GetPathEntries( "$path/$table/rules/$name", $config );
             return if $?;
-            &rule_options_translate($rule);
+            $self->rule_options_translate($rule);
 
             if (!defined $rule->{chain}) {
                 $? = 7;
@@ -460,8 +460,8 @@ sub GetResource {
 
             if (defined $rule->{-j}) {
                 #check if exists
-                if (&uppercase($self, $rule->{-j}) !~ /$OPTION_VALIDATORS{'-j'}/) {
-                    $iptables_totality{$table}{user_targets}{&uppercase($self, $rule->{-j})} = 1;
+                if ($self->uppercase($rule->{-j}) !~ /$OPTION_VALIDATORS{'-j'}/) {
+                    $iptables_totality{$table}{user_targets}{$self->uppercase($rule->{-j})} = 1;
                 }
             }
 
@@ -470,12 +470,13 @@ sub GetResource {
             delete $rule->{command};
             delete $rule->{chain};
 
-            my $val = &regExp(@{$iptables_totality{$table}{commands}});
+            my $val = $self->regExp(@{$iptables_totality{$table}{commands}});
 
             foreach $key (keys %{$rule}) {
                 if (defined $OPTION_MODIFIERS{$key} && $OPTION_MODIFIERS{$key} ne "") {
                     my $opresult;
-                    $opresult = &{$OPTION_MODIFIERS{$key}}($self, $rule->{$key});
+                    my $modifier = $OPTION_MODIFIERS{$key};
+                    $opresult = $self->$modifier($rule->{$key});
                     if (!$opresult) {
                         $self->warn("failed to convert $key : ".$rule->{$key}." - IGNORING THIS RULE");
                         next RULE;
@@ -510,7 +511,7 @@ sub GetResource {
             }
 
             if ( defined $cnt->{$target} ) {
-                next if (!&find_rule($rule,$entries->{$table}->{rules}->{$target}->{$cnt->{$target}}));
+                next if (!$self->find_rule($rule,$entries->{$table}->{rules}->{$target}->{$cnt->{$target}}));
                 $entries->{$table}->{rules}->{$target}->{$cnt->{$target}} = $rule;
                 $cnt->{$target}++;
             }
@@ -533,7 +534,7 @@ sub GetResource {
 #      USE: %OPTION_SORT_ORDER
 #   ASSUME: If rule is not empty then is well formed.
 sub sort_keys {
-    my ($self,$rule) = @_;
+    my ($self, $rule) = @_;
     my ($i, $m, $purge, $swap, $reg);
     my (@keys, @ord);
 
@@ -594,7 +595,7 @@ sub sort_keys {
 #      USE: %OPTION_MAPPINGS
 #   ASSUME: If rule is not empty then is well formed.
 sub rule_options_translate {
-    my ($rule) = @_;
+    my ($self, $rule) = @_;
     my $key;
 
     # Check parameters.
@@ -684,7 +685,7 @@ sub WriteFile {
                     next if ($name !~ /^\d+$/);
                     $rule = $iptables->{$table}->{rules}->{$target}->{$name};
                     $line = '';
-                    foreach $field (&sort_keys($self,$rule)) {
+                    foreach $field ($self->sort_keys($rule)) {
                         $line .= ($line) ? " $field" : $field;
                         $line .= " $rule->{$field}" if $OPTION_VALIDATORS{$field};
                     }
@@ -716,7 +717,7 @@ sub WriteFile {
 #                     - 1 one, or the two rules, is empty or is not an
 #                         hash tables, or the rules are different.
 sub cmp_rules {
-    my ($rule1, $rule2) = @_;
+    my ($self, $rule1, $rule2) = @_;
     my ($field);
     my (@fields1, @fields2);
 
@@ -766,7 +767,7 @@ sub cmp_rules {
 #   OUTPUT: $?        - 0 the rules was found,
 #                     - 1 the rule was not found.
 sub find_rule {
-    my ($rule, $hash) = @_;
+    my ($self, $rule, $hash) = @_;
     my ($name);
 
     # Check parameters.
@@ -789,7 +790,7 @@ sub find_rule {
         next if ($name !~ /^\d+$/);
         next if (ref($hash->{$name}) !~ /^HASH/);
 
-        if (!&cmp_rules($rule, $hash->{$name}) && ! $?) {
+        if (!$self->cmp_rules($rule, $hash->{$name}) && ! $?) {
             $@ = "rule found on the list";
             return $?
         }
@@ -816,7 +817,7 @@ sub Configure {
     };
     $self->error("failed to create temporary iptables file: $@") and return 1 if $@;
 
-    &WriteFile($self, $iptc_temp, $iptables);
+    $self->WriteFile($iptc_temp, $iptables);
     if ($? > 0) {
         # bad - bail out
         $self->error($@);
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -47,7 +47,7 @@ my %iptables_totality => {
     },
 };
 
-sub regExp () {
+sub regExp {
     my $reg = "@_";
     $reg =~ s/\s/\|/g;
     return $reg;
@@ -266,7 +266,7 @@ sub quote_string {
 # SYNOPSYS: $ip dns2ip ( $name )
 #    INPUT: $name     - host name to translate;
 #   OUTPUT: $ip       - ip address.
-sub dns2ip ( $ ) {
+sub dns2ip {
     my ($self, $name) = @_;
     my ($hostname, $alias, $addrtype, $length, $addr);
     my @addr;
@@ -795,7 +795,7 @@ sub find_rule {
     return $?;
 }
 
-sub Configure($$@) {
+sub Configure {
     my ($self, $config) = @_;
     my $iptables;
     local $@;
--- a/ncm-iptables/src/main/perl/iptables.pm
+++ b/ncm-iptables/src/main/perl/iptables.pm
@@ -4,9 +4,6 @@
 #
 # iptables - Setup the IPTABLES firewall.
 #
-# Managed files:
-#   /etc/sysconfig/iptables
-
 package NCM::Component::iptables;
 
 use strict;
@@ -28,6 +25,8 @@ $NCM::Component::iptables::NoActionSupported = 1;
 
 # Global variables
 Readonly::Scalar my $path_iptables => '/software/components/iptables';
+
+Readonly::Scalar my $CONFIG_IPTABLES => '/etc/sysconfig/iptables';
 
 # hash of tables, chains & targets
 my %iptables_totality => {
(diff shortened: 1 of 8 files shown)


jrha:
Do not merge yet, it's missing a commit...

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -31,21 +30,20 @@ $NCM::Component::iptables::NoActionSupported = 1;
 # Global variables
 ##########################################################################
 
-my $path_iptables     = '/software/components/iptables';
-
+my $path_iptables = '/software/components/iptables';
 
 my %iptables_totality = (); # hash of tables, chains & targets
```
stdweird: use readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -57,194 +55,201 @@ sub regExp () {
...
-                    '--log-level'         => 18,
-                    '--log-tcp-options'   => 19,
-                    '--log-ip-options'    => 20,
-                    '--log-uid'           => 20,
-                    '--limit-burst'       => 21,
-                    '--to-destination'    => 22,
-                    '--to-source'         => 22,
-                    '--to-ports'          => 22
-    );
+my %options_ord = (
```
stdweird: readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -57,194 +55,201 @@ sub regExp () {
...
-		    'remove'             => '--remove',
-		    'rdest'              => '--rdest',
-		    'rsource'            => '--rsource',
-		    'rttl'               => '--rttl',
-		    'update'             => '--update',
-		    'seconds'            => '--seconds',
-		    'hitcount'           => '--hitcount',
-		    'name'               => '--name',
-    );
+my %options_tra = (
```
stdweird: readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -57,194 +55,201 @@ sub regExp () {
...
-		    '--update'          => '',
-		    '--remove'          => '',
-		    '--rdest'           => '',
-		    '--rsource'         => '',
-		    '--rttl'            => '',
-		    '--seconds'         => '\d+',
-		    '--hitcount'        => '\d+',
-		    '--name'            => '\S+',
-    );
+my %options_arg = (
```
stdweird: readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -57,194 +55,201 @@ sub regExp () {
...
-my %options_op  = ( '-A'              => \&upercase,
-                    '-D'              => \&upercase,
-                    '-I'              => \&upercase,
-                    '-R'              => \&upercase,
-                    '-N'              => \&upercase,
-                    '-j'              => \&upercase,
-                    '-s'              => \&dns2ip,
-                    '-d'              => \&dns2ip
-    );
+my %options_op  = (
```
stdweird: readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -343,47 +345,47 @@ sub GetPathEntries {
     my $entries = {};
 
     # Check the input parameters.
-    if ( ! defined $path   || $path   eq "" ) {
-	$? = 1;
-	$@ = "resource path to query empty";
-	return $entries;
+    if (!defined $path || $path   eq "") {
+        $? = 1;
+        $@ = "resource path to query empty";
```
stdweird: wherever this used / catched, please make sure it's locallized, see https://github.com/quattor/CAF/pull/128

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -343,47 +345,47 @@ sub GetPathEntries {
     my $entries = {};
 
     # Check the input parameters.
-    if ( ! defined $path   || $path   eq "" ) {
-	$? = 1;
-	$@ = "resource path to query empty";
-	return $entries;
+    if (!defined $path || $path   eq "") {
```
stdweird: where does `$path` come from?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -343,47 +345,47 @@ sub GetPathEntries {
...
+        $? = 1;
+        $@ = "resource path to query empty";
+        return $entries;
     }
-    if ( ! defined $config || $config eq "" ) {
-	$? = 3;
-	$@ = "missing configuration object when getting resource path \"$path\"";
-	return $entries;
+    if (!defined $config || $config eq "") {
```
stdweird: `$config eq ""` ?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -343,47 +345,47 @@ sub GetPathEntries {
...
-	if ($name eq "command"){
-	    $value = $options_tra{$value} if defined $options_tra{$value};
-	}
-
-	$entries->{$name} = $value;
+    while ($content->hasNextElement()) {
+        $entry  = $content->getNextElement();
+        $name   = $entry->getName();
+        $value  = $entry->getValue();
+        $value  = '' if ( $entry->getType() == 33 );
```
stdweird: `33`?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -343,47 +345,47 @@ sub GetPathEntries {
...
-	$value  =~ s/[\s\t]+/ /g;
-
-	#this is a fix for the issue of values with the "command" key
-	#no translation => iptables screwing up
-	if ($name eq "command"){
-	    $value = $options_tra{$value} if defined $options_tra{$value};
-	}
-
-	$entries->{$name} = $value;
+    while ($content->hasNextElement()) {
```
stdweird: use `getTree`

stdweird:
@jrha i probably shouldn't have started reviewing, but i certainly stopped :sob:

jrha:
Yes please be quiet, I'm not going to fix all of it in one PR.

ncm-iptables/src/main/perl/iptables.pm
jrha: I have no idea, I'm trying to reverse engineer this component.

ncm-iptables/src/main/perl/iptables.pm
jrha: `@_`

ncm-iptables/src/main/perl/iptables.pm
jrha: I've added `local $@` to each function that uses it, is that sufficient?

ncm-iptables/src/main/perl/iptables.pm
jrha: Oh that I could, there is some dark juju around L444...

jrha:
@stdweird any better? I'm not trying to fix all the issues with ncm-iptables in this PR, I'll open a new one for porting to a more modern code-base, preferably with unit tests (as it has none).

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
-my %iptables_totality = (); # hash of tables, chains & targets
-
-@{$iptables_totality{filter}{chains}}   = ('input','output','forward');
-@{$iptables_totality{filter}{targets}}  = ('ordered','log','accept','reject', 'return', 'classify', 'ulog', 'drop');
-@{$iptables_totality{filter}{commands}} = ('-A', '-D', '-I', '-R', '-N');
-
-@{$iptables_totality{nat}{chains}}   = ('prerouting', 'output', 'postrouting');
-@{$iptables_totality{nat}{targets}}  = ('dnat', 'snat', 'masquerade', 'redirect', 'log');
-@{$iptables_totality{nat}{commands}} = ('-A', '-D', '-I', '-R', '-N');
+Readonly::Scalar my $path_iptables => '/software/components/iptables';
```
stdweird: uppercase?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
+
+# hash of tables, chains & targets
+my %iptables_totality => {
```
stdweird: readonly?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
-                    '--log-level'         => 18,
-                    '--log-tcp-options'   => 19,
-                    '--log-ip-options'    => 20,
-                    '--log-uid'           => 20,
-                    '--limit-burst'       => 21,
-                    '--to-destination'    => 22,
-                    '--to-source'         => 22,
-                    '--to-ports'          => 22
-    );
+Readonly::Hash my %options_ord => (
```
stdweird: uppercase

stdweird: a comment what is being mapped here?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
-	return '';
+    if (scalar(@addr) != 4) {
+        $self->debug(2, "dns2ip-BAD: weird address format/length?");
+        return '';
     };
-    
-    $name =  "@addr";
+
+    $name = "@addr";
     $name =~ s/\s/\./g;
```
stdweird: you mean `join('.', @addr)`?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
     $self->debug(2, "dns2ip-OK: resolved $name");
 
     if ($isneg) {
-	return "! ".$name;
+        return "! ".$name;
```
stdweird: use `"! $name"`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -21,308 +19,318 @@ use File::Copy;
...
-##########################################################################
-sub upercase {
+#   OUTPUT: $text     - text in uppercase.
+sub uppercase {
     my ($self, $text) = @_;
-
-    return '' if ( ! defined $text );
-
+    return '' if (!defined $text);
     $text =~ tr/a-z/A-Z/;
```
stdweird: whole method is
```perl
return defined $text ? uc($text) : '';
```

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -336,54 +344,54 @@ sub upercase {
 #                     - 4 $config is not an object,
 #                     - 5 $path doesn't exist as a resource path,
 #                     - 6 $path has no entries.
-##########################################################################
 sub GetPathEntries {
     my ($path, $config) = @_;
     my ($content, $entry, $name, $value);
     my $entries = {};
+    local $@;
```
stdweird: you need to localize it where method is called, doing this probably breaks the functionality

ncm-iptables/src/main/perl/iptables.pm
stdweird: sorry, i didn't even look at the first argument, i assumed `$self`.

ncm-iptables/src/main/perl/iptables.pm
stdweird: i'll try to find the `eval{}` that uses it.

ncm-iptables/src/main/perl/iptables.pm
stdweird: it's a boolean (CCM::Element has the constants, but they are not exported)
```perl
Readonly my $CCM_ELEMENT_BOOLEAN => 33;
```

stdweird: but this code is unaware of the value of the boolean? true or false don't matter?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -336,54 +344,54 @@ sub upercase {
...
-	    $value = $options_tra{$value} if defined $options_tra{$value};
-	}
-
-	$entries->{$name} = $value;
+    while ($content->hasNextElement()) {
+        $entry  = $content->getNextElement();
+        $name   = $entry->getName();
+        $value  = $entry->getValue();
+        $value  = '' if ( $entry->getType() == 33 );
+        $value  =~ s/^[\s\t]*|[\s\t]*$//g;
```
stdweird: `\s` should include `\t`; but what is being matched here?

stdweird: nvm, this is `$str =~ s/^\s*|\s*$//g`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -336,54 +344,54 @@ sub upercase {
...
+        $value  =~ s/[\s\t]+/ /g;
```
stdweird: `$value  =~ s/\s+/ /g;`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +414,105 @@ sub GetPathEntries {
 #                     - 6 $path has no entries,
 #                     - 7 resource have at least one bad rule.
 #   ASSUME: The rules content is valid.
-##########################################################################
 sub GetResource {
     my ($self, $path, $config) = @_;
     my ($entries, $table, $target, $rule, $name, $command, $key, $aux, $i);
     my $target_exists = 0;
+    local $@;
```
stdweird: localizing in `Configure` looks sufficient

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +414,105 @@ sub GetPathEntries {
...
-	next if ( ! &find_rule($rule,$entries->{$table}->{rules}->{$target}->{$cnt->{$target}}) );
-	$entries->{$table}->{rules}->{$target}->{$cnt->{$target}} = $rule;
-	$cnt->{$target}++;
-    }
-
-}
+        next if (!defined $entries->{$table});
+
+        #define the regular expressions for -N, -A etc based on the specific targets for table
+        my $tmp = &uppercase($self, &regExp(@{$iptables_totality{$table}{chains}}));
```
stdweird: `$self->uppercase`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,93 +629,78 @@ sub rule_options_translate {
...
+        $@ = 'filename to write missing';
+        return $?;
     }
 
     # Open the file.
-    unless ( open(FILE, ">$filename") ) {
-	$? = 6;
-	$@ = "cannot open $filename";
-	return $?;
+    unless (open(FILE, ">$filename")) {
```
stdweird: `CAF::FileWriter`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -739,22 +716,22 @@ sub WriteFile {
 #   OUTPUT: $?        - 0 the rules are equal,
 #                     - 1 one, or the two rules, is empty or is not an
 #                         hash tables, or the rules are different.
-##########################################################################
 sub cmp_rules {
     my ($rule1, $rule2) = @_;
     my ($field);
     my (@fields1, @fields2);
+    local $@;
```
stdweird: only in `Configure`

stdweird:
@jrha i understand, this is some pretty strange code
* `localize $@` in Configure is enough, certainly not in all methods/functions
* if you can, maybe change the following
 * use `CAF::FileWriter`
 * use methods everywhere (so call `$self->somefunction` and define in each function `my ($self ,...) @_;`

very very strange code

ncm-iptables/src/main/perl/iptables.pm
ned21: Replace with $self->prefix would be better.

ncm-iptables/src/main/perl/iptables.pm
ned21: ugh - this component claims to support NoAction too. This is a nasty bug :(

ncm-iptables/src/main/perl/iptables.pm
jrha: :sob: I don't know...

jrha:
I was planning to switch to `CAF::FileWriter` in a seperate PR, this one was just to get `--comment` working which we need sooner rather than later.

ncm-iptables/src/main/perl/iptables.pm
jrha: I mean nothing. But yes, that would be saner.

ncm-iptables/src/main/perl/iptables.pm
jrha: Seems that way, I think this comes from options like:
```pan
    "syn"                ? boolean
    "nosyn"              ? boolean
```
Which should really be one true/false option as far as I am concerned.

I hate this component.

stdweird: use shorewall? :smile:

jrha: Tempting.

jrha:
My last commit removes the usage of `File::Temp` as recommended by our [Coding Style](https://trac.lal.in2p3.fr/Quattor/wiki/Development/Code/CodingStyle) I am unsure of the wisdom of this though, so some advice would be welcome. Given the importance of firewall rules it would ideally be good if the component tested the configuration and rolled back to a working one if back stuff happened, it would be even better of course if it didn't break things in the first place.

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -530,54 +526,52 @@ RULE:	foreach $name (sort { $a <=> $b } keys %{$entries->{$table}->{rules}}) {
...
+        $? = 1;
+        $@ = "bad rule";
+        return ();
+    }
+
+    my @keys = keys %{$rule};
+
+    my $purge = 1;
+    WHILE: while ($purge) {
+        FOR: for (my $i=0, $purge=0; $i<=$#keys; $i++) {
```
jrha: This confuses me, it seems like the loop test should be `$purge==0`, does the code as it stands work as intended?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -530,54 +526,52 @@ RULE:	foreach $name (sort { $a <=> $b } keys %{$entries->{$table}->{rules}}) {
...
+            next if ($keys[$i] !~ /^(err|checked)$/);
+            splice(@keys,$i,1);
+            $purge = 1;
+            last FOR;
+        }
+    }
+
+    my $swap = 1;
+    while ($swap) {
+        for (my $m=0, $swap=0; $m<$#keys; $m++) {
```
jrha: Same comment as above.

ncm-iptables/src/main/perl/iptables.pm
stdweird: if the code is intended to confuse people, i'd say works perfectly. i have a vague idea what is supposed to do, but i'm not sure.

```perl
my @keys = grep {$_ !~ m/^(err|checked)$/} sort keys %{$rule};
```

ned21:
@jrha - I don't quite follow your comment about 64f7c4. At present it looks like you have reimplemented the old behaviour exactly, there is no functional difference except for the elimination of a tempfile. That makes sense. However your question is whether it would be better to store a backup of the old configuration and roll back if the reload fails? Or is there a way to check the new configuration file for errors before it is loaded? (cf. ncm-ssh). If the former, then CAF:Writer already creates a .bak file (securely?) by default so rollback should be easy to add?

jrha:
@ned21 thanks, Yes that was my question and I you've answered it :grinning:

jrha:
@stdweird any more comments? :innocent:

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
-
-##########################################################################
-sub regExp () {
-    my $reg = "@_";
+Readonly::Scalar my $path_iptables => '/software/components/iptables';
+
+Readonly::Scalar my $CONFIG_IPTABLES => '/etc/sysconfig/iptables';
+
+# hash of tables, chains & targets
+my %iptables_totality = (
```
stdweird: `Readonly::Hash my %IPTABLES_TOTALITY => { ... }`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+        commands => ['-A', '-D', '-I', '-R', '-N'],
+    },
+    mangle => {
+        chains => ['prerouting', 'input', 'output', 'forward', 'postrouting'],
+        targets => ['tos', 'ttl', 'mark', 'netmap', 'classify', 'dscp', 'ecn', 'mark', 'same', 'tcpmss'],
+        commands => ['-A', '-D', '-I', '-R', '-N'],
+    },
+);
+
+sub regExp {
```
stdweird: add comment what this is supposed to do

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+    my ($self, $text) = @_;
+    $text =~ s/\s+/ /g;
+    return defined $text ? $text : '';
+}
+
+sub quote_string {
+    my ($self, $text) = @_;
+    $text = $self->trim_whitespace($text);
+    if ($text =~ /\s/) {
+        # Only quote if string still contains whitespace
```
stdweird: move comment outside function

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+
+sub trim_whitespace {
+    # Trim leading and trailing whitespace
+    my ($self, $text) = @_;
+    $text =~ s/^\s+|\s+$//g;
+    return defined $text ? $text : '';
+}
+
+sub collapse_whitespace {
+    # Collapse multiple whitespace characters into a single space
```
stdweird: outside method

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+    '-R' => 'uppercase',
+    '-N' => 'uppercase',
+    '-j' => 'uppercase',
+    '-s' => 'dns2ip',
+    '-d' => 'dns2ip',
+    '--comment' => 'quote_string',
+);
+
+sub trim_whitespace {
+    # Trim leading and trailing whitespace
```
stdweird: outside method

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+    '-I' => 'uppercase',
+    '-R' => 'uppercase',
+    '-N' => 'uppercase',
+    '-j' => 'uppercase',
+    '-s' => 'dns2ip',
+    '-d' => 'dns2ip',
+    '--comment' => 'quote_string',
+);
+
+sub trim_whitespace {
```
stdweird: we have some convention to put the `{` on the next line

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+    # Trim leading and trailing whitespace
+    my ($self, $text) = @_;
+    $text =~ s/^\s+|\s+$//g;
+    return defined $text ? $text : '';
```
stdweird: if `$text` is not defined, line above will produce a warning.
```perl
sub trim_whitespace
{
    my ($self, $text) = @_;
    $text = '' if ! defined $text;
    $text =~ s/^\s+|\s+$//g;
    return $text;
}
```
perl 5.10 has a defined-or operator, but not in el5...

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+}
+
+sub collapse_whitespace {
+    # Collapse multiple whitespace characters into a single space
+    my ($self, $text) = @_;
+    $text =~ s/\s+/ /g;
+    return defined $text ? $text : '';
```
stdweird: same undefined warning remark

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+sub dns2ip {
     my ($self, $name) = @_;
-    my ($hostname, $alias, $addrtype, $length, $addr);
-    my @addr;
-    my $isneg = 0;
 
-    if ( ! defined $name || $name eq "" ) {
-	$self->debug(2, "dns2ip-BAD: empty name");
-	return '';
+    if (!$name) {
```
stdweird: `$name` better not be `0`.

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
-    if ( $name =~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(\/\d{1,2}){0,1}$|\/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/ ) { # optional netmask or CIDR postfix
-	$self->debug(2, "dns2ip-OK: already numeric");
-	if ($isneg) {
-	    return "! ".$name;
-	} else {
-	    return $name;
-	}
+    if ($name =~ /^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}(\/\d{1,2}){0,1}$|\/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$/) { # optional netmask or CIDR postfix
+        $self->debug(2, "dns2ip-OK: already numeric");
+        if ($isneg) {
```
stdweird: `return ($isneg ? '! ' : '').$name`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
+        return '';
     };
-    
-    $name =  "@addr";
-    $name =~ s/\s/\./g;
+
+    $name = join('.', @addr);
     $self->debug(2, "dns2ip-OK: resolved $name");
 
     if ($isneg) {
```
stdweird: `return ($isned ? '! ' : '').$name;`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +426,105 @@ sub GetPathEntries {
...
-    if ( defined $cnt->{$target} ) {
-	next if ( ! &find_rule($rule,$entries->{$table}->{rules}->{$target}->{$cnt->{$target}}) );
-	$entries->{$table}->{rules}->{$target}->{$cnt->{$target}} = $rule;
-	$cnt->{$target}++;
-    }
+    my $entries = $self->GetPathEntries( $path, $config );
+    return $entries if $?;
 
-}
+    foreach my $table (keys %iptables_totality) {
```
stdweird: `sort keys`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -18,311 +13,332 @@ use vars qw(@ISA $EC);
...
-		    '--update'          => '',
-		    '--remove'          => '',
-		    '--rdest'           => '',
-		    '--rsource'         => '',
-		    '--rttl'            => '',
-		    '--seconds'         => '\d+',
-		    '--hitcount'        => '\d+',
-		    '--name'            => '\S+',
-    );
+my %OPTION_VALIDATORS = (
```
stdweird: Readonly

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +426,105 @@ sub GetPathEntries {
...
+    my $entries = $self->GetPathEntries( $path, $config );
+    return $entries if $?;
 
-}
+    foreach my $table (keys %iptables_totality) {
+        next if (!defined $entries->{$table});
+
+        #define the regular expressions for -N, -A etc based on the specific targets for table
+        my $tmp = $self->uppercase($self->regExp(@{$iptables_totality{$table}{chains}}));
+        $OPTION_VALIDATORS{$_} = $tmp foreach (@{$iptables_totality{$table}{commands}});
```
stdweird: you continously update `OPTION_VALIDATORS`, which sounds very odd.
make `OPTION_VALIDATORS` a readonly and intialise a copy at the start of each loop like
```perl
my %option_validators = %{ { %OPTION_VALIDATORS } };
```
that should work out fine, because the order of the loop nor size can be important/relevant

stdweird: hmm, you will need a deep copy.
```perl
use Storable qw(dclone);
...
my %option_validators = %{ dclone(\%OPTION_VALIDATORS) };
```

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +426,105 @@ sub GetPathEntries {
...
+                        next if $skip;
+                        push(@{$iptables_totality{$table}{targets}}, $rule->{$key});
+                        $iptables_totality{$table}{user_targets}{$rule->{$key}} = 1;
+                    }
+                }
+            }
+
+            my $target;
+            if (defined $rule->{'-j'}) {
+                $target = $rule->{'-j'};
```
stdweird: just `$target = lc($rule->{'-j'});`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +426,105 @@ sub GetPathEntries {
...
+                $target =~ tr/A-Z/a-z/;
+            }
+
+            if (defined $entries->{$table}->{ordered_rules} && $entries->{$table}->{ordered_rules} eq "yes") {
+                $target = "ordered";
+            }
+
+            if (defined $cnt->{$target}) {
+                next if (!$self->find_rule($rule,$entries->{$table}->{rules}->{$target}->{$cnt->{$target}}));
```
stdweird: space after `,`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -407,113 +426,105 @@ sub GetPathEntries {
...
+            }
+
+            if (defined $rule->{-j}) {
+                #check if exists
+                if ($self->uppercase($rule->{-j}) !~ /$OPTION_VALIDATORS{'-j'}/) {
+                    $iptables_totality{$table}{user_targets}{$self->uppercase($rule->{-j})} = 1;
+                }
+            }
+
+            $rule->{command}=$iptables_totality{$table}{commands}[0] if(!defined $rule->{command} || $rule->{command} eq "");
```
stdweird: spaces around `=`; if `$rule->{command}` cannot be `0`, use `if (! $rule->{command})`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
 #                     - 6 cannot open $filename for writing,
 #                     - 7 cannot close $filename.
 #   ASSUME: The component resource path is well formed.
-##########################################################################
 sub WriteFile {
     my ($self, $filename, $iptables) = @_;
-    my ($table, $chain, $target, $rule, $name, $field, $line);
-    my (@names);
+
+    $self->debug(5, "Entering method WriteFile");
```
stdweird: the pseudo docstring above does not match the arguments

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
-    unless ( close(FILE) ) {
-	$? = 7;
-	$@ = "cannot close $filename";
-	return $?;
-    }
-
-    $? = 0;
-    $@ = "modified $filename";
-    return $?;
+    if (defined $iptables && ref($iptables) =~ /^HASH/) {
```
stdweird: what else could it be? (and if so, log an error)

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
+        $self->debug(5, "iterating over tables");
+        foreach my $table (keys %iptables_totality) {
```
stdweird: `sort`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
 #                     - 6 cannot open $filename for writing,
 #                     - 7 cannot close $filename.
 #   ASSUME: The component resource path is well formed.
-##########################################################################
 sub WriteFile {
```
stdweird: replace with TT? :grin:

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
-    }
-
-    $? = 0;
-    $@ = "modified $filename";
-    return $?;
+    if (defined $iptables && ref($iptables) =~ /^HASH/) {
+        $self->debug(5, "iterating over tables");
+        foreach my $table (keys %iptables_totality) {
+            $self->debug(5, "processing table $table");
+            next if (!$iptables->{$table} || ref($iptables->{$table}) !~ /^HASH/);
```
stdweird: what else could it be?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
+            print $fh "*$table\n";
+
+            if (defined $iptables->{$table}->{preamble} && ref($iptables->{$table}->{preamble}) =~ /^HASH/ ) {
```
stdweird: same useless check

ncm-iptables/src/main/perl/iptables.pm
stdweird: i would do a 
```perl
if(! $iptables) {
    $self->info("No iptables passed");
    return 0;
}
```
before opening initialising the `FileWriter` to avoid the indentation
(better yet, do the cehck before the method is called)

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
+                $self->debug(5, "table has preamble $preamble");
+
+                foreach my $chain (@{$iptables_totality{$table}{chains}}) {
+                    $self->debug(5, "processing chain $chain");
+                    next if (!$preamble->{$chain});
+                    my $g = $chain;
+                    $g =~ tr/a-z/A-Z/;
+                    $preamble->{$chain} = $self->trim_whitespace($preamble->{$chain});
+                    $preamble->{$chain} = $self->collapse_whitespace($preamble->{$chain});
+                    print $fh ":$g $preamble->{$chain}\n";
```
stdweird: remove the `$g`, do a `print $fh ":",uc($chain), " ", $preamble->{$chain}, "\n";`?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
+                    my $rule = $iptables->{$table}->{rules}->{$target}->{$name};
+                    my $line = '';
+                    foreach my $field ($self->sort_keys($rule)) {
+                        $line .= ($line) ? " $field" : $field;
+                        $line .= " $rule->{$field}" if $OPTION_VALIDATORS{$field};
+                    }
+                    print $fh "$line\n" if $line and $line !~ /^-N/;
+                }
+            }
+            print $fh "$iptables->{$table}->{epilogue}\n" if (defined $iptables->{$table}->{epilogue} && $iptables->{$table}->{epilogue} ne "");
```
stdweird: if epilogue cannot be `0`, simplify the `if`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -636,101 +611,84 @@ sub rule_options_translate {
...
+                $self->debug(5, "processing rules for target $target");
+                next if (!defined $iptables->{$table}->{rules}->{$target});
+                next if (ref($iptables->{$table}->{rules}->{$target}) !~ /^HASH/);
+                next if (!scalar(%{$iptables->{$table}->{rules}->{$target}}));
+
+                foreach my $name (sort { $a <=> $b; } keys %{$iptables->{$table}->{rules}->{$target}}) {
+                    $self->debug(5, "processing rule $name for target $target");
+                    next if ($name !~ /^\d+$/);
+                    my $rule = $iptables->{$table}->{rules}->{$target}->{$name};
+                    my $line = '';
```
stdweird:
```perl
my @line;
foreach my $field ($self->sort_keys($rule)) {
    push(@line, $field);
    push(@line, $rule->{$field}) if $OPTION_VALIDATORS{$field};
};
print $fh  join(" ", @line),"\n" if @line and $line[0] !~ /^-N/;
```

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -739,43 +697,41 @@ sub WriteFile {
...
+        $@ = "second rule is not an hash table";
+        return $?;
     }
 
     $? = 1;
     $@ = "rule is not in the list";
 
-    @fields1 = keys %{$rule1};
-    @fields2 = keys %{$rule2};
+    my @fields1 = keys %{$rule1};
```
stdweird: `sort` both, use `%$rule` (no need for `{}`)

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -739,43 +697,41 @@ sub WriteFile {
...
+    my @fields2 = keys %{$rule2};
 
-    return $? if ( scalar(@fields1) <= 0 && scalar(@fields2) >  0 );
-    return $? if ( scalar(@fields1) >  0 && scalar(@fields2) <= 0 );
+    return $? if (scalar(@fields1) <= 0 && scalar(@fields2) >  0);
```
stdweird: `< 0`, really? btw, are these 2 lines `scalar(@fields1) != scalar(@fields2)`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -739,43 +697,41 @@ sub WriteFile {
...
+    return $? if (scalar(@fields1) >  0 && scalar(@fields2) <= 0);
 
-    if ( scalar(@fields1) <= 0 && scalar(@fields2) <=  0 ) {
-	$? = 0;
-	$@ = "rules are equal";
-	return $?;
+    if (scalar(@fields1) <= 0 && scalar(@fields2) <=  0) {
```
stdweird: if you know they are equal size, you can test one of them

stdweird: there's no reason for this code to exist, it's handled correctly by code below

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +783,38 @@ sub find_rule {
...
-
-##########################################################################
-
-##########################################################################
-sub Configure($$@) {
-##########################################################################
-
+sub Configure {
     my ($self, $config) = @_;
+    local $@;
```
stdweird: the old code was evil, i don not understand that ncm-ncd didn't give more errors when this was used without `local $@` (there is code above that sets `$@` on success, like, wtf.)

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +783,38 @@ sub find_rule {
...
-		}
-	    } else {
-		$self->error("command \"/sbin/service iptables condrestart\" failed:\n$ip_stdouterr");
-	    }
-	}
+    my $changes = $self->WriteFile($CONFIG_IPTABLES, $iptables);
+    $self->debug(5, "WriteFile returned $changes");
+    if ($changes) {
+        # Reload the service - file changed
+        if ($NoAction) {
```
stdweird: makes no sense anymore now that you use `CAF::FileWriter`, the file will not be modifed in `NoAction`

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +783,38 @@ sub find_rule {
...
+            $self->info("Would run \"/sbin/service iptables condrestart\"");
+        } else {
+            # allow no "dangling" file descriptors, this may be executing in a restricted targeted SELinux context
```
stdweird: what does this even mean?

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +783,38 @@ sub find_rule {
...
+            my $ip_stdouterr;
+            if (LC::Process::execute([qw(/sbin/service iptables condrestart)], "stdout" => \$ip_stdouterr, "stderr" => "stdout")) {
```
stdweird: `CAF::Process`
```perl
use CAF::Process;
...
my $proc = CAF::Process->new([qw(/sbin/service iptables condrestart)], log=> $self);
my $ip_stdouterr = $proc->output();
if ($?) {
    $self->error("command \"$proc\" failed:\n$ip_stdouterr");
} else {
    $self->info("ran \"$proc\"");
     $self->info($ip_stdouterr) if ($ip_stdouterr);
}
```

stdweird:
@jrha when viewing the diff with `?w=0` you can make comments, so my aplogies for reading the code.
just make sure that whatever you settle on as polished enough, you have tested the code on a real system.
i can keep making comments, i know it's not what you want, so feel free to address whatever you like.

ncm-iptables/src/main/perl/iptables.pm
ned21: That's not quite true. In NoAction mode, $changes will be true if the file _would_ be modified. Therefore the service reload does still need to be NoAction aware, and LC::Process is not. Using CAF::Process instead of LC::Process would take care of this automatically.

stdweird: indeed, would-be change is also reported with NoAction=1. so use `CAF::Process`

ncm-iptables/src/main/perl/iptables.pm
jrha: Yeah it's odd, but it's not me doing it :stuck_out_tongue:

ncm-iptables/src/main/perl/iptables.pm
jrha: Yeah I've thought about it, but not yet.

jrha:
Thanks, I appreciate the comments, just don't want the blame for the original code!

stdweird:
:beers: waiting in FOSEDM bar on saturday :smile:

ncm-iptables/src/main/perl/iptables.pm
jrha: I... don't know :cry:

jrha:
I won't be there unfortunately :disappointed:, I'll have beer anyway.

ncm-iptables/src/main/perl/iptables.pm
jrha: I'm not sure that's easier to read, but whatever.

ncm-iptables/src/main/perl/iptables.pm
jrha: Damn right. But this will also ensure it isn't `0` won't it?

ncm-iptables/src/main/perl/iptables.pm
jrha: Ugh, thanks, I thought I had found all of these.

ncm-iptables/src/main/perl/iptables.pm
jrha: Does now.

ncm-iptables/src/main/perl/iptables.pm
jrha: I'm now fairly sure this function doesn't actually work as documented, or at least the unit tests I have written seem to prove it doesn't.

ncm-iptables/src/main/perl/iptables.pm
jrha: I think it means that all file descriptors including `stdout` and `stderr` (but not @stdweird) are defined and therefore not left "dangling"

jrha:
I should note that everything I've pushed has been tested on a VM and the unit tests are all based on snippets of real configuration at RAL. So far the component has performed no worse than the unmodified code-base.

jrha:
`iptables-restore` has a (almost undocumented) `--test` option that could be used to check the validity of the configuration before applying it, I'll have a play with it.

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +784,32 @@ sub find_rule {
...
-		$self->error("command \"/sbin/service iptables condrestart\" failed:\n$ip_stdouterr");
-	    }
-	}
+    my $changes = $self->WriteFile($CONFIG_IPTABLES, $iptables);
+    $self->debug(5, "WriteFile returned $changes");
+    if ($changes) {
+        my $proc = CAF::Process->new([qw(/sbin/service iptables condrestart)], log=> $self);
+        my $ip_stdouterr = $proc->output();
+        if ($?) {
+            $self->error("command \"$proc\" failed:\n$ip_stdouterr");
```
stdweird: multiline logging should be avoided (but it's the case in the original code)

ncm-iptables/src/main/perl/iptables.pm
```diff
@@ -830,80 +784,32 @@ sub find_rule {
...
+        } else {
+            $self->info("ran command \"$proc\"");
+            $self->info($ip_stdouterr) if ($ip_stdouterr);
```
stdweird: this standalone `info` call should be avoided. use
```perl
$self->info("ran command \"$proc\"", $ip_stdouterr ? $ip_stdouterr : '');
```

stdweird:
@jrha minor remark, LGTM otherwise.
disclaimer: we do not use this at our site, people who do might want to be make sure they test  this during rc phase. unless noone objects, i'll merge this is before the 16.2 window closes

jrha:
I don't understand the failure of the tests on Jenkins, they pass on my desktop and (SL6) development VM and the component runs correctly on a (SL6) test VM.

stdweird:
@jrha it's actually odd it works on elsewhere. as of 1.49, you need to do a `use Test::Quattor;` before you do a `use NCM::Component::...` in the unittests (see https://github.com/quattor/configuration-modules-core/pull/730)

jouvin:
I think we use it at LAL and in fact still rely on an old version because of problems in newest... We may try to test it during RC.

jrha:
If you can remember the problems you had, feel free to add more unit tests!

jouvin:
No it was ages ago, during the StratusLab project... Will just check if it works as expected.

jrha:
Hooray! :tada: :green_apple: :green_book: :green_heart: :tea: :recycle: :fireworks:

stdweird:
@jrha any update on actually running this code? when you anncounce rc1, should really ask for testers, like those RAL people :grin:

jrha:
It's working perfectly on my test nodes, but yes I should get @drossy to test it :smile:

stdweird:
@jrha so merge and test in rc? or wait for a test before?

jrha:
Test in RC I think.

stdweird:
@jrha i like the way you think

PR #695, opened 2016-01-28
WIP: yumng - Jan 2016

ned21:
Another sync of our internal development efforts on yumng.

This code was forked from 14.10 and we had to make some schema changes it's becoming more difficult to sync back. Can we aim to get this merged at the workshop so at least we don't fall further behind?


ncm-spma/pom.xml
```diff
@@ -195,12 +224,32 @@
...
+              </mapping>
             </mappings>
             <requires>
               <require>yum &gt;= 3.2.29</require>
               <require>yum-utils &gt;= 1.1.30</require>
               <require>yum-versionlock</require>
               <require>yum-priorities</require>
+              <require>yum-fastestmirror</require>
+              <require>yum-list-data</require>
+              <require>redhat-release</require>
```
stdweird: what do you need from this package? (it's provided by `centos-release` on centos, but maybe we can abstract it a bit more)

ncm-spma/src/main/bin/spma
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+touch /.spma-run
+ncm-ncd --configure spma $*
```
stdweird: are you sure you need `$*` and not `"$@"` (incl the double quotes)?

ncm-spma/src/main/bin/spma
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+touch /.spma-run
```
stdweird: what is this used for?

ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -2,11 +2,10 @@
 # ${developer-info}
 # ${author-info}
 
-
 declaration template components/spma/schema;
 
 include 'quattor/schema';
-include 'components/spma/functions';
+include 'components/spma/functions-'+PACKAGER;
```
stdweird: a default value for `PACKAGER`? and maybe also for the 2 includes below?

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -5,6 +5,12 @@
...
-    "userpkgs_retry" : boolean = true  
-    "fullsearch" : boolean = false  
+    "userpkgs_retry" : boolean = true
+    "fullsearch" : boolean = false
+    "excludes"      ? string[] # packages to be excluded from metadata
+    "yumconf"       ? string # /etc/yum.conf YUM configuration
+    "whitelist"     ? string[] # packages not shipped by repositories but generated by 3rd party installer
+    "quattor_os_file" ? string # file to write quattor_os_release as confirmation of successful YUM spma pass
+    "quattor_os_release" ? string # string to write to quattor_os_file
+    "suffix"        ? string # suffix of alternative packaging module to include instead of standard one
```
stdweird: some restrictions?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+    return $status;
+}
+
+sub Configure
+{
+    my ( $self, $config ) = @_;
+
+    # Make sure stdout and stderr is flushed every time to not to
+    # put mess on serial console
+    autoflush STDOUT 1;
```
stdweird: this is already set by `ncm-ncd`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+    autoflush STDERR 1;
+
+    # We are parsing some outputs in this component. We must set a
+    # locale that we can understand.
+    local $ENV{LANG}   = 'C';
```
stdweird: @ned21 is there anything preventing us from setting this in `ncm-ncd`?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+    my $repos = $config->getElement(REPOS_TREE)->getTree();
+    my $t     = $config->getElement(CMP_TREE)->getTree();
+    
+    # Display system info
+    if ( defined($t->{quattor_os_release}) ) {
+	$self->info("target OS build: ", $t->{quattor_os_release});
+    }
+    
+    # Detect OS
+    my $fhi;
```
stdweird: can you use `CAF::FileReader`?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+rely on this!).
+
+Yum-based commands require annoying error handling: they may exit 0
+even upon errors.  In those cases, we have to detect errors by looking
+at stderr.  This method just encapsulates all that logic, keeping the
+callers clean.
+
+=cut
+
+my $os_major = "";
```
stdweird: i'm sorry, but i can't let you do this :smile:. define them where needed, unless they are truly intended to be used as class variables in instead of convenient way to redefine them.

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+		$os_major = substr($line, $i+8, 1);
+                last;
+            }
+        }
+        $fhi->close();
+    }
+
+    if ( $os_major eq "" ) {
+        $self->error("Unable to determine OS release.");
+	return 0;
```
stdweird: fix indentation

stdweird: also, `return 1` i think (error is enough)

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+    }
+
+    $self->execute_command( ["rpm -q --qf %{NAME}-%{VERSION}-%{RELEASE}.%{ARCH} ncm-spma"], "checking for spma version", 1);
```
stdweird: what are you trying to achieve here?

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+    return 1 if $cmd_exit;
+    $self->info("SPMA version: ", $cmd_out);
+    $self->info("user packages permitted: $t->{userpkgs}");
+
+    # Convert these crappily-defined fields into real Perl booleans.
+    $t->{run}      = $t->{run} eq 'yes';
+    $t->{userpkgs} = defined( $t->{userpkgs} ) && $t->{userpkgs} eq 'yes';
```
stdweird: no need for `defined()`

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+
+    # test for sharp spma execution early
+    my $spma_run = 0;
+    if ( -e "/.spma-run" ) {
```
stdweird: the location of this magic filenme has to be a constant/readonly

ncm-spma/src/main/perl/spma/yumng.pm
```diff
@@ -0,0 +1,866 @@
...
+        if ( !unlink "/.spma-run" ) {
+            $self->error("unable to remove file /.spma-run: $!");
+        }
+	$spma_run = 1;
```
stdweird: indentation

ncm-spma/src/main/bin/spma
ned21: This is a simple way of implementing the "run" functionality without needing a separate code base. The component checks if that file exists and only executes the package changes if it does. (The code currently does not honour the "run" attribute - that will be fixed shortly.

ncm-spma/src/main/pan/components/spma/yum/schema.pan
ned21: On the suffix? Or the quator_os_file? Or both?

ncm-spma/src/main/perl/spma/yumng.pm
ned21: Not AFAIK

ncm-spma/src/main/perl/spma/yumng.pm
ned21: Really?! Surely 1 should only be returned if it's successful? Indentation has been fixed in our internal repo, will be in next push.

ncm-spma/src/main/perl/spma/yumng.pm
ned21: userpkgs is an optional field so should this be exists() rather than defined()? Except that if it does exist but equal to undef then that will lead to an error when compared against "yes". Does defined() cover the exists case too?

Some experimentation with Perl shows that: ```defined( $t->{userpkgs} ) && $t->{userpkgs} eq 'yes';``` is probably safest, and is also what's currently in use in yum.pm

ncm-spma/pom.xml
jnovy: We should likely be good with system-release. Assuming centos-release provides it as well?

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: I had weird problems on RHEL5 with it missing. I can retry to confirm.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Need to recheck - had problems with CAF::FileReader in RHEL5. That's why it's done like this.

ncm-spma/pom.xml
jnovy: Actually - system-release doesn't work for RHEL5. Does centos-release provide redhat-release as well?

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: https://github.com/quattor/ncm-ncd/blob/master/src/main/perl/ComponentProxy.pm#L443
completely unused, so `0` is also fine (i always found it funny though; and making it meaningfull should be easy)

jnovy: Nope, it needs to be critical as YUM parsing code depends on OS version.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: We need to know major OS release as YUM output parsing depends on it.

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: I have this from upstream ;-)

ncm-spma/src/main/perl/spma/yumng.pm
jnovy: Indentation will be fixed globally.

ncm-spma/src/main/perl/spma/yumng.pm
stdweird: https://github.com/quattor/ncm-ncd/issues/54

ncm-spma/src/main/perl/spma/yumng.pm
jrha: If you have problems please open issues, otherwise nothing will ever improve.

ned21:
Superseded by #761

jrha:
Discussed!

PR #697, opened 2016-02-01
metaconfig: service logstash: support escaped add_field keys

stdweird:
Based on #691 and #660


jrha:
Needs rebasing now that #660 is merged?

stdweird:
@jrha done

PR #708, opened 2016-02-13
ncm-metaconfig: service ncm-ncd - do not try to restart daemons

stdweird:
Don't try to restart ncm-ncd, it is not a demon, so there is no ncm-ncd service!


stdweird:
@jrha this is a hotfix of some sort, have no clue why i hadn't run into this before

jrha:
Wow, what the hell, yeah not sure how we missed that.

PR #709, opened 2016-02-13
ncm-ccm: ccm group_readable option

stdweird:


stdweird:
@jrha related to https://github.com/quattor/CCM/pull/83 (if that PR does not get merged 16.2, this one should also be moved to next milestone)

ncm-ccm/src/main/pan/components/ccm/schema.pan
```diff
@@ -7,6 +7,7 @@ declaration template components/ccm/schema;
 
 include 'quattor/types/component';
 include 'pan/types';
+include 'components/accounts/functions';
```
ned21: This if for defined_group? Reading the definition from ncm-accounts, it looks like this assumes the group is created locally by quattor, i.e. it has no provision for using a group supplied by LDAP or any other NSS source. This feels overly restrictive.

stdweird: it can be a kept group too (like one created by an rpm etc). i don't think it's overly protective given the central role of ccm.
however, you will enter chicken-and-egg world if the group can't be found and quattor is supposed to config ldap or nss.
the current ccm code fails hard the group can't found. maybe it should fall back to root-only (so the system is not unusable?)

stdweird: @ned21 if group does not exist, it now fallsback to root only (https://github.com/stdweird/CCM/commit/d6a4f79992ad61bb474bdd2294fa8168a545aec2), so feel free to add the ldap group to `kept_groups`

ned21: There definitely feels like potential for a bootstrap problem here. I guess ncm-ccm is guaranteed to only run after the initial ccm bootstrap so it's safe to put things in here which depend on other parts of the tree (such as ncm-accounts). 

So a more practical concern is that this makes it difficult to use the useful group_readable functionality on sites that don't use ncm-accounts. Can we make that possible somehow?

stdweird: if the `group_readable` is configured and does not exist during `ccm-fetch` or `ccm-initialise` (the only parts of CCM that currently write; i.e. not this component), the group bits are not set by CCM (so `root.root` and possible world_readable).
only the next time around, new configuration settings will start to work. i have no clue how to make this better atm.
`ccm-fetch` will simply log error until the group is available, nothing more. (in particular, at present, `ncm-ccm` can't fix the group issue. even if it would log an error when the group does not exist (which it doesn't now); it would rerun and not change anything to the ccm data).

wrt to defined groups, when you are not using `ncm-accounts`, you could simply inject the data like the unittest does (maybe with a `kept_groups` like 
```pan
"/software/components/accounts/kept_groups/theadmins" = dict();
```
)
we could ofcourse extend the `defined_groups` to allow any string if `ncm-accounts` is not in use (at all or not active).
but then we should move the `defined_` functions out of the `ncm-accounts` functions imho.

i actually need to think about this a bit more, i'd like to play around with it a bit more. maybe we can merge this as tech preview :smile: 
(it reminds me that some changes in aii are required as well, PR soonish).

jrha:
I'm not sure I like the sounds of "tech preview", unless you have a compelling reason to merge this I'll bump it back to 16.4

stdweird:
@jrha usefullness is limited without https://github.com/quattor/CCM/pull/88 anyway

ncm-ccm/src/main/pan/components/ccm/schema.pan
stdweird: need to cover the possibility of not using `ncm-accounts`

stdweird:
@ned21 i addressed support for non-ncm-accounts in the schema. tests are ok again

ned21:
The include at line 10 will still fail if ncm-accounts is not in the site's template tree (we only pull in the templates for components we use). I can't think of a way around that off the top of my head, except moving the function definition to the file of shared functions in template-library-core rather than being in the component definition.

stdweird:
@ned21 i added an `if_exists` to the include, that should work

PR #710, opened 2016-02-13
ncm-spma: add support for more yum.conf options

### LINKED ISSUE #650: ncm-spma: add installonly_limit in yum.conf
For use with el7.2:
The initramfs files are now significantly bigger than in CentOS-7
(1503). You may want to consider lowering installonly_limit in
/etc/yum.conf to reduce the number of installed kernels if your /boot
partition is smaller than 4GB. New installations should consider
using 1GB as the size of the /boot partition.

stdweird:
Fixes #650

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/yum/schema.pan
+++ b/ncm-spma/src/main/pan/components/spma/yum/schema.pan
@@ -34,10 +34,10 @@ type spma_yum_plugins = {
 };
 
 @documentation{
-    Configuration options for yum.conf.
+    Main configuration options for yum.conf.
     The cleanup_on_remove, obsoletes, reposdir and pluginpath are set internally.
 }
-type spma_yum_conf = {
+type spma_yum_main_options = {
     "exclude" ? string[]
     "installonly_limit" ? long(0..) = 3
     "keepcache" ? boolean
@@ -49,5 +49,5 @@ type component_spma_yum = {
     "userpkgs_retry" : boolean = true
     "fullsearch" : boolean = false
     "plugins" ? spma_yum_plugins
-    "yumconf" ? spma_yum_conf
+    "main_options" ? spma_yum_main_options
 };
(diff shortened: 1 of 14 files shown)


ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -33,8 +33,21 @@ type spma_yum_plugins = {
...
+    "keepcache" ? boolean
+    "retries" ? long(0..) = 10
+    "timeout" ? long(0..) = 30
+};
+
 type component_spma_yum = {
     "userpkgs_retry" : boolean = true
     "fullsearch" : boolean = false
     "plugins" ? spma_yum_plugins
+    "yumconf" ? spma_yum_conf
```
jrha: Is there not a more descriptive name than `yumconf`?

stdweird: suggestions welcome. the way the schema is buildup, i'm reluctant to use something more generic.

jrha: I can't think of anything better that isn't very generic (`main`, `core`, etc.)

jrha: If we just called it `yum`, then the type could be `spma_yum_yum` :yum:

jrha: Can we call it `main_options`? It's what the man pages refer to these options as which makes it usefully googlable.

stdweird: the schema is shared with `yum` and `ips` (and `yumng`), and the options are not generic enough imho. but i'll change it to whatever is most suitable ;)

jrha: Eugh.

jrha: Call it `main_options` and I'll open an issue to have a think about how to improve the way the schema is built without breaking things.

stdweird: done

PR #711, opened 2016-02-13
ncm-spma: yum create and use cache

### LINKED ISSUE #689: ncm-spma: bug in cleanup of repos
it seems there is no cleanup between failed yum transaction and second attempt.

This way, some repos can be enabled that shouldn't be there in that stage

stdweird:
Fixes #689 
Based on #710


stdweird:
@jrha feel free to move the milestone to 16.4

stdweird:
this is not ready for 16.2

jrha:
Is it ready for 16.4?

stdweird:
we are running this everywhere

PR #713, opened 2016-02-15
ncm-metaconfig: icinga-web - Use FILTER to convert to uppercase, not string method

jrha:
upper() is not available in earlier versions of template toolkit so this breaks pre SL7 support.


jrha:
@stdweird this fixes the failure I saw in configuration-modules-core.

Issue #715, opened 2016-02-17
ncm-spma: Schema includes options for all packagers


jrha:
Current the schema includes both `component_spma_ips` and `component_spma_yum` regardless of which packager is actually being used, this is at best confusing.

This follows on from the discussion in #710.

PR #716, opened 2016-02-17
ncm-spma: Refactor schema and only include relevant types

jrha:
This is a proposed fix for #715.


jrha:
Untested, just a brain dump of what we might be able to do.

stdweird:
i think @jnovy did some similar/related work in the yumng PR

jrha:
That seemed to be about including the correct packaging functions for the packager (also important).

jrha:
retest this please

jrha:
Tests fail due to renaming of common schema causing confusion with Jenkin's reference copy of template-library-core.

stdweird:
@jrha merging the current PR will only break the unittest of all others PRs with same error. ()

to resolve the error and allow further cleanup, i propose to add a `pan/legacy` template to the template-library-core (and include it in `pan/types`) and define a new type
```pan
type fancy_name_for_yes_no_string_instead_of_boolean = string with match(SELF, "^(yes|no)$");
```
(i couldn't care less about the name, the worse the better imho; people might actually feel some resistance to continue using it :smile: )

and use this new type name in this PR. lot of components already have this or also create a new type (e.g. `authconfig` has `yesnostring`, `ssh` has `ssh_yesnostring`), and those should then also be cleaned up.

if you would use the same type name as any of the current one from spma//authconfig/ssh, do not forget to delete the type in the component schema in the template library core PR (i know they are generated, but this is chicken and egg situation).

jrha:
Will need rebasing after #798 is merged.

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
```diff
@@ -0,0 +1,73 @@
...
+
+declaration template components/spma/yumng/schema;
+
+include 'components/spma/schema';
+
+type SOFTWARE_GROUP = {
+    "default" : boolean = true
+    "mandatory" : boolean = true
+    "optional" : boolean = false
+    "names" : string[]
```
ned21: aha - if yumng can have its own schema then we can drop this and revert to the code we actually run where /software/groups is a list not a dict. I will send a patch for that tomorrow.

stdweird: i've probably missed the discussion why a dict is not ok for yumng, but  /software/groups ideally has the same type for all backends.

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
```diff
@@ -0,0 +1,73 @@
...
+    "whitelist"     ? string[] # packages not shipped by repositories but generated by 3rd party installer
+    "quattor_os_file" ? string # file to write quattor_os_release as confirmation of successful YUM spma pass
+    "quattor_os_release" ? string # string to write to quattor_os_file
+    "suffix"        ? string # suffix of alternative packaging module to include instead of standard one
+    "plugins" ? spma_yum_plugins
+    "main_options" ? spma_yum_main_options
+};
+
+bind "/software/components/spma" = component_spma_yum;
+bind "/software/groups" = SOFTWARE_GROUP{};
```
ned21: My earlier comment was slightly incorrect. The incompatibility in the schema comes from yumng requiring this to be: ```
bind "/software/groups" = SOFTWARE_GROUP
```

(i.e. a simple resource and not an nlist).

File: ncm-spma/src/main/pan/components/spma/yumng/schema.pan (hunk 1 above)
Comment (ned21) [inline ncm-spma/src/main/pan/components/spma/yumng/schema.pan]: Me too. :( I think the problem was that the removal of the {} on the line below was so subtle it didn't get caught in code review. :(

File: ncm-spma/src/main/pan/components/spma/yumng/schema.pan (hunk 2 above)
Comment (stdweird) [inline ncm-spma/src/main/pan/components/spma/yumng/schema.pan]: ah, and that's why the https://github.com/quattor/configuration-modules-core/pull/807 was needed. but yumng could easily be modified to accept a dict with single key that has no real meaning.
the yumng schema could actually add another bind to verify that there is only one key (and even hardcode the key name)

e.g. additonally (ie not as a replacement)
```pan
bind "/software/groups" = SOFTWARE_GROUP{} with length(SELF) == 1 && is_defined(SELF['yumng']);
```

stdweird: actually, the code already assumes this (`/software/groups/yumng/names` is what is used) (well, from a quick glance at the code)

ned21: That's what I tried but as reported in #807 it seems to break yum because it interprets yumng as a group name. So we would need a separate schema for yumng anyway.

stdweird: but then this line could also be
```pan
bind "/software/groups/yumng" = SOFTWARE_GROUP;
```

ned21: Yep - I've just sent @jrha a PR for that.

ncm-spma/src/main/pan/components/spma/config.pan
```diff
@@ -5,10 +5,10 @@
 
 unique template components/${project.artifactId}/config;
 
-variable CONFIG_MODULES_CONFIG_SUFFIX ?= 'rpm';
+variable SPMA_BACKEND ?= 'yum';
```
ned21: Why do we have the SPMA_BACKEND variable as well as the /software/components/spma/packager element? Can we just use the latter instead of the global?

ncm-spma/src/main/pan/components/spma/yumng/config.pan
```diff
@@ -0,0 +1,22 @@
+# ${license-info}
+# ${developer-info}
+# ${author-info}
+
+unique template components/spma/yumng/config;
+
+# Prefix for packages/groups
+prefix '/software';
+'groups/yumng/names' ?= list();
```
ned21: This should be 'groups/names' ?= list(); but even better, let's move it to schema.pan

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
ned21: As per the comment on config.pan above, let's make ```"names" : string[] = list()``` to avoid having to set the default in config.pan

ncm-spma/src/main/pan/components/spma/config.pan
jrha: I don't know!

ncm-spma/src/main/pan/components/spma/schema.pan
```diff
@@ -4,18 +4,14 @@
...
 include 'components/spma/software';
 
-type component_spma_type = {
-    include structure_component
-    include component_spma_ips
-    include component_spma_yum
+type component_spma_common = {
     "cmdfile"       : string = "/var/tmp/spma-commands" # where to save commands for spma-run script
-    "packager"      : string = "yum" with match (SELF, '^(yum|ips|rpm)$') # system packager to be used (yum,ips,rpm)
+    "packager"      : string = "yum" with match (SELF, '^(yum|yumng|ips)$') # system packager to be used
```
ned21: yumng isn't actually valid. See [spma.pm](https://github.com/quattor/configuration-modules-core/blob/master/ncm-spma/src/main/perl/spma.pm#L49). @jnovy had an idea for how this would be used to implement DNF support along side yum for pre-DNF OSes.

jrha: I'm being stupid I'm sure, but I don't understand why it isn't valid.

ncm-spma/src/main/pan/components/spma/config.pan
jouvin: @ned21 @jrha I think I may have introduced it... Basically the same thing as in the template library. This provides greater flexibility to use a variable that direct assignment to the path and allows to change the schema without impacting the site configuration (it is enough to assign the variable (if necessary after some massaging) to the new schema element. It also allows to define the variable very early in the config without having the `ncm-spma` configuration already defined.

jouvin: As for the variable name change, it's fine with me. I checked in the template library and found no usage of it outside `ncm-spma`.

ncm-spma/src/main/pan/components/spma/schema.pan
ned21: The way you switch between yum and yumng is you set the packager to yum and then set the "suffix" attribute to ng. The argument is that this keeps "packager" correct (it's still yum) but allows multiple variants of the packager (thinking ahead to DNF). Hopefully @jnovy can explain his idea in more detail.

jrha: Hmm ok, I missed that. It seems more complex to me, but I'll @jnovy explain.

jnovy:
@ned21 described the point correctly - one can simply check whether packager == "yum" so that one can detect rpm vs auto-updated builds.

For DNF there is a possibility to implement a separate module as "yumng" using a different suffix, i.e. "yumdnf". It would be cool if we can use also separate configurable schema for it on top of the default one.

jrha:
@jnovy I'm not sure having two places to change the module is a good idea, both will need validation (suffix currently has none), the restructuring allows modules to have separate schemas and your example of comparison could be done with `match(packager, '/^yum/')`.

I'm inclined to remove it unless anyone really objects.

jrha:
Discussed with @jnovy at workshop and agreed to remove `suffix`, but we will document a naming scheme expected of packager back-ends to assist with identification in pan code.

jrha:
Updated to remove suffix following discussions at workshop, added backward incompatible tag as this may catch someone out and rebased on upstream master.

ncm-spma/src/main/pan/components/spma/yumng/config.pan
ned21: This does not match line 29 of ncm-spma/src/main/perl/spma/yumng.pm. If schema.pan sets it to list() by default then this line can be dropped.

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
ned21: Set names to list() by default

jrha:
Thanks, fixed.

ncm-spma/src/main/pan/components/spma/schema.pan
jrha: So I've removed `suffix` from the schema and code following a discussion with @jnovy at the last workshop.

jouvin:
Checked quickly but not tested... LGTM.

jouvin:
@ned21 is your concern addressed by last change?

ned21:
Changes are OK. Is this running anywhere or is testing still required?

jrha:
The yum back-end has been tested at RAL, I've just fixed the last commit which mistakenly removed the empty `dict()` default in the yum `config.pan`.

PR #721, opened 2016-02-17
ncm-spma: Support yum gpgkey/gpgcakey/repo_gpgcheck repository settings

### LINKED ISSUE #717: ncm-spma: yum schema does not support gpgkey option
The yum schema has support for the gpgcheck option to check GPG signatures on packages but does not support the gpgkey option to allow per repository gpg keys - we sign our locally built rpms and have a seperate repository for them and it would be good to be able to specify the key in quattor

stdweird:
Fixes #717


stdweird:
@jrha you can change the milestone :smile:

jrha:
gpgcakey = gpg:cake:y?

ncm-spma/src/main/pan/components/spma/software.pan
```diff
@@ -7,6 +7,8 @@ declaration template components/spma/software;
 include 'pan/types';
 include 'components/spma/functions';
 
+type software_repository_url = string with match(SELF, '^(file|https?|ftp)://');
```
jrha: What about one of the URI types from `pan/types`?

stdweird: i saw that, but have no clue what ips wants.

jrha: @msmark?

jrha:
Good enough for now, will open an issue to think about the URL type.

PR #723, opened 2016-02-20
ncm-metaconfig: add service nrpe to replace ncm-nrpe

### LINKED ISSUE #714: fix ncm-nrpe nrpe restart
Code is missing a test that the restart only happens when the config file is modified. best fixed by replacing with metaconfig

stdweird:
Fixes #714


ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
```diff
@@ -0,0 +1,27 @@
...
+declaration template metaconfig/nrpe/schema;
+
+include 'pan/types';
+
+@{Configuration options for Nagios Remote Plugin Executor (NRPE)}
+type nrpe_service = {
+    'log_facility'              : string = 'daemon'
+    'pid_file'                  : string = '/var/run/nrpe.pid'
+    'server_port'               : type_port = 5666
+    'server_address'            ? string
```
jrha: How about one of the network types from `pan/types`?

ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
```diff
@@ -0,0 +1,27 @@
...
+    'nrpe_user'                 : string = 'nagios'
+    'nrpe_group'                : string = 'nagios'
+    'allowed_hosts'             : type_hostname[]
+    'dont_blame_nrpe'           : boolean = false
+    'command_prefix'            ? string
+    'debug'                     : boolean = false
+    'command_timeout'           : long = 60
+    'connection_timeout'        : long = 300
```
jrha: Validate that timeout is a positive number?

jrha:
Looks good, but feels like it needs more validation rules.

ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
jrha: As I understand it, this is the address for nrpe to bind to on start, so either `type_ip` or `type_hostname` (if hostnames are resolved by nrpe) should work here.

ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
```diff
@@ -0,0 +1,27 @@
...
+# ${author-info}
+
+declaration template metaconfig/nrpe/schema;
+
+include 'pan/types';
+
+@{Configuration options for Nagios Remote Plugin Executor (NRPE)}
+type nrpe_service = {
+    'log_facility'              : string = 'daemon'
+    'pid_file'                  : string = '/var/run/nrpe.pid'
```
jrha: Presumably this **must** be an absolute path to a file, perhaps we should split out the inner part of `valid_absolute_file_paths` to allow it to be used on a single string.

ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
jrha: Comments in `nrpe.cfg` say "Do not set this too low." I wonder what that means :grin:

ncm-metaconfig/src/main/metaconfig/nrpe/pan/schema.pan
stdweird: i opened https://github.com/quattor/template-library-core/issues/116

stdweird:
@jrha remarks addressed, except the abspath one.

PR #724, opened 2016-02-21
ncm-nfs: rewrite for NoActionSupported and unittests

### LINKED ISSUE #607: ncm-nfs: code cleanup
* Replace LC::Check and NCM::Check usage by CAF
* use CAF::FileReader instead of open
* Identify overlap with other fstab editing code from ncm-fstab / lib-blockdevices

This issue is not meant to replace ncm-nfs

stdweird:
Backwards incompatible due to changed behaviour: unused mountpoints are not cleaned up anymore.
Fixes #607.


stdweird:
the WIP is about actual testing and more unittests

stdweird:
@ned21 should i just squash all commits? it started out as a cleanup, but it's basically a rewrite

ned21:
The first 4 commits make sense as they make no functional changes. 8038e75's log message of "cleanup" is a bit vague, and if that's the start of a re-write then I could see that squashing it with c58b0ad makes sense. I've read and agree with the refactoring commits, need to spend some more time reviewing the rewrite.

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -2,305 +2,364 @@
...
+Readonly my $FSTAB_ACTION_UMOUNT_MOUNT => 'umount/mount';
+
+Readonly::Array my @CMD_UMOUNT_LAZY => qw(umount -l);
+Readonly::Array my @CMD_MOUNT => qw(mount);
+Readonly::Array my @CMD_REMOUNT => qw(mount -o remount);
+
+# Compares two fstab entries C<new> and C<old> for equality,
+# and return action to be taken.
+# If old does not exist, mount.
+# If equal, do nothing.
```
kwaegema: Should mounts which are unchanged in fstab be mounted if they aren't ?

stdweird: all entries in fstab should be mounted (it's what happens at boot)

stdweird:
@ned21 rewite is done, please advise on what to squash :smile: 
also there's a TODO in the code i think it was a bug in the old code, so some comment there is welcome

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -2,305 +2,468 @@
...
+
+    return $FSTAB_ACTION_MOUNT if (! $old);
+
+    # Can numeric 'A != B' be different than 'A ne B' ?
+    my $ne = sub {
+        my $attr = shift;
+        return $new->{$attr} ne $old->{$attr};
+    };
+
+    return $FSTAB_ACTION_UMOUNT_MOUNT if (&$ne($FSTAB_DEVICE) || &$ne($FSTAB_MOUNTPOINT));
```
ned21: The amount of SHOUTING in this method makes it hard to read. The multiple exit points via end of line conditionals are also why this style gets a bad name. This looks like a good fit for a switch/given statement?

To make this code more extensible, how about comparing the hash and computing all the differences, and then switching on the result of that computation? e.g.
```
my %diff = find_keys_that_differ();
my $ret = FSTAB_ACTION_NONE;
if ( $diff{$FSTAB_DEVICE} || diff{$FSTAB_MOUNTPOINT}) {
   $ret = $FSTAB_ACTION_UMOUNT_MOUNT
} elif {
...
}
return $ret;
```

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -2,305 +2,468 @@
...
+
+sub exports
+{
+    my ($self, $tree) = @_;
+
+    # Now just create the new configuration file.  Be careful to save
+    # a backup of the previous file if necessary.
+    my $fh = CAF::FileWriter->new($EXPORTS, backup => ".old", log => $self);
+
+    # No timestamp anymore, otherwsie file changes trivially
```
ned21: This comment is temporal and should be moved to the commit log. Or rewrite as:
# Do not include timestamp, otherwise file changes every run

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -2,305 +2,468 @@
...
+
+    return fstab_add_defaults(\%fstab);
+};
+
+
+sub fstab
+{
+    my ($self, $tree) = @_;
+
+    # Now just create the new configuration file.  Be careful to save
```
ned21: "Now" makes no sense at the start of a method. A method should have POD

ned21:
I think the only squash that is needed is 8038e75 into the subsequent commit. 8038e75's log is too vague the changes would make more sense as part of the rewrite.

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -2,305 +2,468 @@
...
+            $action_taken = 1;
+            if ($self->do_mount(\@CMD_UMOUNT_LAZY, $old->{$device})) {
+                # TODO: previous code always did his,
+                #       without recreating the mountpoint for UMOUNT_MOUNT devices
+                #       That was probably a bug
+                if (! defined($new->{$device})) {
+                # Try removing mount point, giving warning on error.
+                    my $mntpt = $old->{$device}->{$FSTAB_MOUNTPOINT};
+                    # Disable any backup
+                    if (! $self->_cleanup($mntpt, '')) {
```
ned21: Is it worth even removing the mount point? I would leave it alone and not clean up.

ncm-nfs/src/main/perl/nfs.pm
stdweird: what do you mean with SHOUTING?
i'll try to clean it up again (wasn't that happy with the result myself)

ncm-nfs/src/main/perl/nfs.pm
stdweird: remove it or make it conditional? (and if it is conditional, was the old code doing the wrong thing?)

stdweird:
@ned21 remarks addressed, only need to know if we should remove the cleanup part or make it conditional (i don't mind removing it). maybe wait for some feedback from @jouvin @jrha ?

ned21:
My vote is to remove it altogether. The cost of the code complexity seems to outweigh the usefulness of the feature but happy to be convinced otherwise!

stdweird:
Removed the cleanup functionality in in commit https://github.com/stdweird/configuration-modules-core/commit/cc05314956ecec026d93fca27658bc8efba97da5, can be reverted later if people insist

ned21:
LGTM

kwaegema:
fixes #607

PR #726, opened 2016-02-22
Add templates for new issues and pull-requests

jrha:
The practices we have adopted for managing issues and pull-requests on GitHub have not been documented anywhere, this has the advantage of both documenting those practices and then directly exposing them to everyone.

See https://github.com/blog/2111-issue-and-pull-request-templates

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,11 @@
+Ensure the title of this issue starts with the component name followed by a colon, e.g.
+> ncm-example: demonstrate what titles should look like
+
+Describe the problem here.
+
+If this is a bug report, remember to include:
+
+* The version of the component you are using (rpm -q $component)
+* What OS version you are using (/etc/redhat-release or similar)
+* Steps required to reproduce the problem
+* Actual result versus expected result
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,7 @@
+Ensure the title of this pull-request starts with the component name followed by a colon, e.g.
+> ncm-example: demonstrate what titles should look like
+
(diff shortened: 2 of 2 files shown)


jrha:
If we decide to adopt this, I will open corresponding PRs for the other repositories.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+Ensure the title of this pull-request starts with the component name followed by a colon, e.g.
+> ncm-example: demonstrate what titles should look like
+
+Describe the change you are making here, in particular we would like to know:
+
+* Why the change is necessary.
+* What backwards incompatibility it may introduce.
```
stdweird: (and if so, add label)

jrha: Can anyone add labels?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,4 @@
+Ensure the title of this issue starts with the component name followed by a colon, e.g.
+> ncm-example: demonstrate what titles should look like
+
+Describe the problem here.
```
ned21: Describe the problem here, remembering to include:
- what version of the component you are using (rpm -q $component)
- what OS version you are using (/etc/redhat-release or similar)
- steps to reproduce
- actual result versus expected result

Or something similar? (I guess not all issues are bug reports so perhaps this is a bit too specific?)

jrha: I've added some text below for this.

ned21:
LGTM

.github/PULL_REQUEST_TEMPLATE.md
jrha: Just got a colleague to check, only us developers are privileged enough to set the labels.

PR #731, opened 2016-02-29
ncm-ccm: allow lowercase characters in trust realm.

ajf8:
Using upper case characters for Kerberos realms is the convention, but
some organisations may be using lower case realms without issue.


stdweird:
LGTM

PR #732, opened 2016-03-04
ncm-syslogng: fix flags support.

ajf8:
it passes $cfg->{flags} into print_log_flags, then checks this hash for flags too, so it always returns and never prints any flags.

### DIFF
--- a/ncm-syslogng/src/main/perl/syslogng.pm
+++ b/ncm-syslogng/src/main/perl/syslogng.pm
@@ -136,14 +136,9 @@ sub print_log_flags
 {
 	my ($fh, $cfg) = @_;
 
-	return unless exists $cfg->{flags};
+    my @flags = grep {$cfg->{$_}} sort keys %$cfg; 
 
-	my @flags;
-	while (my ($k, $v) = each (%$cfg)) {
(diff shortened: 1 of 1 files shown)


stdweird:
@ajf8 do you want me to comment on all the legacy crap, like wrong indentation and unsorted each :smile: 
nice catch.

ncm-syslogng/src/main/perl/syslogng.pm
```diff
@@ -136,14 +136,12 @@ sub print_log_flags
 {
 	my ($fh, $cfg) = @_;
 
-	return unless exists $cfg->{flags};
-
 	my @flags;
```
stdweird: sorted and one line:
```perl
my @flags = grep {$cfg->{$_}} sort keys %$cfg; 
```
(must... resists... other remarks ....)

ajf8:
I've made the one change you suggested (and retested), but I don't really have time to work on this component any further :) mainly because (for reasons involving a bunch of legacy) we're moving away from this particular component at some point. We won't be using this fix, I just wanted to send it upstream anyway.

stdweird:
@ajf8 ok, thx for doing so!

LGTM

PR #734, opened 2016-03-14
ncm-opennebula: Include VNET update_ar

### LINKED ISSUE #678: ncm-opennebula: AR sections are not updated after virtual network creation
At this moment if the VNET already exists we use `one.vn.update` method to update the whole VNET template, as example:

```
AR = [
ip = "10.120.1.100",
size = "20",
type = "IP4"
]
BRIDGE = "br1"
DNS = "10.120.1.1"
GATEWAY = "10.120.1.1"
NAME = "mypool"
NETWORK_MASK = "255.255.0.0"
VLAN = "YES"
```
But using this method only `BRIDGE= ...DNS=` will be updated, if you want to update also the ARs you must use `one.vn.update_ar` method as well.
We have to change the logic of the ONE component function to detect these changes and to update the virtual networks in a correct way.

Another important thing is that in order to update the AR OpenNebula requires the AR id and should be included within AR section like this:

```
AR = [
ar_id = "1",
ip = "10.120.1.100",
size = "20",
type = "IP4"
]
```

If not we get this error message from RPC and the AR is not updated:
```
[VirtualNetworkUpdateAddressRange] Error modifying network leases. AR/AR_ID attribute is missing. (code 8192)
```

alvarosimon:
* Fix #678 AR sections are not updated after virtual network creation

After VNETs and ARs (address ranges) creation the available VNET ARs must be updated using `one.vn.update_ar` method, `one.vn.update`is not enough in those cases.


ncm-opennebula/src/main/perl/opennebula.pm
```diff
@@ -105,7 +105,7 @@ sub create_or_update_something
         $new = $one->$cmethod("$template");
```
stdweird: bump required `NEt::OpenNebula` version

alvarosimon: @stdweird fixed, and sorry for the noise we don't need a new `NEt::OpenNebula` version

stdweird:
@jrha LGTM

PR #736, opened 2016-03-15
ncm-metaconfig: add service for limits.conf file

kwaegema:


ncm-metaconfig/src/main/metaconfig/limits_conf/pan/schema.pan
```diff
@@ -0,0 +1,19 @@
...
+type limits_conf_entry = {
+    'domain' : string
+    'type' : string with match(SELF, '^(soft|hard|-)$')
+    'item' : limits_conf_item
+    'value' : long(-1..)
+};
+
+@{ type for configuring the limits.conf file @}
+type limits_conf_file = {
+    'entries'           : limits_conf_entry[]
```
stdweird: too much whitespace before `:`

ncm-metaconfig/src/main/metaconfig/limits_conf/tests/regexps/config
```diff
@@ -0,0 +1,9 @@
+Base test for limitsconfig
+---
+/etc/security/limits.d/99-quattor.conf
+---
+^ftp\s{4}hard\s{4}priority\s{4}5$
+^student\s{4}soft\s{4}maxlogins\s{4}1$
+^ceph\s{4}-\s{4}nproc\s{4}1048576$
```
stdweird: don't tell me we use this for ceph? on el7, all service limits should be configured via the unit limits

ncm-metaconfig/src/main/metaconfig/limits_conf/pan/config.pan
```diff
@@ -0,0 +1,8 @@
+unique template metaconfig/limits_conf/config;
+
+include 'metaconfig/limits_conf/schema';
+
+bind "/software/components/metaconfig/services/{/etc/security/limits.d/99-quattor.conf}/contents" = limits_conf_file;
```
ned21: IMHO 99-quattor is too limited. What if some other system wants to deliver an override? This should go one higher than the OS default, or somewhere with some leeway for future expansion.

ncm-metaconfig/src/main/metaconfig/limits_conf/tests/regexps/config
kwaegema: I use it to specify the ceph user limits. this is shared between the ceph processes and ceph-deploy

jrha:
LGTM

PR #745, opened 2016-03-29
ncm-ssh: Fix for multiline option values with special chars.

ajf8:
The escaping was for the regex passed into add_or_replace_lines, not the
sshd_config file itself. Multiline options don't use regex, so the
values should not have been escaped in the same way.


stdweird:
@ajf8 unittests would be nice, but non blocking until we get the docs sorted out how add new ones
LGTM

PR #747, opened 2016-04-04
ncm-opennebula: Update oneadmin user template and user public ssh keys

### LINKED ISSUE #746: ncm-opennebula: Allow ONE component to update oneadmin user variables
At this moment the OpenNebula component creates and updates oneadmin password but it does not change anything else (`ssh_public_key` variable is not set or updated for example). 

This is done on purpose due to `QUATTOR` variable is not set for this user (oneadmin user is generated automatically by opennebula). We have to detect this special case and update the oneadmin variables if they are set.

alvarosimon:
**NOTE:** This PR depends on #734, PR 734 should be reviewed/merged first

 * Fix #746 Allow ONE component to update oneadmin user variables


ncm-opennebula/src/main/resources/tests/regexps/user_one/simple
```diff
@@ -2,5 +2,6 @@ Opennebula user test
 ---
 multiline
 ---
-^SSH_PUBLIC_KEY\s?=\s?".+"\s*$
+^SSH_PUBLIC_KEY\s?=\s?".+\s*$
+^\"$
```
stdweird: extend the regexp to make sure 2 keys are set

stdweird:
@alvarosimon minor remark about the unittest. do we have this running already?

alvarosimon:
@stdweird yes I'm testing it in altaria with `cloud-init` to insert admins ssh pub keys

jrha:
LGTM. #734 is merged.

PR #749, opened 2016-04-11
ncm-*: Remove variants of yesno pseudo-boolean

jrha:
Standardise on new type from pan/legacy

quattor/template-library-core#111 introduced a common type for the variously named versions of a string containing either yes or no, this switches the components to use the new common version rather than their own differently named types.

It does not yes address the components that define types of string with a validation rule to essentially do the same thing, but without defining an explicit type.

Turns out coding on a train isn't that easy... just left Stirling, next stop Gleneagles.


stdweird:
@jrha is this ok to merge (do you have more commits?) also, maybe create an issue to handle the remaining cleanup

jrha:
You can merge this, I probably won't have any more time to finish this until I get back, also the wireless in the Hotel in Pitlochry is getting overloaded by our meeting.

PR #751, opened 2016-04-12
ncm-metaconfig: httpd support GSSAPI auth

stdweird:
Add support for GSSAPI auth and example/unittests of apache config for retrieving profiles using kerberos


jrha:
Looks reasonable, but I have no experience with Apache and GSSAPI.

jrha:
@stdweird where did all the ncm-nfs commits come from?

stdweird:
@jrha rebased

jrha:
Thanks!

PR #756, opened 2016-04-14
ncm-ceph: add options to schema

kwaegema:
add purge options to mds and op threads to osds


jrha:
LGTM, will merge after @gvasilak takes a look.

PR #727, opened 2016-02-22
ncm-metaconfig: Add support for haproxy

apdibbo:
Add a component for using metaconfig to configure haproxy

PR #757, opened 2016-04-23
ncm-acccounts: fix failing unittests

stdweird:
Address the failing unittests in unrelated #727


stdweird:
@jrha seems like the failure went away with bumpioing the build-tools.  but the actual cause is not fixed.
@apdibbo if you're lucky, a rebase on master after this is merged will be sufficient. if not, i'll be able to reproduce the issue again :smile:

PR #762, opened 2016-05-06
Unittest hotfixes

stdweird:
Handle ncm-postgres unittest failure due to new json_typed default ; and prepare ncm-spma and ncm-nfs for new buildtools


stdweird:
This does not fix @apdibbo fialing tests nor the spma failures from @ned21

PR #765, opened 2016-05-11
ncm-filesystems: pom.xml fix rpmlint 'description too long' error

stdweird:


jrha:
Are the brackets meaningful any more?

stdweird:
i'd say yes because it's called `ncm-filesystems` and not `ncm-filesystems-and-blockdevices`

PR #766, opened 2016-05-11
ncm-named: resolv.conf: generate correct options

### LINKED ISSUE #763: ncm-named: Options not correctly passed through to resolv.conf
When I add a line like

`"/software/components/named/options" = list('timeout:2');`

to my quattor configuration, the JSON appears to be correctly created:

      "named": {
        "active": true,
        "dependencies": {
          "pre": [
            "spma"
          ]
        },
        "dispatch": true,
        "options": [
          "timeout:2"
        ],

However when the component runs the entry added to resolv.conf is 

`options ARRAY(0x1aadb48)`

whereas I'd expect it to be 

`options timeout:2`

I'm using version 15.2.0 on RHEL 6.7

stdweird:
Fixes #763


stdweird:
@drossy feel free to try the patch, thanks for reporting!

PR #767, opened 2016-05-12
ncm-metaconfig: lmod: scDescript needs quoted values

stdweird:
Values without quotes generates an invalid lmod rc file.


stdweird:
retest this please

PR #769, opened 2016-05-13
Remove DTA typeglob

### LINKED ISSUE #692: remove 'local *DTA'
some of the components have a 
```perl
local(*DTA);
```
statement, but i have no clue why it's there or what it does (`DTA` is not used anyewhere else and only used in this way). googling for `DTA` revealed nothing, so i don't think it's a perlism.

we also have components that do not have this (and work fine), so i'm inclined to remove this unless someone knows something more.

stdweird:
Fixes #692

Also remove `use warnings` as it is mainly used to point to coding issues — most warnings are non-fatal, so this should be harmless.


stdweird:
Best viewed with https://github.com/quattor/configuration-modules-core/pull/769?w=1

ned21:
LGTM although I am slightly wary of introducing such a large change this close to release. Wait for 16.6?

stdweird:
@ned21 @jrha i don't mind, but then it should be merged right after the release (the whitespace changes have a high probability for merge conflicts later on)

jouvin:
I had a look with whitespace changes ignored and it looks fine. If anything was wrong in what has been modified, the `00-load.t` test would detect it (the main change is the removal of `()` in method declaration and `{ ` opening brace put on a separate line, any mistake in them would produce a syntax error).
Personnally, I support merging it **right now**.

jrha:
Can you remind us why you are also removing `use warnings;` in the PR description please?

stdweird:
@jrha the main reason of `use warnings` is to point to coding issues. as most warnings are non-fatal, this should be harmless.

PR #770, opened 2016-05-17
ncm-dirperm: add 'accounts' to list of pre dependencies

brunocanning:
I think the accounts component should be added to the list of pre dependencies for the dirperm component. If a new profile arrives on a host with dirperm config for accounts that do not exist, the component will err. If the accounts component were to run beforehand and create new accounts for which new dirperm config has been written, the dirperm component will run successfully.

I doubt this will introduce any backwards incompatibility.

### DIFF
--- a/ncm-dirperm/src/main/pan/components/dirperm/config.pan
+++ b/ncm-dirperm/src/main/pan/components/dirperm/config.pan
@@ -12,7 +12,7 @@ include "components/dirperm/schema";
 
 prefix '/software/components/${project.artifactId}';
 
-'dependencies/pre' ?= list('spma');
+'dependencies/pre' ?= list('spma', 'accounts');
 'register_change' ?= list('/system/filesystems');
 'version' = '${no-snapshot-version}';
 'active' ?= true;


brunocanning:
Yes.

jrha:
please test this

ned21:
Unfortunately not everyone uses ncm-accounts. e.g. we have nss in our pre-dependencies because that is what configures our user and group information. This  config should be easy enough to go into your site/ directory?

brunocanning:
@ned21 I hadn't thought of that, it is a good point. It came up from a colleague at our site (Rutherford Appleton Laboratory) asking if ncm-dirperm should cause ncm-accounts to run beforehand as standard. I added accounts as a pre dependency of dirperm in the feature I was working on as I wanted directories created for a new account and dirperm failled to run as the accounts had not yet been created.

jouvin:
@brunocanning This is an issue discussed a number of times but the agreement so far has been that one component should not have an hardcoded dependency on another one in its templates. But rather document that the it may have some requirements and let the site do what is appropriate. The template library in particular takes care of doing the right/sensible things!

stdweird:
or components need the systemd equivalent of `after` and `before`: 
```pan
'dependices/pre' = list('spma');
'dependencies/before' = list('accoutns', 'nss');
```

meaning: spma must be included and run before dirperm; and if accounts or nss is configured, run it before dirperm (but none are required)

brunocanning:
@stdweird that sounds like a solution, force dirperm to run after accounts and nss _if_ configured. I like it.

brunocanning:
@jouvin yes, I realise the mistake after @ned21's contribution.

stdweird:
@brunocanning start with opening an issue in `ncm-ncd` :smile:

jrha:
Closing for now, if you come up with a more suitable solution please feel free to reopen.

PR #775, opened 2016-05-24
ncm-filecopy: Update filecopy.pod example to make it work

GRyall:
The config value in the second example needs double quotes for the escape character to work


stdweird:
test this please

jrha:
Thanks!

PR #781, opened 2016-06-02
ncm-nss: wait for nscd to restart before returning

ned21:
This is a very simple change we'd like in 16.4. I have a another branch that replaces LC::Check with FileWriter and includes unit tests which I can push for 16.6.

### DIFF
--- a/ncm-nss/src/main/perl/nss.pm
+++ b/ncm-nss/src/main/perl/nss.pm
@@ -113,6 +113,10 @@ sub Configure {
             my $nscdrestart = CAF::Process->new(
                 ['service', 'nscd', 'condrestart'], log => $self);
             $nscdrestart->execute();
+	    # wait for nscd to restart before returning in case
+	    # another component in the same ncm-ncd run depends on
+	    # any changed config
+	    sleep 2;
         } else {
             $self->warn("Unrecognised platform, $^O, not restarting nscd");
         }


ncm-nss/src/main/perl/nss.pm
```diff
@@ -113,6 +113,10 @@ sub Configure {
             my $nscdrestart = CAF::Process->new(
                 ['service', 'nscd', 'condrestart'], log => $self);
             $nscdrestart->execute();
+	    # wait for nscd to restart before returning in case
```
stdweird: are you aware of https://github.com/quattor/configuration-modules-core/blob/master/ncm-authconfig/src/main/perl/authconfig.pm#L379? it's not condrestart, but for sure does not assume this a simple restart will just work.

ned21:
Unit test failure seems unrelated to this change?

ncm-nss/src/main/perl/nss.pm
ned21: I was not and we've not seen that particular problem.

stdweird:
@ned21 @alvarosimon is on it. expect a PR soonish

alvarosimon:
@ned21 @stdweird https://github.com/quattor/configuration-modules-core/pull/782 should fix ncm-opennebula issue

stdweird:
retest this please

PR #782, opened 2016-06-03
ncm-opennebula: Fix RPC history

alvarosimon:
New RPCClient.pm generates less calls to RPC endpoint. This PR updates RPC unittests to use this new version.


stdweird:
@alvarosimon thanks
@ned21 @jrha please merge this to restore the unittest

PR #783, opened 2016-06-03
ncm-grub: support setting password in grub.conf.

ajf8:

We have a requirement to set grub bootloader passwords.

Should be none. This is why I added the optional "enabled" field on the password type - I didn't want the component to remove any password in the config (since ncm-grub does not assume full control of the file), but wanted a way to remove a password if necessary.

It is possible to set the password in the profile, or alternatively refer to a file which the component will use to look it up. This could be useful where it is not desirable to have a password (even crypted) in the profile. The file should have the user (default is root) and then password in colon delimited fields (like /etc/shadow).

While doing this, setting the serial and terminal settings has been refactored to use CAF::FileEditor.


ncm-grub/src/main/perl/grub.pm
```diff
@@ -16,6 +16,7 @@ package NCM::Component::grub;
 #
 
 use strict;
```
stdweird: can you add `use warnings;` or better yet replace everything from here to the top with
`#${PMpre} NCM::Component::grub${PMpost}` and let the templating do its magic

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +76,70 @@ sub grubbyArgsOptions {
 
 }
 
+sub Configure_password {
+    my ($self, $config, $editor) = @_;
+
+    my $password = undef;
+    my $passwordpath = $self->prefix . "/password";
```
stdweird: it is much better to use `$config->getTree($self->prefix);` (or `$config->getElement($self->prefix())->getTree()` for old CCM versions, and you can use the element and data as a hashref (so in Configure you could do
```perl
my $tree = $config->getTree($self->prefix);
...
$self->Configure_password($tree->{password}, $editor) if $tree->{password};
```
and then you can use the passed password tree as regular hashref
in particular, `$config->getElement("$passwordpath/enabled")->getValue()` below returns the stirng 'false' for the boolean `false`, so it is always true. with getTree, booleans are converted in 0/1 so they can be used as booleans.

ajf8:
ping, hopefully addressed the points in feedback.

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -7,6 +7,29 @@ declaration template components/grub/schema;
...
+
+type type_grub_password = {
+    "option"        : string with match (SELF, "^(md5|encrypted)$")
+    "password"      ? string
+    "enabled"       ? boolean
+    "file"          ? string
+    "file_user"     : string = "root"
+};
+
+function validate_grub_password = {
```
stdweird: if you do not reuse this function, just add the verification code to the type as follows
```pan
type type_grub_password = {
  ....
} with {
  # function code
}
```

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -7,6 +7,29 @@ declaration template components/grub/schema;
...
+    if (is_defined(SELF["enabled"]) && SELF["enabled"]) {
+        if (!is_defined(SELF["file"]) && !is_defined(SELF["password"])) {
```
stdweird: no need for a new `if` blok, just continue with more `&&` like
```pan
if (is_defined(SELF["enabled"]) && SELF["enabled"] &&
    !is_defined(SELF["file"]) && !is_defined(SELF["password"])) {
    error(...)
}
```

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -7,6 +7,29 @@ declaration template components/grub/schema;
 
 include {'quattor/schema'};
 
+# the crypted password can be supplied either in the password field
```
stdweird: wrap this in annotation braclets like
```pan
@{....
}
```

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -7,6 +7,29 @@ declaration template components/grub/schema;
 
 include {'quattor/schema'};
```
stdweird: can you check if the schema is part of a unittest somewhere?

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+
+    # if passwords have not been explicitly enabled or disabled then do nothing.
+    return unless ($config->elementExists($passwordpath));
+    my $tree = $config->getElement($passwordpath)->getTree();
+
+    if (!defined($tree->{enabled})) {
+        return;
```
stdweird: add a verbose message that password section was defined, but neither enabled nor disabled

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
 
 }
 
+sub Configure_password {
```
stdweird: add comment what this method does. there's also no need for a `Configure_` prefix

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+    my ($self, $config, $editor) = @_;
+
+    my $password = undef;
```
stdweird: no need for `= undef`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+    my $passwordpath = $self->prefix . "/password";
+
+    # if passwords have not been explicitly enabled or disabled then do nothing.
+    return unless ($config->elementExists($passwordpath));
+    my $tree = $config->getElement($passwordpath)->getTree();
+
+    if (!defined($tree->{enabled})) {
+        return;
+    } elsif (!$tree->{enabled}) {
+        $editor->remove_lines(qr/^password\s+/, '');
```
stdweird: add an info message that password was disabled and (possibly) removed

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+
+    if (my $passwordfile = $tree->{file}) {
+        my $fileuser = $tree->{file_user};
+        if (! -R $passwordfile) {
+            $self->error("grub password file $passwordfile does not exist or is not readable.");
+            return;
+        }
+        my $pf_fh = CAF::FileEditor->new($passwordfile, log => $self);
+        $pf_fh->cancel();
+        for my $line (split(/\n/, "$pf_fh")) {
```
stdweird: `for` -> `foreach` (it doens't actaully do anything different, but we use `foreach` in all the other code)

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+            chomp $line;
+            my @fields = split(/:/, $line);
```
stdweird: i'm going to assume you/someone "invented" this format? next time, please use `.ini` or something that already exists, so you can use an existing parser like `Config::Tiny`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+            if ($fields[0] eq $fileuser) {
+                $password = $fields[1];
```
stdweird: this will set the wrong password if the password itslef contains `:`.
also, add unittests for this

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
...
+        }
+        if (!defined($password)) {
+            $self->error("unable to find user $fileuser in grub password file $passwordfile");
+            return;
+        }
+    } else {
+        $password = $tree->{password};
+    }
+
+    my $val = "";
```
stdweird: use 
```perl
my $val = $tree->{option} ? "--$tree->{option} " : "";
```

ncm-grub/src/main/perl/grub.pm
```diff
@@ -97,14 +160,19 @@ sub Configure {
...
-  my $newterminal = "";
-  if ($config->elementExists($consolepath)) {
-    my %consnode = $config->getElement($consolepath)->getHash();
+  my $editor = CAF::FileEditor->new($GRUB_CONF,
+                                owner => "root",
+                                group => "root",
+                                mode => 0400,
+                                log => $self);
 
+  if (length("$editor") == 0) {
```
stdweird:
```perl
if(! "$editor") {
```

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,6 +77,66 @@ sub grubbyArgsOptions {
 
 }
 
+sub Configure_password {
+    my ($self, $config, $editor) = @_;
```
stdweird: use better variable name for `editor`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -97,14 +160,19 @@ sub Configure {
...
   }
 
-  my $cons = undef;
-  my $consolepath = "/hardware/console/serial";
-  my $serialcons  = undef;
-  my $newserial = "";
-  my $newterminal = "";
-  if ($config->elementExists($consolepath)) {
-    my %consnode = $config->getElement($consolepath)->getHash();
+  my $editor = CAF::FileEditor->new($GRUB_CONF,
```
stdweird: use better variable name for editor, like `$grub_fh`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -97,14 +160,19 @@ sub Configure {
...
+                                owner => "root",
+                                group => "root",
+                                mode => 0400,
+                                log => $self);
 
+  if (length("$editor") == 0) {
+      print $editor "# Generated by ncm-grub\n";
+  }
+  
+  my $cons = undef;
```
stdweird: no need for `=undef`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -86,8 +149,8 @@ sub Configure {
   }
   my $kernelname='vmlinuz';
```
stdweird: spaces around `=`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -97,14 +160,19 @@ sub Configure {
...
+                                mode => 0400,
+                                log => $self);
 
+  if (length("$editor") == 0) {
+      print $editor "# Generated by ncm-grub\n";
+  }
+  
+  my $cons = undef;
+  if ($config->elementExists($PATH_CONSOLE_SERIAL)) {
+    my %consnode = $config->getElement($PATH_CONSOLE_SERIAL)->getHash();
```
stdweird: 4 space indentation

ncm-grub/src/main/perl/grub.pm
ttyS4: This could help here:
`split(/:/, $line,2);`

ncm-grub/src/main/perl/grub.pm
```diff
@@ -97,14 +160,19 @@ sub Configure {
...
-  my $serialcons  = undef;
-  my $newserial = "";
-  my $newterminal = "";
-  if ($config->elementExists($consolepath)) {
-    my %consnode = $config->getElement($consolepath)->getHash();
+  my $editor = CAF::FileEditor->new($GRUB_CONF,
+                                owner => "root",
+                                group => "root",
+                                mode => 0400,
+                                log => $self);
```
stdweird: this is potentially dangerous, since this file might contain plain text password. not sure how to deal with this. i'd say that when no password is removed/added/modified (ie password configuration is missing), you could leave it.

stdweird:
@ajf8 ping?

ned21:
He's back on Monday.

jrha:
@ajf8 any update?

ajf8:
Sorry for the delay, I was on holiday, but hopefully these points are addressed now.

jrha:
Thanks, LGTM

ncm-grub/src/main/perl/grub.pm
```diff
@@ -74,534 +77,543 @@ sub grubbyArgsOptions {
...
     }
 
-    my $parity = "n";
-    if (exists $consnode{"parity"}) {
-      $parity = $consnode{"parity"}->getValue();
+    my $grub_fh = CAF::FileEditor->new( $GRUB_CONF,
+        owner => "root",
+        group => "root",
+        mode  => 0400,
+        log   => $self );
```
stdweird: please be fully aware that you will log any changes to the grub config file (when level is high enough), incl the password (hash).
this "feature" should be at least mentioned in the pod

PR #786, opened 2016-06-07
Resolver list lengths

### LINKED ISSUE #785: ncm-resolver/ncm-named: Lengths of lists should be validated
Both ncm-resolver and ncm-named can configure a list of name-servers in `resolv.conf` (although we agreed in #434 to deprecate this feature of ncm-named).

The man pages and source code for the GNU resolver state that:
> Up to MAXNS (**currently 3**, see &lt;resolv.h&gt;) name servers may be listed, one  per  keyword.
```c
# define MAXNS	3	/* max # name servers we'll track */
```

The documentation for Solaris agrees:
> Succeeding lines list the IP addresses of up to three DNS master, secondary, or cache-only name servers that the resolver should consult to resolve queries. (Do not list **more than three** primary or secondary servers.)

Therefore I propose limiting the length of the `servers` property to 3.

jrha:
All supported resolvers (GNU/glib/BIND and Solaris) support a maximum of:
* Three nameservers
* Six search domains with a total of 256 characters

This also switches the plain strings to network types.

Small chance of backwards incompatibility.

Fixes #785.

### DIFF
--- a/ncm-named/src/main/pan/components/named/schema.pan
+++ b/ncm-named/src/main/pan/components/named/schema.pan
@@ -17,11 +17,11 @@ function component_named_valid = {
   if ( ARGC != 1 ) {
     error(function_name+': this function requires 1 argument');
   };
-  
+
   if ( exists(SELF['serverConfig']) && exists(SELF['configfile']) ) {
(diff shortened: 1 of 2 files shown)


jouvin:
LGTM.

PR #790, opened 2016-06-14
ncm-metaconfig: add rsync template for xinetd

wdpypere:
add support for rsync in xinet.d and add some basic tests.


stdweird:
@wdpypere tested? LGTM otherwise

wdpypere:
@stdweird yes, running on one of our systems.

PR #798, opened 2016-07-04
ncm-spma: Backport changes from reverting yumng schema

### LINKED ISSUE #797: ncm-spma: schema and config conflict
Merging #761 has broken the default spma configuration, it looks like this was spotted once during the review, but missed when the PR was merged.

The schema defines:
```pan
type SOFTWARE_GROUP = {
    "default" : boolean = true
    "mandatory" : boolean = true
    "optional" : boolean = false
};
```

```pan
bind "/software/groups" = SOFTWARE_GROUP {};
```

`config-rpm` and `config-ips` both then effectively do:
```pan
'/software/groups/names' ?= list();
```
Changing the `list()` to `dict()` passes validation, but what was the intention here?

jrha:
Fixes #797.

`yumng` properties have no business being anywhere near `ips` either.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/config-ips.pan
+++ b/ncm-spma/src/main/pan/components/spma/config-ips.pan
@@ -5,8 +5,6 @@
 
 unique template components/${project.artifactId}/config-ips;
 
-'/software/groups/names' ?= list();
-
 # Set prefix to root of component configuration.
 prefix '/software/components/${project.artifactId}';
 
@@ -20,4 +18,3 @@ prefix '/software/components/${project.artifactId}';
                          '/software/requests',
                          '/software/uninstall');
 'flagfile' = '/var/tmp/spma-run-flag';
-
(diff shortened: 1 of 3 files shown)


jrha:
Tested with aquilon at RAL.

ned21:
Sorry for the problem. :( Given that anyone attempting to follow history here is going to be very confused, is it worth being a bit more detailed in exactly what the problem was and why this fixes it? From what I can tell, 1fba420 fixed it, but that fix failed to make it across when schema.pan was refactored into software.pan?

jouvin:
This is a side issue but I have a strange feeling with both `spma/ips/schema` and `spma/rpm/schema` being unconditionally included in `spma/schema`. Why they are not condtionally included bease on `CONFIG_MODULES_CONF_SUPFFIX` variable (the name may need to be updated)? Is it worth another issue?

jrha:
@jouvin see #716

jouvin:
Sorry I didn't look at the linked PR and missed the change! Perfect!

jouvin:
Tested successfully at LAL

PR #799, opened 2016-07-05
ncm-modprobe: overwrite initramfs if it exists, instead of initrd.

ajf8:
RHEL6 uses initramfs in /boot/initramfs-* instead of initrd in
/boot/initrd-*. It provides a mkinitrd compatability wrapper which calls
dracut, but this module outputs them to the wrong place, leading to a
duplicate which isn't actually used.

### DIFF
--- a/ncm-modprobe/src/main/perl/modprobe.pm
+++ b/ncm-modprobe/src/main/perl/modprobe.pm
@@ -20,8 +20,10 @@ use EDG::WP4::CCM::Configuration;
 use CAF::Process;
 use CAF::FileWriter;
 use LC::File qw(directory_contents);
+use Readonly;
 
 no strict 'refs';
+Readonly my $DEFAULT_MODPROBE_CONF => "/etc/modprobe.d/quattor.conf";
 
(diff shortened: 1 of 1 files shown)


ncm-modprobe/src/main/perl/modprobe.pm
```diff
@@ -57,7 +58,9 @@ sub mkinitrd
     foreach my $i (@$dir) {
         if ($i =~ m{^System\.map\-(2\.6\..*)$}) {
```
stdweird: not relevant for kernels more recent than 2.6? or is this code that means older than 2.4?

ncm-modprobe/src/main/perl/modprobe.pm
```diff
@@ -22,6 +22,7 @@ use CAF::FileWriter;
 use LC::File qw(directory_contents);
 
 no strict 'refs';
+use constant DEFAULT_MODPROBE_CONF => "/etc/modprobe.d/quattor.conf";
```
ned21: Since there are no existing constants, can we use Readonly here instead?

jrha:
LGTM

PR #804, opened 2016-07-07
ncm-spma: spma-run: Lock use reporter instance for reporting

stdweird:
Depends on https://github.com/quattor/CAF/pull/173


jouvin:
@stdweird looks as a trivial change, why assign it to 16.8. Seems as something that should help debugging, is not it?

jrha:
It is trivial, but sometimes trivial changes have big consequences, but LGTM.

jouvin:
:-) I wanted to say that I had the feeling it should be easy to validate that it works!

PR #807, opened 2016-07-11
config-rpm: do not define groups/names/yumng by default

### LINKED ISSUE #806: ncm-spma 16.6.0-rc3: groups/names/yumng should not be defined by default
Unfortunately #798 still contains a `yumng` related definition, `/software/components/spma/groups/names/yumng` which is interpreted as a group name by `yum`. This causes an error. 

This is a showstopper for [16.6.0-rc3](https://github.com/quattor/release/issues/250)

jouvin:
Fixes #806.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/config-rpm.pan
+++ b/ncm-spma/src/main/pan/components/spma/config-rpm.pan
@@ -7,7 +7,6 @@ unique template components/spma/config-rpm;
 # Prefix for packages/groups
 prefix '/software';
 'groups' ?= dict();
-'groups/yumng/names' ?= list();
 # Package to install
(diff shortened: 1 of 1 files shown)


stdweird:
@ned21 this might affect MS

jrha:
Perhaps this should be renamed to `config-yum` and copy made for `config-yumng`?

ned21:
yumng relies on that path in the schema so it means yumng is not usable in this release. I think that's fine (let's label yumng as a "tech preview" in the release notes so only adventurous souls that can modify a schema try to use it), but Jindrich is going to have to think of another way to unwind our problem. :(

jrha:
I think I may have a solution in #716.

PR #808, opened 2016-07-11
ncm-postfix: use CCM::TextRender

stdweird:
Do not rely on legacy `NCM::Component` template method, use `CCM::TextRender`


ned21:
LGTM

Issue #370, opened 2014-11-05
preferred component pan structure


stdweird:
The current pan code (`config.pan`, `schema.pan` and so forth) is not entirely consistent throught the components.

What is the preferred structure to have
* `config.pan` and `schema.pan` (eg https://github.com/quattor/configuration-modules-core/tree/master/ncm-libvirtd/src/main/pan/components/libvirtd)
* or all of `config.pan`, `schema.pan` `config-rpm.pan` and `config-common.pan`
* and do we still need `config-xml.pan` (most of these have the `code` attribute set)

Other questions: 
* do we need the license in the pan templates
* is panc recent enough to support
 * `prefix` with absolute path
 * `include 'x'` instead of `include {'x'}`

jouvin:
We can do everything! I am not completely sure about what is your question? `schema.pan` by convention is where we put the `bind` statement. So if you want to build the schema from some common parts, I'd suggest to put this part into another template like `schema-common.pan` and include this template in `schema.pan` rather than using `schema.pan` for the common part and put the `bind` statement in `config.pan` for example.

PR #810, opened 2016-07-13
ncm-nss: replace LC::Check with FileWriter

ned21:
Use FileWriter instead of LC::Check and add a simple unit test of the
behaviour.


ncm-nss/src/test/resources/simple.pan
```diff
@@ -0,0 +1,15 @@
+object template simple;
+
+#function pkg_repl = { null; };
```
stdweird: remove if unused

ncm-nss/src/test/perl/configure.t
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/perl
+# -*- mode: cperl -*-
+use strict;
+use warnings;
+
+use Test::More;
+use Test::Quattor qw(simple);
+use NCM::Component::sysctl;
```
stdweird: probably not what you want (current works because `Test::MockModule` below takes care of loading nss module)

ned21: Oops - thanks!

ncm-nss/src/test/resources/simple.pan
ned21: It would be nice to test the schema too by including config.pan but I can't remember why this didn't work now - I wrote the code too long ago. :( What's recommended best practice here?

stdweird: remove these comments to start with :smile: 
and change https://github.com/quattor/configuration-modules-core/blob/master/ncm-nss/src/main/pan/components/nss/schema.pan#L11 in https://github.com/quattor/configuration-modules-core/blob/master/ncm-autofs/src/main/pan/components/autofs/schema.pan#L7
(and include any other types needed in the schema, but from quick glance, none are required)

ncm-nss/src/test/perl/configure.t
stdweird: and because you don't do anything with mock below, you can maybe remove those and use the correct one here?

ncm-nss/src/main/perl/nss.pm
```diff
@@ -95,16 +95,16 @@ sub Configure {
...
-                                 contents => $contents,
-                                 owner => "root",
-                                 group => "root",
-                                 mode => "0444",
-                                );
+    my $fh = CAF::FileWriter->open("/etc/nsswitch.conf",
+                                   backup => ".OLD",
+                                   owner => "root",
+                                   group => "root",
+                                   mode => "0444",
```
stdweird: maybe remove the `""` (i just checked the LC code https://github.com/quattor/LC/blob/master/src/main/perl/Check.pm#L332, this actually works with quotes, but it's confusing; but then again, this is LC...)

ncm-nss/src/main/perl/nss.pm
```diff
@@ -95,16 +95,16 @@ sub Configure {
...
+                                   log => $self,
+        );
+    print $fh join("", @new);
+    my $result = $fh->close();
+
     if ($result) {
         $self->log("updated /etc/nsswitch.conf");
```
stdweird: use `info`? (log only writes to logfile)

ncm-nss/src/test/resources/simple.pan
ned21:
```
user-defined validation failed
element path: '/software/components/nss/version'
element value: 16.2.1-SNAPSHOT
```

because the [component type](https://github.com/quattor/template-library-core/blob/master/quattor/types/component.pan#L13) only allows digits in this field.

stdweird: use `'version' ?= '${no-snapshot-version}';`

ncm-nss/src/main/perl/nss.pm
ned21: ? I get a failure in the unit test if I replace ```mode => "0444"``` with  ```mode => 0444```

ncm-nss/src/test/resources/simple.pan
```diff
@@ -0,0 +1,18 @@
+object template simple;
+
+# mock pkg_repl
+function pkg_repl = { null; };
+include 'components/nss/config';
+"/software/components/nss/dependencies/pre" = null;
+# a version of 16.2.1-SNAPSHOT fails validation so unset
```
stdweird: please remove again

ncm-nss/src/main/pan/components/nss/config.pan
```diff
@@ -9,9 +9,9 @@
 
 unique template components/${project.artifactId}/config;
 
-include { 'components/${project.artifactId}/schema' };
+include 'components/${project.artifactId}/schema';
 
-include { 'pan/functions' };
+include 'pan/functions';
 
 # Package to install.
```
stdweird: https://github.com/quattor/configuration-modules-core/issues/370#issuecomment-61801754

ncm-nss/src/main/perl/nss.pm
stdweird: i just pull this branch and both mvn test and mvn package work fine (and produces an rpm)

stdweird: (also when i use the non-string `0444`)

ned21: Thanks for checking. Something is weird with maven on this one machine. Using another also works. :(

stdweird: try with mvnprove.pl (which does not use maven at all)?

ncm-nss/src/main/perl/nss.pm
```diff
@@ -95,28 +95,28 @@ sub Configure {
...
+                                   group => "root",
+                                   mode => 0444,
+                                   log => $self,
+        );
+    print $fh join("", @new);
+    my $result = $fh->close();
+
     if ($result) {
-        $self->log("updated /etc/nsswitch.conf");
+        $self->info("updated /etc/nsswitch.conf") unless $NoAction;
```
stdweird: this looks like some idiom to add to CAF, but also related to https://github.com/quattor/CAF/issues/100.
in any case, most components do not make this distinction (and it should be handeled on e.g. CAF level).

anyway, what is interesting is that the tests pass without any definition of `$NoAction` (i expected `use strict` or `use warning` to at least make some noise)

stdweird: wrt `NoAction`, in the old components code, we extend the `NCM::Component` namespace by diretc manipulation of `ISA`, and the `$NoAction` variable is the one set here: https://github.com/quattor/ncm-ncd/blob/master/src/main/perl/Component.pm#L15
if you would switch to the `use parent qw(NCM::Component)`, this would fail, and you would have to use `$CAF::Object::NoAction`. as the `NoAction` is read most of the time, this is ok.

ned21:
Anything to be done here to get this into 16.8?

ncm-nss/src/main/pan/components/nss/config.pan
```diff
@@ -9,15 +9,15 @@
...
 "/software/packages" = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
 
 # standard component settings
-"/software/components/${project.artifactId}/active"   ?=  true ;
-"/software/components/${project.artifactId}/dispatch" ?=  false ;
-"/software/components/${project.artifactId}/version"   = "${project.version}";
+"/software/components/${project.artifactId}/active"   ?= true ;
```
stdweird: copy paste the code in the comment referenced above, e.g. to use `prefix` here

stdweird:
small remark wrt to the pan template cleanup. can you also open an issue in `CAF` to provide something generic to handle noaction-releated reporting like the `$self->info("updated /etc/nsswitch.conf") unless $NoAction;`
is ready to merge otherwise

ned21:
config.pan updated and https://github.com/quattor/CAF/issues/185 opened.

stdweird:
@ned21 thx
LGTM

PR #813, opened 2016-07-13
ncm-spma: handle group names containing spaces

### LINKED ISSUE #812: ncm-spma doesn't handle properly group names with embedded spaces
A YUM group name can be made of multiple words (a string with spaces). This is not handled properly by `ncm-spma`. In addition, if the group name is incorrect (the group doesn't exist), the error logged is difficult to interpret (no explicit error logged).

jouvin:
- Also improve logging in case of error during the group expansion
- Fixes #812

### DIFF
--- a/ncm-spma/src/main/perl/spma/yum.pm
+++ b/ncm-spma/src/main/perl/spma/yum.pm
@@ -363,11 +363,19 @@ sub expand_groups
 
     my $pkgs = Set::Scalar->new();
 
-    while (my ($group, $types) = each(%$groups)) {
+    while (my ($group_e, $types) = each(%$groups)) {
+        # A group can contain spaces and thus is escaped
+        my $group = unescape($group_e);
+        $self->debug(2, "Getting packages for group '$group'");
         my $what = join(",", grep($types->{$_}, keys(%$types)));
         my $lst = $self->execute_yum_command([REPOGROUP, $what, $group],
-                                             "Group expansion", 1)
-            or return undef;
+                                             "Group expansion", 1);
+        # If a group doesn't exist, the repoquery command returns an empty string
+        # with a success exit code.
+        unless ( $lst ) {
+          $self->error("Error during group '$group' expansion (group probably doesn't exist)");
+          return undef;
+        }
         $pkgs->insert(split(/\n/, $lst));
     }
     return $pkgs;
--- a/ncm-spma/src/main/perl/spma/yum.pm
+++ b/ncm-spma/src/main/perl/spma/yum.pm
@@ -363,10 +363,11 @@ sub expand_groups
 
     my $pkgs = Set::Scalar->new();
 
-    while (my ($group_e, $types) = each(%$groups)) {
+    foreach my $group_e (sort(keys(%$groups))) {
         # A group can contain spaces and thus is escaped
         my $group = unescape($group_e);
         $self->debug(2, "Getting packages for group '$group'");
+        my $types = $groups->{$group_e};
         my $what = join(",", grep($types->{$_}, keys(%$types)));
         my $lst = $self->execute_yum_command([REPOGROUP, $what, $group],
                                              "Group expansion", 1);


ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -363,11 +363,20 @@ sub expand_groups
 
     my $pkgs = Set::Scalar->new();
 
-    while (my ($group, $types) = each(%$groups)) {
+    while (my ($group_e, $types) = each(%$groups)) {
+        # A group can contain spaces and thus is escaped
```
stdweird: spaces are no reason to escape keys afaik.

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -363,11 +363,20 @@ sub expand_groups
...
-                                             "Group expansion", 1)
-            or return undef;
+        # repoquery command requires that each word in a group name be a separate parameter
+        my $lst = $self->execute_yum_command([REPOGROUP, $what, split(/\s/,$group)],
+                                             "Group expansion", 1);
+        # If a group doesn't exist, the repoquery command returns an empty string
+        # with a success exit code.
+        unless ( $lst ) {
+          $self->error("Error during group '$group' expansion (group probably doesn't exist)");
+          return undef;
```
stdweird: return in the middle of an each loop is evil https://github.com/quattor/CAF/issues/109, switch to foreach loop (and sort the keys for reproducibility).

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -363,11 +363,20 @@ sub expand_groups
...
-    while (my ($group, $types) = each(%$groups)) {
+    while (my ($group_e, $types) = each(%$groups)) {
+        # A group can contain spaces and thus is escaped
+        my $group = unescape($group_e);
+        $self->debug(2, "Getting packages for group '$group'");
         my $what = join(",", grep($types->{$_}, keys(%$types)));
-        my $lst = $self->execute_yum_command([REPOGROUP, $what, $group],
-                                             "Group expansion", 1)
-            or return undef;
+        # repoquery command requires that each word in a group name be a separate parameter
```
stdweird: what? aaargh...... :scream:

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -363,11 +363,20 @@ sub expand_groups
 
     my $pkgs = Set::Scalar->new();
 
-    while (my ($group, $types) = each(%$groups)) {
+    while (my ($group_e, $types) = each(%$groups)) {
+        # A group can contain spaces and thus is escaped
+        my $group = unescape($group_e);
+        $self->debug(2, "Getting packages for group '$group'");
         my $what = join(",", grep($types->{$_}, keys(%$types)));
```
stdweird: sort the keys

jouvin:
Looking in more details, I think the fix is incorrect. My problem was that `yum grouplist` lists group whose names seems to contain space (like `GNOME Desktop`) but the real name of the group is `gnome`. So splitting the name in several parameter just allows `GNOME` to mach and `Desktop` is just ignored... I'll redo the PR just to add more verbosity in case of errors.

ncm-spma/src/main/perl/spma/yum.pm
jouvin: I'm afraid they are... At least this is my experience! panc considers a key with an embedded space as invalid... I think that internally panc has no problem with spaces in keys but that the restriction was maintained until we removed the support for all the backends which cannot deal with them.

ncm-spma/src/main/perl/spma/yum.pm
jouvin: I'll do it. This is the original code from Luis...

ncm-spma/src/main/perl/spma/yum.pm
jouvin: A mistake of mine, See https://github.com/quattor/configuration-modules-core/pull/813#issuecomment-232466766.

ncm-spma/src/main/perl/spma/yum.pm
jouvin: Fixed. Thanks for spotting this.

jouvin:
The PR should be ok now. It really implements support for group names with embedded spaces as I found one that I needed: `X Window System`.

stdweird:
@jouvin could you add a unittest for this case (or add a group with spaces in one of the existing unittests)

jouvin:
Sure... even if I have to check what already exists in term of coverage for the group expansion as it requires to execute `repoquery`. At least we can unit test the processing of an empty list returned by `repoquery`

jrha:
@jouvin there really are no groups with spaces in their IDs, the problem is that yum will allow you to use the group name (description) as an identifier.

For example the group `x11` can also be referred to by the name `X Window System`, which I think is a really bad idea.

jrha:
Just tested, the international names work as well. I really think it would be simpler for us to just insist on the ID and not worry about spaces.

jouvin:
@jrha this is what I understood in my initial tests. Just I didn't find how to get the ID rather than the descriptive name. Anyway, I think this is harmless to support also the descriptive name. I can't think of any potential drawback: 
- a non-escaped name that contains the escaped character sequence is impossible I guess (would be a very strange name and I can't imagine RH producing such a name). So unescaping a non-escaped name is harmless.
- If somebody adds both the ID and the descriptive name to the configuration, this will be 2 entries under `/software/groups` but the group expansion will not add packages (provided that the same mandatory/default/optional flag is used but the result would be the same as modifying them... except if somebody set to false an attribute first set to true).

jrha:
`curl -s http://YOUR_MIRROR_SERVER/sl-7x-x86_64/RPMS.os/repodata/comps.xml | grep '<id>\|<name'`

jouvin:
... which is not exactly what I call an easy command! For me, one more reason to add the descriptive name returned by `yum grouplist` or that you can find in some doc (like  https://www.centos.org/forums/viewtopic.php?t=47825).

jrha:
Well that was just my way of finding the international names, `yum grouplist -v` shows the IDs.

jrha:
I'd like this to be discussed more before accepting it, so I'm pushing it back until you return from holiday.

piojo:
Unit test plz?

ned21:
Did we agree at the workshop that since groups can lead to unpredictable and undesirable behaviour that we want to move away from using them, and ultimately drop support for them from spma? Is this fix still needed?

jrha:
We did indeed! So no, I don't think this is needed.

PR #816, opened 2016-07-15
ncm-afsclt: fix handling of size AUTOMATIC

### LINKED ISSUE #815: ncm-afsclt fails if cache size is set to AUTOMATIC
`nfs-afsclt`, since changes introduced in Dec. 2015 (ca8692b8c787da2dd80d21f1e174011c7831a13b), fails if the cache size is set to AUTOMATIC when parsing the existing `cacheinfo` file and when doing some sanity checks on the size.

Should be considered as a showstopper for [16.6.0-rc4](https://github.com/quattor/release/issues/253) (missed at LAL in 16.2 in fact as we didn't deploy 16.2).

jouvin:
Fixes #815.

### DIFF
--- a/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
+++ b/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
@@ -5,7 +5,7 @@
 
 declaration template components/afsclt/schema;
 
-include 'quattor/schema';
+include 'quattor/types/component';
 
 type component_afsclt_entry = {
    include structure_component
--- a/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
+++ b/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
@@ -17,8 +17,7 @@ type component_afsclt_entry = {
    "afs_mount"  ? string         # AFS mount point (e.g. /afs)
    "cachemount" ? string         # AFS cache location (/usr/vice/etc/cache)
    "cachesize"  ? string         # AFS cache size in kB
-   "options"    ? nlist          # AFS client options
-   "enabled"    : string with match (SELF, 'yes|no') # Shall AFS client be active ?
+   "enabled"    : legacy_binary_affirmation_string  # Shall AFS client be active ?
    "afsd_args"  ? string{}       # /etc/afsd.args values for rc.afs
 };
 
--- a/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
+++ b/ncm-afsclt/src/main/pan/components/afsclt/schema.pan
@@ -10,16 +10,16 @@ include 'quattor/schema';
 type component_afsclt_entry = {
    include structure_component
    
-   "thiscell"   : string         # "AFS home cell"
-   "thesecells" ? string[]       # "Cell list to authenticate to"
-   "settime"    ? boolean        # "Shall AFS client sync sys time?"
-   "cellservdb" ? string         # "Where Master CellServDB can be found"
-   "cachemount" ? string         # "AFS cache location (/usr/vice/etc/cache)"
-   "cachesize"  ? string         # "AFS cache size in kB"
-   "options"    ? nlist          # "AFS client options"
-   "enabled"    : string with match (SELF, 'yes|no') 
-   				 # "Shall AFS client be active ?"
-   "afsd_args"  ? string{} # /etc/afsd.args values for rc.afs
+   "thiscell"   : string         # AFS home cell
+   "thesecells" ? string[]       # Cell list to authenticate to
+   "settime"    ? boolean        # Shall AFS client sync sys time?
+   "cellservdb" ? string         # Where Master CellServDB can be found
+   "afs_mount"  ? string         # AFS mount point (e.g. /afs)
+   "cachemount" ? string         # AFS cache location (/usr/vice/etc/cache)
+   "cachesize"  ? string         # AFS cache size in kB
+   "options"    ? nlist          # AFS client options
+   "enabled"    : string with match (SELF, 'yes|no') # Shall AFS client be active ?
+   "afsd_args"  ? string{}       # /etc/afsd.args values for rc.afs
 };
 
 bind "/software/components/afsclt" = component_afsclt_entry;
--- a/ncm-afsclt/src/main/perl/afsclt.pm
+++ b/ncm-afsclt/src/main/perl/afsclt.pm
@@ -48,10 +48,10 @@ sub Configure_Cell {
     }
 
     my $thiscell_fh = CAF::FileWriter->new( $THISCELL, log => $self );
-    print $thiscell_fh $config->{thiscell}."\n";
+    print $thiscell_fh "$config->{thiscell}\n";
 
     if ( $thiscell_fh->close() ) {
-        $self->info("Updated thiscell to ".$config->{thiscell});
+        $self->info("Updated thiscell to $config->{thiscell}");
     }
 
     return 0;
@@ -292,7 +292,7 @@ sub Configure_Afsd_Args {
     if ( defined($config->{afsd_args}) ) {
         my $fh = CAF::FileWriter->new( $AFSD_ARGS, log => $self, backup => ".old" );
         foreach my $key (sort keys %{$config->{afsd_args}}) {
-            print $fh "$key:" . $config->{afsd_args}->{$key} . "\n";
+            print $fh "$key:$config->{afsd_args}->{$key}\n";
         }
         if ( $fh->close() ) {
             $self->info("Updated afsd.args");
--- a/ncm-afsclt/src/main/perl/afsclt.pm
+++ b/ncm-afsclt/src/main/perl/afsclt.pm
@@ -206,9 +206,8 @@ sub Configure_CellServDB {
     if ($cellservdb_fh->close()) {
         $self->info("Updated CellServDB");
     }
-    $self->update_afs_cells();
 
-    return 1;
+    return $self->update_afs_cells();
 }
 
 # update the list of known AFS cells and run "fs newcell" when needed
--- a/ncm-afsclt/src/main/perl/afsclt.pm
+++ b/ncm-afsclt/src/main/perl/afsclt.pm
@@ -106,9 +106,8 @@ sub Configure_Cache {
         $self->warn("Cannot determine current AFS cache size, changing only config file");
     }
 
-    my $afs_cacheinfo_fh = CAF::FileEditor->new( $AFS_CACHEINFO, log => $self ); # TODO use FileReader
-    $afs_cacheinfo_fh->cancel();
-    if ( "$afs_cacheinfo_fh" =~ m;^([^:]+):([^:]+):(\d+)$; ) {
+    my $afs_cacheinfo_fh = CAF::FileReader->new( $AFS_CACHEINFO, log => $self );
+    if ( "$afs_cacheinfo_fh" =~ m;^([^:]+):([^:]+):(\d+|AUTOMATIC)$; ) {
         $file_afsmount   = $1;
         $file_cachemount = $2;
         $file_cache      = $3;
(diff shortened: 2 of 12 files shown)


jrha:
That and many sites have stopped using AFS.

jouvin:
Unit tests added for test configurations. Uncovered a few potential issues, in particular if the initial `cacheinfo` file was empty.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -106,34 +111,36 @@ sub Configure_Cache {
         $self->warn("Cannot determine current AFS cache size, changing only config file");
     }
 
-    my $afs_cacheinfo_fh = CAF::FileEditor->new( $AFS_CACHEINFO, log => $self ); # TODO use FileReader
-    $afs_cacheinfo_fh->cancel();
-    if ( "$afs_cacheinfo_fh" =~ m;^([^:]+):([^:]+):(\d+)$; ) {
+    my $afs_cacheinfo_fh = CAF::FileReader->new( $AFS_CACHEINFO, log => $self );
+    $self->debug(2, "$AFS_CACHEINFO current contents: >>>$afs_cacheinfo_fh<<<");
```
stdweird: only in else block? (and in that case, maybe in the error message)

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -106,34 +111,36 @@ sub Configure_Cache {
...
-    # sanity check - don't allow cachesize bigger than 95% of a partition size
-    $proc = CAF::Process->new( [ "df", "-k", $file_cachemount ], log => $self, keeps_state => 1 );
-    foreach my $line ( split ( "\n", $proc->output() ) ) {
-        if ($line =~ m{^.*?\s+(\d+)\s+\d+\s+\d+\s+\d+%\s+(.*)}) {
-            my $disk_cachesize = $1;
-            my $mount          = $2;
-            if ( $mount eq $file_cachemount && $new_cache > 0.95 * $disk_cachesize ) {
-                $self->error("Cache size ($disk_cachesize) on $mount cannot exceed 95% of its partition size. Not changing.");
-                return 1;
+    unless ( $new_cache eq "AUTOMATIC" ) {
```
stdweird: use `if ($new_cache ne "AUTOMATIC") {` ?

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -83,16 +83,21 @@ sub Configure_Cache {
     my $file_cache      = 0;     # how much cache is actually configured now in the config file
     my $file_cachemount = '';    # where should the cache be mounted per config file
     my $new_cache       = 0;     # in 1k blocks.
-    my $file_afsmount   = '';
+    my $file_afsmount   = '/afs';
```
stdweird: why is this needed?

ncm-afsclt/src/test/resources/automatic.pan
```diff
@@ -0,0 +1,9 @@
+object template automatic;
+
```
stdweird: it would also be nice to do some schema tests, something similar like https://github.com/quattor/configuration-modules-core/blob/master/ncm-autofs/src/test/resources/maps.pan#L6 (with the lines above an below).
you will also have to replace https://github.com/quattor/configuration-modules-core/blob/master/ncm-afsclt/src/main/pan/components/afsclt/schema.pan#L8 with https://github.com/quattor/configuration-modules-core/blob/master/ncm-autofs/src/main/pan/components/autofs/schema.pan#L7

ncm-afsclt/src/main/perl/afsclt.pm
stdweird: ok, i sort of understand, this is for the empty file case.
better to use add the default values as readonly on top the module, and do
```perl
use Readonly;
Readonly my $DEFAULT_AFSMOUNT => '/afs';

...
my $file_afsmount   = $DEFAULT_AFSMOUNT;
```

and probably similar for file_cachemount and new_cache.

```

**Comment by stdweird** [comment]
some minor remarks, LGTM otherwise

**Comment by jouvin** [comment]
@stdweird thanks as usual for the remarks. I'll address them tomorrow. Most of them in fact apply to the original code and its last modification in december. I tried to address an urgent problem but didn't attempt to fix all the possible ones. I'll also try to improve a little bit the unit tests.

File: ncm-afsclt/src/main/perl/afsclt.pm (hunk 1 above)
Comment (jouvin) [inline ncm-afsclt/src/main/perl/afsclt.pm]: In fact, I realized that my mod was wrong because if was no longer possible to override value in `cacheinfo` by the Quattor configuration. `cacheinfo` should only be used as default values. fixed.

File: ncm-afsclt/src/main/perl/afsclt.pm (hunk 3 above)
Comment (jouvin) [inline ncm-afsclt/src/main/perl/afsclt.pm]: Done.

**Comment by jouvin** [comment]
Should be ok now. Remarks addresssed. Unit tests improved. `cacheinfo` values used only as default values.

File: ncm-afsclt/src/test/resources/automatic.pan (hunk 4 above)
Comment (jouvin) [inline ncm-afsclt/src/test/resources/automatic.pan]: Done

**Comment by jouvin** [comment]
@ajf8 you may want to have a look and to check that this doesn't break anything for you, compared to the last modifications you committed in December 2015.

File: ncm-afsclt/src/main/pan/components/afsclt/schema.pan [hunk 5]
```diff
@@ -5,21 +5,20 @@
...
-   				 # "Shall AFS client be active ?"
-   "afsd_args"  ? string{} # /etc/afsd.args values for rc.afs
+   "thiscell"   : string         # AFS home cell
+   "thesecells" ? string[]       # Cell list to authenticate to
+   "settime"    ? boolean        # Shall AFS client sync sys time?
+   "cellservdb" ? string         # Where Master CellServDB can be found
+   "afs_mount"  ? string         # AFS mount point (e.g. /afs)
+   "cachemount" ? string         # AFS cache location (/usr/vice/etc/cache)
+   "cachesize"  ? string         # AFS cache size in kB
+   "enabled"    : string with match (SELF, 'yes|no') # Shall AFS client be active ?
```
stdweird: use `legacy_binary_affirmation_string` instead.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -85,14 +88,27 @@ sub Configure_Cache {
     my $new_cache       = 0;     # in 1k blocks.
     my $file_afsmount   = '';
 
-    if ( $config->elementExists("$PREFIX/cachesize") ) {
-        $new_cache = $config->getValue("$PREFIX/cachesize");    #new cache size
-    }
-    else {
-        $self->info("Cannot get CDB $PREFIX/cachesize - not setting cache size");
+    my $afs_config = $config->getElement($PREFIX)->getTree();
```
stdweird: even better would be `$config->getTree($self->prefix())`, but maybe keep it in mind for a next time :smile:

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -106,34 +122,40 @@ sub Configure_Cache {
...
 
-    my $afs_cacheinfo_fh = CAF::FileEditor->new( $AFS_CACHEINFO, log => $self ); # TODO use FileReader
-    $afs_cacheinfo_fh->cancel();
-    if ( "$afs_cacheinfo_fh" =~ m;^([^:]+):([^:]+):(\d+)$; ) {
-        $file_afsmount   = $1;
-        $file_cachemount = $2;
+    my $afs_cacheinfo_fh = CAF::FileReader->new( $AFS_CACHEINFO, log => $self );
+    $self->debug(2, "$AFS_CACHEINFO current contents: >>>$afs_cacheinfo_fh<<<");
+    if ( "$afs_cacheinfo_fh" =~ m;^([^:]+):([^:]+):(\d+|AUTOMATIC)$; ) {
+        $file_afsmount   = $1 if $file_afsmount eq "";
```
stdweird: these cannot be emtpy, since the regex have `+`; same below

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -106,34 +122,40 @@ sub Configure_Cache {
...
+        $file_cachemount = $2 if $file_cachemount eq "";
         $file_cache      = $3;
-    }
-    else {
+    } elsif ( "$afs_cacheinfo_fh" ne "" ) {
         $self->error("Cannot parse stored AFS cache mount or size from $AFS_CACHEINFO");
```
stdweird: add the unparseable content in the message?

stdweird:
@jouvin thanks for the further cleanup.

ncm-afsclt/src/main/perl/afsclt.pm
jouvin: I'm working on it, not sure if it will make it in 16.6... I first want to add unit test before making the changes everywhere in the code.

ncm-afsclt/src/main/perl/afsclt.pm
jouvin: I am not sure to understand... You probably misread... The point is that you use the value from the file only if it is not defined in the configuration (in which case `$file_afs_mount` is not the default empty string).

ncm-afsclt/src/main/perl/afsclt.pm
jouvin: Not sure it adds anything... It is enough to look at fhe `cacheinfo` file contents.

ncm-afsclt/src/main/perl/afsclt.pm
stdweird: indeed, i misread.

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -106,34 +123,40 @@ sub Configure_Cache {
...
     }
 
     # adjust stored cache size (gets overwritten on restart for OpenAFS-1.4)
-    if ( $new_cache != $file_cache ) {
+    # Force string interpolation of cache size as it can be AUTOMATIC
+    if ( $file_afsmount eq "" ) {
+        $self->debug(1, "AFS mount point not defined: using default ($AFS_MOUNTPOINT_DEF)");
+        $file_afsmount = $AFS_MOUNTPOINT_DEF;
+    }
+    if ( "$new_cache" ne "$file_cache" ) {
```
stdweird: with this check, the `AFS_CACHEINIFO` is only updated is `file_cache` is modified, nit `file_afsmount` or `file_cache_mount`
i would always write the file, and with the `close` returns a change in content, do the block below.

jouvin: Good catch!

jouvin:
As for me, ready for merging! @stdweird any other remark?!

ncm-afsclt/src/main/perl/afsclt.pm
```diff
@@ -25,55 +25,60 @@ Readonly my $THESECELLS    => '/usr/vice/etc/TheseCells';
...
-        $self->error("Cannot get $PREFIX/thiscell (required in profile)");
+    unless ( defined($config->{thiscell}) ) {
+        $self->error("$PREFIX/thiscell missing in the configuration");
         return 1;
     }
 
-    my $afscell = $config->getValue("$PREFIX/thiscell");
     my $thiscell_fh = CAF::FileWriter->new( $THISCELL, log => $self );
-    print $thiscell_fh "$afscell\n";
+    print $thiscell_fh $config->{thiscell}."\n";
```
stdweird: there's no need for the stirng concat, `"$config->{thiscell}\n"` should also work, now that `$config` is a hashref.

stdweird:
@jouvin LGTM (and :+1:  for further cleanup and tests)
@jrha not sure if we should wait for feedback from @ajf8 , esp if this is for 16.6.

jouvin:
@ajf8 feedback would be great but I really think I introduced no feature changes but just fixed things that were addressing just the MS way of configuring AFS but had some flaws due to the lack of unit testing (the main change introduced by this PR!).

ncm-afsclt/src/main/pan/components/afsclt/schema.pan
```diff
@@ -5,21 +5,20 @@
...
-   				 # "Shall AFS client be active ?"
-   "afsd_args"  ? string{} # /etc/afsd.args values for rc.afs
+   "thiscell"   : string         # AFS home cell
+   "thesecells" ? string[]       # Cell list to authenticate to
+   "settime"    ? boolean        # Shall AFS client sync sys time?
+   "cellservdb" ? string         # Where Master CellServDB can be found
+   "afs_mount"  ? string         # AFS mount point (e.g. /afs)
+   "cachemount" ? string         # AFS cache location (/usr/vice/etc/cache)
+   "cachesize"  ? string         # AFS cache size in kB
+   "enabled"    : legacy_binary_affirmation_string  # Shall AFS client be active ?
```
ned21: IMHO this is a retrograde step for readability. I now need to locate this type defined in some other file before I can understand how to populate this mandatory field, which has no default. legacy_yes_or_no_string would be more intuitive.

jouvin: I tend to agree... I was thinking the same thing when doing the modification suggested by @stdweird ...

stdweird: Feel free to open an issue to change. This is work done in https://github.com/quattor/template-library-core/pull/111 and https://github.com/quattor/configuration-modules-core/issues/749; but renaming the type should be very trivial.

jouvin: Done. https://github.com/quattor/template-library-core/issues/118.

jrha: IMHO it would be better to make the effort to change the type to boolean instead of renaming the legacy type.

ajf8:
Thanks for the improvements @jouvin and sorry about the break. I think this module needed some work, hopefully I got some of it done, but it looks much better now especially with the unit tests.

Please feel free to merge, I will test it sometime and submit a new pull request if anything is needed.

stdweird:
@ajf8 thanks for the feedback

PR #818, opened 2016-07-20
ncm-afsclt: Fix warning caused redeclaration of variable with my.

ajf8:
The updates to afsclt seem to work fine here. I just noticed a perl warning while running ncm-ncd.


jrha:
Thanks!

PR #821, opened 2016-07-27
ncm-spma: Add backend for apt

jrha:
Add a SPMA backend for configuring debian derived systems using apt,
this should be considered a technology preview.

The code currently manages source lists (repositories) and packages and not much else.

Tested at RAL on Ubuntu 14.04.4 LTS and Linux Mint 17.3 Rosa.

Based on and depends on #716, so it might be better to wait until that is merged before attempting too much of a review.


jouvin:
Thanks for this great contribution! This is certainly something we may be interested in...

jrha:
I should also add that at at 31.4% the test coverage is not yet where I would like it to be, so I'll be working on that when I have some more time.

jrha:
Discovered that you can review a subset of changes, so https://github.com/quattor/configuration-modules-core/pull/821/files/339a6e87c28fde81e3ac7bc593296738925b65fa..HEAD can be used for review until #716 is merged.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my $current = Set::Scalar->new(map(m{(.*)\.list$}, readdir($dir)));
+
+    closedir($dir);
+    my $allowed = Set::Scalar->new(map($_->{name}, @$allowed_repos));
+    my $rm = $current-$allowed;
+    foreach my $i (@$rm) {
+        # We use $f here to make Devel::Cover happy
+        my $f = "$repo_dir/$i.list";
+        $self->verbose("Unlinking outdated repository $f");
+        if (!unlink($f)) {
```
ned21: This isn't NoAction safe.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+
+# If user packages are not allowed, removes any repositories present
+# in the system that are not listed in $allowed_repos.
+sub cleanup_old_repos
+{
+    my ($self, $repo_dir, $allowed_repos, $allow_user_pkgs) = @_;
+    $self->debug(5, 'Entered cleanup_old_repos()');
+
+    return 1 if $allow_user_pkgs;
```
ned21: I know this is a problem with the yum code too but at some point we need to split this into "allow_user_repos" and "allow_user_pkgs". For new code I would say "allow_user_pkgs" means that only -- manually installed packages are allowed to stay but manually configured repos get deleted.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+# Creates the repository dir if needed.
+sub initialize_repos_dir
+{
+    my ($self, $repo_dir) = @_;
+    $self->debug(5, 'Entered initialize_repos_dir()');
+
+    if (! -d $repo_dir) {
+        $self->verbose("$repo_dir didn't exist. Creating it");
+        if (!eval{mkpath($repo_dir)} || $@) {
```
ned21: What happens under NoAction?

jrha: Bad things I assume, I took this code from the yum backend, which handles NoAction in strange ways, I'll take a look at it again.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+{
+    my ($self, $config) = @_;
+    $self->debug(5, 'Entered Configure()');
+
+    # Get configuration trees
+    my $tree_repos = $config->getElement($TREE_REPOS)->getTree();
+    my $tree_pkgs = $config->getElement($TREE_PKGS)->getTree();
+    my $tree_component = $config->getElement($TREE_COMPONENT)->getTree();
+    # Convert these crappily-defined fields into real Perl booleans.
+    $tree_component->{run} = $tree_component->{run} eq 'yes';
```
ned21: If you intend to start converting these fields in other code to be either booleans or text, you should start that effort here by accepting both.

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,116 @@
...
+
+In this case, the error must be reported
+
+=cut
+
+initialize_repos();
+if ($EUID) {
+  chmod(0500, $REPO_DIR);
+} else {
+  diag("Trying to set immutable bit for root user to create unremovable files. ".
```
ned21: Surely any running of unit tests root should be an immediate fail?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
+# ${license-info}
```
stdweird: use new style like
https://github.com/quattor/configuration-modules-core/pull/808/files#diff-b5f0d02e208de8bb63cc453ae9e11a33L1

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+Packages listed under C</software/packages> will be installed, version and architecture locking (including multiarch) is fully implemented.
+
+=cut
+
+use strict;
+use warnings;
+use NCM::Component;
+our $EC=LC::Exception::Context->new->will_store_all;
+our @ISA = qw(NCM::Component);
```
stdweird: `use parent qw(NCM::Component);`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+use EDG::WP4::CCM::Element qw(escape unescape);
```
stdweird: once https://github.com/quattor/CCM/pull/102 is merged, you should import the escape functions from `CCM::Path` (altough current code will still work)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+use CAF::Process;
+use CAF::FileWriter;
+use CAF::FileEditor;
+use LC::Exception qw(SUCCESS);
```
stdweird: replace with `use CAF::Object qw(SUCCESS);` (we are phasing out `LC`)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+use Set::Scalar;
+use File::Path qw(mkpath);
+use Text::Glob qw(match_glob);
+use Readonly;
+
+Readonly::Scalar my $DIR_REPOS => "/etc/apt/sources.list.d";
+Readonly::Scalar my $DIR_PREFERENCES => "/etc/apt/preferences.d";
+Readonly::Scalar my $TEMPLATE_REPOS => "spma/apt-source.tt";
```
stdweird: move the TT files to an apt subdir (and open an issue for all non-common TT to move to a subdir as well)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+Readonly::Scalar my $TEMPLATE_PREFERENCES => "spma/apt-preferences.tt";
+Readonly::Scalar my $TREE_REPOS => "/software/repositories";
+Readonly::Scalar my $TREE_PKGS => "/software/packages";
+Readonly::Scalar my $TREE_COMPONENT => "/software/components/${project.artifactId}";
```
stdweird: wherever this is used, use `$self->prefix()` instead

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+    my $dir;
+    if (!opendir($dir, $repo_dir)) {
+        $self->error("Unable to read repositories in $repo_dir");
+        return 0;
+    }
+
+    my $current = Set::Scalar->new(map(m{(.*)\.list$}, readdir($dir)));
+
+    closedir($dir);
```
stdweird: replace all the above with `my $current = Set::Scalar->new(glob("$repo_dir/*.list"));`
non-existing directories should be handled with `CAF::Path` diretcory method. 
do a `use parent (NCM::Component CAF::Path);` above and `$self->directory_exists($repodir)` as test.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my $allowed = Set::Scalar->new(map($_->{name}, @$allowed_repos));
+    my $rm = $current-$allowed;
```
stdweird: `$rm` -> `$to_remove`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    foreach my $i (@$rm) {
```
stdweird: `$i` -> `$repo`?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+        # We use $f here to make Devel::Cover happy
+        my $f = "$repo_dir/$i.list";
```
stdweird: `$f` -> `$fn`

ncm-spma/src/main/perl/spma/apt.pm
stdweird: use `CAF::Path` `cleanup` method

ncm-spma/src/main/perl/spma/apt.pm
stdweird: use `CAF::Path` `directory` method

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+# Creates the repository dir if needed.
+sub initialize_repos_dir
+{
+    my ($self, $repo_dir) = @_;
+    $self->debug(5, 'Entered initialize_repos_dir()');
+
+    if (! -d $repo_dir) {
```
stdweird: test with `CAF::Path` `directory_exists` (or not at all, i think `CAF::Path` `directory` doesn't mind existing directories)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+            return 0;
+        }
+    }
+    return 1;
+}
+
+
+# Returns the URLs of the repositories listed in $prots, with their
+# host part replaced by each of the reverse @proxies.
+sub generate_reverse_proxy_urls
```
stdweird: is this code similar to other code? like yum or aii? (you can use existing methods in other modules :smile: )

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+{
+    my ($self, $prots, @proxies) = @_;
+    $self->debug(5, 'Entered generate_reverse_proxy_urls()');
+
+    my @l;
```
stdweird: no single letter variables

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    foreach my $pt (@$prots) {
+        foreach my $px (@proxies) {
+            my $url = $pt->{url};
+            $url =~ s{^(.*?):(/+)[^/]+}{$1:$2$px};
```
stdweird: should you handle the case that this regex doesn't match anything? it also looks like a confusiong way to write 
```perl
push(@l, ($pt->{url} =~ m/^(.*?):(/+)/) ? "$1:$2$px" : $pt->{url})
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+            push(@l, { url => $url });
+        }
+    }
+    return \@l;
+}
+
+
+# Generate the URLs for the forward proxies, based on whether the
+# repository is accessible over HTTP or HTTPS.
```
stdweird: i fail to see how the code below has anythig to do with `https?`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my ($self, $proto, @px) = @_;
+    $self->debug(5, 'Entered generate_forward_proxy_urls()');
+
+    $proto->[0]->{url} =~ m{^(.*?):};
+    return "$1://$px[0]";
+}
+
+
+# Generates the repository files in $repos_dir based on the contents
+# of the $repos subtree. It uses Template::Toolkit $template to render
```
stdweird: it should use `CCM::TextRender`, the `template` moethdo was removed in https://github.com/quattor/ncm-ncd/pull/68
see https://github.com/quattor/configuration-modules-core/pull/808 for an example

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+    my @px = split(",", $proxy) if $proxy;
+    @px = map("$_:$port", @px) if $port;
+
+    my $changes = 0;
+
+    foreach my $repo (@$repos) {
+        my $prots = $repo->{protocols}->[0];
+        my $fh = CAF::FileWriter->new("$repos_dir/$repo->{name}.list", log => $self);
+        print $fh "# File generated by ", __PACKAGE__, ". Do not edit\n";
```
stdweird: pass this via a header to textrender, or in the TT file, or not at all

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+        if (!exists($repo->{proxy}) && @px) {
+            if ($type eq 'reverse') {
+                $repo->{protocols} = $self->generate_reverse_proxy_urls($repo->{protocols}, @px);
+            } else {
+                $repo->{proxy} = $self->generate_forward_proxy_urls($repo->{protocols}, @px);
+            }
+        }
+        my $rs = $self->template()->process($template, $repo, $fh);
```
stdweird: use `CCM::TextRender`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+# Returns a set of all installed packages
+sub get_installed_pkgs
+{
+    my $self = shift;
+    $self->debug(5, 'Entered installed_pkgs()');
+
+    my $cmd = CAF::Process->new($CMD_DPKG_QUERY, keeps_state => 1, log => $self);
```
stdweird: as you do not use `$cmd` below, do a `my $out = ...->output();`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+    my $out = $cmd->output();
+    if ($?) {
+        return 0;
```
stdweird: better to return with `return;`, unless the `0` has any significance

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    }
+    my @pkgs = grep(m{^ii }, split(/\n/, $out));
```
stdweird: only a ` ` or `\s`?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    for (@pkgs) {s/^...;//g}; # Strip package status characters
```
stdweird: ? (esp the `/g` in combination with the `^`)
 add it with a map above 
```perl
my @pkgs = map {s/^.{3};//} grep {m/^ii\s/} split("\n", $out);
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+    return Set::Scalar->new(@pkgs);
+}
+
+
+sub get_package_version_arch
```
stdweird: small comment/docstring

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+                    push(@versions, sprintf('%s:%s=%s', $name, $arch, unescape($ver)));
+                }
+            }
+            else {
+                $self->debug(5, '  Adding package ', $name, ' with version ', $ver, ' but without architecture to list');
+                push(@versions, sprintf('%s=%s', $name, unescape($ver)));
+            }
+        }
+    }
+    else {
```
stdweird: use `} else {` on one line

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my @versions;
+
+    if ($details) {
+        while (my ($ver, $archs) = each(%$details)) {
+            if ($archs->{arch}) {
+                while (my $arch = each(%$archs->{arch})) {
+                    $self->debug(5, '  Adding package ', $name, ' with version ', $ver, ' and architecture ', $arch, ' to list');
+                    push(@versions, sprintf('%s:%s=%s', $name, $arch, unescape($ver)));
+                }
+            }
```
stdweird: one line

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+sub get_package_version_arch
+{
+    my ($self, $name, $details) = @_;
+    $self->debug(5, 'Entered get_package_version_arch()');
+
+    my @versions;
+
+    if ($details) {
+        while (my ($ver, $archs) = each(%$details)) {
```
stdweird: each is evil https://github.com/quattor/CAF/issues/109 (and not sorted)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    else {
+        $self->debug(5, '  Adding package ', $name, ' without version or architecture to list');
+        push(@versions, $name);
+    }
+
+    return \@versions;
+}
+
+
+sub apply_package_version_arch
```
stdweird: docstring / comment

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+{
+    my ($self, $packagelist, $packagetree) = @_;
+
+    $self->debug(5, 'Entered apply_package_version_arch()');
+
+    my @packages = @$packagelist;
+    my @results;
+
+    while (my $name = shift(@packages)) {
```
stdweird: simple loop thorugh them, the `while/shift` adds nothing
```perl
foreach my $name (@$packagelist) {
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+        my $name_escaped = escape($name);
+        if (exists($packagetree->{$name_escaped})) {
+            my $versions = $self->get_package_version_arch($name, $packagetree->{$name_escaped});
```
stdweird: does the `apt` support globs in package name like `kernel*`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+            push(@results, @$versions);
+        }
+        else {
+            $self->error('Could not find package ', $name, ' in tree');
```
stdweird: i would collect all packages that are not found (`push*(@notfound, $name)`) , and once report one error
```perl
$self->error("Could not find packages", join(", ", @notfound)) if @notfound`
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my ($self, $pkgs) = @_;
+    $self->debug(5, 'Entered get_desired_pkgs()');
+
+    my $packages = Set::Scalar->new();
+
+    while (my $name_escaped = each(%$pkgs)) {
+        my $name = unescape($name_escaped);
+        if (!$name) {
+            $self->error("Invalid package name: ", unescape($name));
+            return 0;
```
stdweird: return in the middle of an each... :see_no_evil: :hear_no_evil: :speak_no_evil:  https://github.com/quattor/CAF/issues/109

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+sub get_desired_pkgs
+{
+    my ($self, $pkgs) = @_;
+    $self->debug(5, 'Entered get_desired_pkgs()');
+
+    my $packages = Set::Scalar->new();
+
+    while (my $name_escaped = each(%$pkgs)) {
+        my $name = unescape($name_escaped);
+        if (!$name) {
```
stdweird: any examples?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+
+# Update package metadata from upstream repositories
+sub resynchronize_package_index
+{
+    my $self = shift;
+    $self->debug(5, 'Entered resynchronize_package_index()');
+
+    my $cmd = CAF::Process->new($CMD_APT_UPDATE, keeps_state => 1, log => $self);
+    $cmd->execute() or return 0;
```
stdweird:
```perl
return $cmd->execute() ? 1 : undef;
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+# Upgrade existing packages
+sub upgrade_packages
+{
+    my ($self) = @_;
+    $self->debug(5, 'Entered upgrade_packages()');
+
+    my $cmd = CAF::Process->new($CMD_APT_UPGRADE, keeps_state => 1, log => $self);
```
stdweird: this is not `keeps_state => 1`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+# Install packages
+sub install_packages
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered install_packages()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_INSTALL, @$packages], keeps_state => 1, log => $self);
```
stdweird: also not `keeps_state => 1`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+
+# Mark packages as automatically installed
+# this signals to apt that they are no longer required and may be cleaned up by autoremove
+sub mark_packages_auto
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered mark_packages_auto()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_MARK, 'auto', @$packages], keeps_state => 1, log => $self);
```
stdweird: is this `keeps_state => 1`? i think not.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+# Remove automatically installed packages
+sub autoremove_packages
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered autoremove_packages()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_AUTOREMOVE, @$packages], keeps_state => 1, log => $self);
```
stdweird: no keeps_state

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+}
+
+
+sub Configure
+{
+    my ($self, $config) = @_;
+    $self->debug(5, 'Entered Configure()');
+
+    # Get configuration trees
+    my $tree_repos = $config->getElement($TREE_REPOS)->getTree();
```
stdweird: new style
```perl
$config->getTree($TREE_REPOS);
```

ncm-spma/src/main/perl/spma/apt.pm
stdweird: i'd rather not. this should be handled in CCM (via some gettree option), before everyone starts to copy paste something that will be some pretty ugly code. (if you open an issue in CCM, i might come up with something)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my $tree_pkgs = $config->getElement($TREE_PKGS)->getTree();
+    my $tree_component = $config->getElement($TREE_COMPONENT)->getTree();
+    # Convert these crappily-defined fields into real Perl booleans.
+    $tree_component->{run} = $tree_component->{run} eq 'yes';
+    $tree_component->{userpkgs} = defined($tree_component->{userpkgs}) && $tree_component->{userpkgs} eq 'yes';
```
stdweird: make it mandatory in the schema with default false?

stdweird: or at least `$tree_component->{userpkgs} = ($tree_component->{userpkgs} || "no") eq 'yes';`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+
+    $self->initialize_repos_dir($DIR_REPOS);
```
stdweird: check return codes

ncm-spma/src/main/resources/apt-source.tt
```diff
@@ -0,0 +1,17 @@
...
+# for more details.
+
+# Format is:
+#     deb [ options ] uri suite [component1] [component2] [...]
+
+# Name: [% name %]
+# Owner: [% owner %]
+
+[%- IF gpgcheck -%]
+[%- SET trust = "trusted=yes " -%]
```
stdweird: indent

stdweird: or better yet
```
[% trust = gpgcheck ? "trusted=yes " : "" %]
```

stdweird: or below, with the inermediate `trust`

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,116 @@
...
+use Test::Quattor;
+use Test::More;
+use NCM::Component::spma::apt;
+
+use File::Path qw(mkpath rmtree);
+use LC::File;
+
+Readonly my $REPO_DIR => "target/test/clean.sources.list.d";
+
+sub initialize_repos
```
stdweird: this is not needed if you switch to `CAF::Path`, it's all mocked

ncm-spma/src/test/perl/apt-cleanup-sources.t
stdweird: `CAF::Path` methods are unittested, but there is no way to trigger failed mocked operation (like a mcoked failed file removal). open an issue for that in `maven-tools`; it shouldn't be that hard to add.

ncm-spma/src/test/perl/apt-list-installed.t
```diff
@@ -0,0 +1,40 @@
...
+
+=cut
+
+use strict;
+use warnings;
+use Test::Quattor;
+use Test::More;
+use NCM::Component::spma::apt;
+
+plan skip_all => "Cannot test get_installed_pkgs, dpkg not present" if ! -x "/usr/bin/dpkg";
```
stdweird: ? why is this needed? should be mocked, if not, open issue in `maven-tools` for whatever is missing.

ncm-spma/src/test/perl/apt-sources-generate.t
```diff
@@ -0,0 +1,253 @@
...
+=cut
+
+ok(defined($cmp->generate_repos($REPOS_DIR, $repos, $REPOS_TEMPLATE)), "Basic repository correctly created");
+
+
+my $fh = get_file("$REPOS_DIR/a_source.list");
+ok(defined($fh), "Correct file opened");
+
+my $name = $repos->[0]->{name};
+like("$fh", qr{^\# Name: $name$}m, "Source got the correct name");
```
stdweird: use TT unittests rather than reinventing these here.

ncm-spma/src/main/perl/spma/apt.pm
piojo: And it is merged now, so please rely on `CCM::Path`.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+use CAF::Process;
+use CAF::FileWriter;
+use CAF::FileEditor;
+use LC::Exception qw(SUCCESS);
+use Set::Scalar;
+use File::Path qw(mkpath);
+use Text::Glob qw(match_glob);
+use Readonly;
+
+Readonly::Scalar my $DIR_REPOS => "/etc/apt/sources.list.d";
```
piojo:
```perl
Readonly my $DIR_REPOS => "/etc/apt/sources.list.d";
```
is shorter and more readable IMHO. Not a big deal, though.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+    my $dir;
+    if (!opendir($dir, $repo_dir)) {
+        $self->error("Unable to read repositories in $repo_dir");
+        return 0;
+    }
+
+    my $current = Set::Scalar->new(map(m{(.*)\.list$}, readdir($dir)));
+
+    closedir($dir);
+    my $allowed = Set::Scalar->new(map($_->{name}, @$allowed_repos));
```
piojo: I know standard Quattor code relies on parentheses a lot. But Perl usually prefers

```perl
map {$_->{name}}, @$allowed_repos
```
Parentheses in `map`, `grep` and `sort` cause lots of headaches because of ambiguities.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Keep locals short, please. `$i` or `$r` are more than enough for this little loop.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Locals should be as short as possible. :)

ncm-spma/src/main/perl/spma/apt.pm
piojo: @stdweird please provide an example of how this code should look like

ncm-spma/src/main/perl/spma/apt.pm
piojo: Absolutely.  `$self->prefix` is what lets component aliasing work.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Please provide an example here to help the developer. :smile:

ncm-spma/src/main/perl/spma/apt.pm
piojo: Indeed. Please remember that `keeps_state => 1` makes the command be executed even under `NoAction`. It is meant for trivial things like `ls`, `ip addr show` and the like.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Also, line too long.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Line too long, but that is fixed as soon as `keeps_state` is removed.

ncm-spma/src/main/perl/spma/apt.pm
piojo: Yup. We are not in a position to accept exceptions yet.

```perl
$self->initialize_repos_dir($DIR_REPOS) or return 0;
```

and so on.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,433 @@
...
+        $tree_component->{proxyport},
+    );
+
+    $self->resynchronize_package_index();
+
+    $self->upgrade_packages();
+
+    my $packages_installed = $self->get_installed_pkgs();
+    my $packages_desired = $self->get_desired_pkgs($tree_pkgs);
+    my $packages_unwanted = $packages_installed->difference($packages_desired);
```
piojo: Sets support direct substraction, which I find nicer:

```perl
my $packages_unwanted = $packages_installed - $packages_desired;
```

ncm-spma/src/test/perl/apt-sources-generate.t
piojo: Provide an example, please :)

jrha:
I managed to devote some time to this today, it's looking much better but not quite there yet... still needs more work.

ncm-spma/src/main/pan/components/spma/apt/schema.pan
```diff
@@ -0,0 +1,16 @@
...
+# ${author-info}
+
+declaration template components/spma/apt/schema;
+
+include 'components/spma/schema';
+
+type component_spma_apt = {
+    include structure_component
+    include component_spma_common
+    @{ Allow user defined (i.e. unmanaged) repositories to be present on the system }
```
stdweird: ah, the shortest way to for users to take over the machine. you sure you want this?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
+#${PMpre} NCM::Component::spma::apt${PMpost}
+
+=head1 NAME
+
+NCM::Component::spma::apt - NCM SPMA backend for apt
```
stdweird: all modules `C<>`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+=head1 SYNOPSIS
+
+This document describes how to control the behaviour of the package manager itself.
+For information on how to manage packages with Quattor, please check
+L<http://quattor.org/documentation/2013/04/05/package-management.html>.
+
+=head1 DESCRIPTION
+
+apt.pm implements an apt backend for ncm-spma, the approach taken is to defer as much work as possible to apt.
```
stdweird: `apt.pm` -> `This plugin`?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+A single SPMA run consists of the following steps:
+
+=over
+
+=item
+
+Setup source directory if required
```
stdweird: out of interest, does the pod look different if you put this after the `=item` or in a new paragraph like this?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+Mark any packages installed but not in the profile as automatically installed
+
+=item
+
+Ask apt to remove all automatically installed packages that are not satisfying dependencies of other packages
+
+=back
+
+Support for groups (tasks) is not currently provided.
```
stdweird: i tought we were going to remove it?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+=head1 RESOURCES
+
+Only a very minimal schema is implemented.
+
+Sources listed under C</software/repositories> will be configured,
+URLs should be followed by the suite and sections required e.g. http://example.org/debian unstable main
```
stdweird: but the example in `C<>`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+use EDG::WP4::CCM::Path qw(escape unescape);
+use EDG::WP4::CCM::TextRender;
+use NCM::Component::spma::yum;
+use CAF::Object qw(SUCCESS);
+use Set::Scalar;
+use File::Path qw(mkpath);
+use Text::Glob qw(match_glob);
+use Readonly;
+
+Readonly::Scalar my $DIR_SOURCES => "/etc/apt/sources.list.d";
```
stdweird: you can  drop the `::Scalar`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+sub cleanup_old_sources
+{
+    my ($self, $sources_dir, $allowed_sources) = @_;
+    $self->debug(5, 'Entered cleanup_old_sources()');
+
+    my $current = Set::Scalar->new(glob("$sources_dir/*.list"));
+    my $allowed = Set::Scalar->new(map($_->{name}, @$allowed_sources));
+    my $to_remove = $current-$allowed;
+
+    if (-d $sources_dir) {
```
stdweird: use `CAF::Path` and `->directory_exists`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+        foreach my $s (@$to_remove) {
+            my $fn = "$sources_dir/$s.list";
+            $self->verbose("Unlinking outdated source $fn");
+            if (!defined (CAF::Path->cleanup($fn, ""))) {
```
stdweird: with `use CAF::Path` i meant, add `CAF::Path` to the parent list at the end do `$self->cleanup()`. this probably doesn't work like you think it does.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+}
+
+
+# Creates the sources directory if needed.
+sub initialize_sources_dir
+{
+    my ($self, $sources_dir) = @_;
+    $self->debug(5, 'Entered initialize_sources_dir()');
+
+    if (CAF::Path->directory($sources_dir) != $sources_dir) {
```
stdweird: this is not what you want. for string comparison you need `ne`, but on failure, this will give a warning, since `->directory` retruns undef.
you can do
```perl
if (! $self->directory($sources_dir)) {
    $self->error("Unable to create source dir $sources_dir: $self->{fail}");
    return 0;
}
```

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+    # db:Status-Abbrev is three characters, we are looking for
+    #   desired action  i (but anything is fine here)
+    #   package status  i (indicating currently installed)
+    #   error flag      a single space (indicating no error condition)
+    my @pkgs = map {s/^.{3};//} grep {m/^.i /} split("\n", $out);
+
+    return Set::Scalar->new(@pkgs);
+}
+
+# Build list of packages that should be present on system from configuration tree
```
stdweird: is this correct? what do `name` and `details` do?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+sub get_package_version_arch
+{
+    my ($self, $name, $details) = @_;
+    $self->debug(5, 'Entered get_package_version_arch()');
+
+    my @versions;
+
+    if ($details) {
+        foreach my $version (keys %$details) {
```
stdweird: `sort keys`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+            my $params = $details->{$version};
+            if ($params->{arch}) {
+                foreach my $arch (keys %{ $params->{arch} }) {
+                    $self->debug(5, '  Adding package ', $name, ' with version ', $version, ' and architecture ', $arch, ' to list');
```
stdweird: unescape version?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+                    push(@versions, sprintf('%s:%s=%s', $name, $arch, unescape($version)));
+                }
+            } else {
+                $self->debug(5, '  Adding package ', $name, ' with version ', $version, ' but without architecture to list');
```
stdweird: unescape version?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+    my ($self, $name, $details) = @_;
+    $self->debug(5, 'Entered get_package_version_arch()');
+
+    my @versions;
+
+    if ($details) {
+        foreach my $version (keys %$details) {
+            my $params = $details->{$version};
+            if ($params->{arch}) {
+                foreach my $arch (keys %{ $params->{arch} }) {
```
stdweird: sort keys

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+    foreach my $name (@$packagelist) {
+        my $name_escaped = escape($name);
+        if (exists($packagetree->{$name_escaped})) {
+            my $versions = $self->get_package_version_arch($name, $packagetree->{$name_escaped});
+            push(@results, @$versions);
+        } else {
+            push(@notfound, $name);
+        }
+        $self->error("Could not find packages", join(", ", @notfound)) if @notfound;
```
stdweird: do this outside the loop?

stdweird: also add a space after `packages` (all arguments to reporter methods are appended, no spacing is added)

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+# Return a set of desired packages.
+sub get_desired_pkgs
+{
+    my ($self, $pkgs) = @_;
+    $self->debug(5, 'Entered get_desired_pkgs()');
+
+    my $packages = Set::Scalar->new();
+
+    foreach my $name_escaped (keys %$pkgs) {
```
stdweird: sort keys

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+        my $name = unescape($name_escaped);
+        if (!$name) {
+            $self->error("Invalid package name: ", unescape($name));
```
stdweird: just `$name`? otherwsie double unescaped name

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+sub get_desired_pkgs
+{
+    my ($self, $pkgs) = @_;
+    $self->debug(5, 'Entered get_desired_pkgs()');
+
+    my $packages = Set::Scalar->new();
+
+    foreach my $name_escaped (keys %$pkgs) {
+        my $name = unescape($name_escaped);
+        if (!$name) {
```
stdweird: this is an odd test. why is it here?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+}
+
+
+# Update package metadata from upstream sourcesitories
+sub resynchronize_package_index
+{
+    my $self = shift;
+    $self->debug(5, 'Entered resynchronize_package_index()');
+
+    my $cmd = CAF::Process->new($CMD_APT_UPDATE, log => $self);
```
stdweird: if it is only updating package metadata, you could do `keeps_state => 1`; if it does more than that, please update the comment

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+
+# Upgrade existing packages
+sub upgrade_packages
+{
+    my ($self) = @_;
+    $self->debug(5, 'Entered upgrade_packages()');
+
+    my $cmd = CAF::Process->new($CMD_APT_UPGRADE, log => $self);
+    $cmd->execute() or return 0;
```
stdweird: do a `return $cmd->execute() ? 1 : 0` as in previous method

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+
+# Install packages
+sub install_packages
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered install_packages()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_INSTALL, @$packages], log => $self);
+    $cmd->execute() or return 0;
```
stdweird: same as above

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+# Mark packages as automatically installed
+# this signals to apt that they are no longer required and may be cleaned up by autoremove
+sub mark_packages_auto
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered mark_packages_auto()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_MARK, 'auto', @$packages], log => $self);
+    $cmd->execute() or return 0;
```
stdweird: same as above

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+
+# Remove automatically installed packages
+sub autoremove_packages
+{
+    my ($self, $packages) = @_;
+    $self->debug(5, 'Entered autoremove_packages()');
+
+    my $cmd = CAF::Process->new([@$CMD_APT_AUTOREMOVE, @$packages], log => $self);
+    $cmd->execute() or return 0;
```
stdweird: same as above

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+
+
+sub Configure
+{
+    my ($self, $config) = @_;
+    $self->debug(5, 'Entered Configure()');
+
+    # Get configuration trees
+    my $tree_sources = $config->getTree($TREE_SOURCES);
+    my $tree_pkgs = $config->getElement($TREE_PKGS)->getTree();
```
stdweird: no `getElement`, use getTree on config directlyt as previous line

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,385 @@
...
+    my $tree_component = $config->getElement($self->prefix())->getTree();
```
stdweird: same remark

ncm-spma/src/test/perl/apt-apply-package-version-arch.t
```diff
@@ -0,0 +1,52 @@
...
+    "example-package:i686=1.5.0-1",
+    "example-package:amd64=1.5.0-1",
+    "example-package:amd64=1.7.1-1",
+    "example-package=1.9.0-1",
+    "another-package-name",
+];
+
+my $cmp = NCM::Component::spma::apt->new("spma");
+
+cmp_bag($cmp->apply_package_version_arch($set, $tree), $desired_details, "Package details translated");
```
stdweird: why not `is_deeply` from `Test::More`?

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,96 @@
...
+    );
+
+    system("chattr +i $SOURCE_DIR/*list");
+}
+is($cmp->cleanup_old_sources($SOURCE_DIR, $sources), 0, "Error reported when an outdated repo cannot be removed");
+is($cmp->{ERROR}, 2, "Error in unlink is reported");
+
+# Restore permissions on the source for future executions.
+if ($EUID) {
+  chmod(0700, $SOURCE_DIR);
```
stdweird: fix indent

ncm-spma/src/test/perl/apt-get-package-version-arch.t
```diff
@@ -0,0 +1,42 @@
...
+my $desired_details = [
+    "example-package:i686=1.5.0-1",
+    "example-package:amd64=1.5.0-1",
+    "example-package:amd64=1.7.1-1",
+    "example-package=1.9.0-1",
+];
+
+my $cmp = NCM::Component::spma::apt->new("spma");
+
+cmp_bag($cmp->get_package_version_arch('example-package', $details), $desired_details, "Package details translated");
```
stdweird: `is_deeply`

ncm-spma/src/test/perl/apt-list-installed.t
```diff
@@ -0,0 +1,38 @@
...
+
+use strict;
+use warnings;
+use Test::Quattor;
+use Test::More;
+use NCM::Component::spma::apt;
+
+my $cmp = NCM::Component::spma::apt->new("spma");
+
+my $pkgs = $cmp->get_installed_pkgs();
```
stdweird: how does this work without mocked command outpput?

ncm-spma/src/test/perl/apt-sources-generate.t
```diff
@@ -0,0 +1,111 @@
...
+            excludepkgs => [qw(baz quux)],
+        }
+    ];
+}
+
+my $sources = initialise_sources();
+
+my $mock = Test::MockModule->new('CAF::FileWriter');
+
+$mock->mock('cancel', sub {
```
stdweird: why is this needed?

stdweird:
@jrha nice progress!

ncm-spma/src/main/pan/components/spma/apt/schema.pan
jrha: Yes!
It's something @ned21 suggested, currently `userpkgs` implies this. Having a separate switch for repositories which defaults to `false` seems like an improvement.

ncm-spma/src/main/perl/spma/apt.pm
jrha: Good point.

ncm-spma/src/main/perl/spma/apt.pm
jrha: Nope! I just tried:

```perl
=item Remove sources that are not found in the profile

=item

Update source cache from upstream sources
```

For which `pod2html` generates:

```html
<li><p>Remove sources that are not found in the profile</p>

</li>
<li><p>Update source cache from upstream sources</p>

</li>
```

ncm-spma/src/main/pan/components/spma/apt/schema.pan
stdweird: hmm, true. never tought about it. do we allow userpkgs = false with userrepos = true?

ncm-spma/src/main/perl/spma/apt.pm
jrha: What does unescape return if it fails to unescape?

ncm-spma/src/test/perl/apt-apply-package-version-arch.t
jrha: `is_deeply` will complain about result are ordering, `cmp_bag` will not.

ncm-spma/src/main/pan/components/spma/apt/schema.pan
jrha: Why not, it would make sense if you wanted something else (e.g. RedHat Satellite) managing your repositories.

ncm-spma/src/test/perl/apt-list-installed.t
jrha: It doesn't?

ncm-spma/src/main/perl/spma/apt.pm
jrha: Isn't updating the system's metadata cache something that changes the state of the system?

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -65,6 +65,10 @@ type component_spma_yum = {
     "quattor_os_release" ? string # string to write to quattor_os_file
     "plugins" ? spma_yum_plugins
     "main_options" ? spma_yum_main_options
+    @{ Run the SPMA after configuring it }
+    "run" ? legacy_binary_affirmation_string
```
ned21: I don't believe the yum backend supports this attribute so it should not appear here?

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
```diff
@@ -66,6 +66,10 @@ type component_spma_yum = {
     "quattor_os_release" ? string # string to write to quattor_os_file
     "plugins" ? spma_yum_plugins
     "main_options" ? spma_yum_main_options
+    @{ Run the SPMA after configuring it }
+    "run" ? legacy_binary_affirmation_string
+    @{ Allow user installed (i.e. unmanaged) packages to be present on the system }
+    "userpkgs" ? legacy_binary_affirmation_string
 };
 
 bind "/software/components/spma" = component_spma_yum;
```
ned21: Should be component_spma_yumng instead?

ncm-spma/src/main/perl/spma/apt.pm
ned21: I would "yes" but it can be difficult to make NoAction give useful info if you don't have an up to date cache so I think keeps_state=>1 might be OK?

ncm-spma/src/main/pan/components/spma/yumng/schema.pan
jrha: I've been working on cleaning up the schema in a seperate branch, I think this would be better fixed there.

ncm-spma/src/main/pan/components/spma/yum/schema.pan
jrha: Good point, removed.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,386 @@
...
+sub cleanup_old_sources
+{
+    my ($self, $sources_dir, $allowed_sources) = @_;
+    $self->debug(5, 'Entered cleanup_old_sources()');
+
+    my $current = Set::Scalar->new(glob("$sources_dir/*.list"));
+    my $allowed = Set::Scalar->new(map($_->{name}, @$allowed_sources));
+    my $to_remove = $current-$allowed;
+
+    if ($self->directory_exists($sources_dir)) {
```
jrha: @stdweird I must be having another stupid moment - this is giving me problems with the unit tests.
`directory_exists` appears to get mocked by `Test:Quattor` and then doesn't behave as expected, replacing the line with `if (-d $sources_dir) {` allows the tests to pass but that just seems like a bad idea.

stdweird: what test is failing? you also need latest build-tools and everything has to be mocked, but should all work

ncm-spma/src/test/perl/apt-apply-package-version-arch.t
stdweird: ok, if the order is not relevant.

ncm-spma/src/main/perl/spma/apt.pm
jrha: `apt-cleanup-sources.t` is the source of my pain, coming from the yum backend it's mostly not mocked and I'm getting confused.

jrha: :relieved:  I've rewritten the whole set of tests with lots of mocking.

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,394 @@
...
+Readonly my $CMD_APT_INSTALL => [$BIN_APT_GET, qw(-qq install)];
+Readonly my $CMD_APT_AUTOREMOVE => [$BIN_APT_GET, qw(-qq autoremove)];
+Readonly my $CMD_APT_MARK => [$BIN_APT_MARK, qw(-qq)];
+Readonly my $CMD_DPKG_QUERY => [$BIN_DPKG_QUERY, qw(-W -f=${db:Status-Abbrev};${Package}\n)];
+
+our $NoActionSupported = 1;
+
+
+# Wrap glob builtin to allow for unittesting
+sub _glob
```
stdweird: use https://github.com/quattor/CAF/pull/230, which has support for Test::Quattor unittests

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,394 @@
...
+# the file. Optionally, proxy information will be used. In that case,
+# it will use the $proxy host, wich is of $type "reverse" or
+# "forward", and runs on the given $port.
+# Returns undef on errors, or the number of source files that were changed.
+sub generate_sources
+{
+    my ($self, $sources_dir, $sources, $template, $proxy, $type, $port) = @_;
+    $self->debug(5, 'Entered generate_sources()');
+
+    #my @px = split(",", $proxy) if $proxy;
```
stdweird: add an issue to fix/address or remove the comments

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,394 @@
...
+    #@px = map("$_:$port", @px) if $port;
+
+    my $changes = 0;
+
+    foreach my $source (@$sources) {
+        $self->debug(5, "Generating source: $source->{name}");
+        my $prots = $source->{protocols}->[0];
+        # if (!exists($source->{proxy}) && @px) {
```
stdweird: open an issue to include them or remove these comments

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+
+my $cmp = NCM::Component::spma::apt->new("spma");
+my $mock = Test::MockModule->new('NCM::Component::spma::apt');
+
+my $immutable;
+my $directory_exists;
+my $sourcefiles;
+
+
+$mock->mock('_glob', sub {
```
stdweird: if you use listdir from CAF::Path, this is not needed

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+    return @$sourcefiles;
+});
+
+
+$mock->mock('cleanup', sub {
```
stdweird: whole of CAF::Path is already mocked and integrated in Test::Quattor 1.54 (even the not yet merged CAF::Path listdir)

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+    if (!$immutable) {
+        my $filename = shift;
+        $sourcefiles->delete($filename) || return undef;
+        return 1;
+    };
+    return undef;
+});
+
+
+$mock->mock('directory_exists', sub {
```
stdweird: no need to mock this. you can create a mock directory within Test::Quattor by either creating a mocked file (which will create the mocked parent tree) or directly creating a mocked directory.
example below or in PR, lets see what is easier

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+=pod
+
+=head2 The source directory doesn't exist
+
+If the source directory doesn't exist, the error should be
+reported.
+
+=cut
+
+$immutable = 0;
```
stdweird: remove these lines and mocking above, just delete the directory using either

CAF::Path cleanup

```perl
$cmp->cleanup($SOURCE_DIR);
```
or Test::Quattor function
```perl
remove_any($SOURCE_DIR);
```

and that's it. remember, `CAF::Path` is mocked entirely in `Test::Quattor`. the only thing you need to do is to make the fake file/directory structure.

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+
+=pod
+
+=head2 The source exists and contains files that should be removed
+
+In this case, any unwanted files should be removed
+
+=cut
+
+$immutable = 0;
```
stdweird: remove these, just create the files, parent dir will exist.

```perl
foreach my $fn (qw(source1 source2)) {
  set_file_contents("$SOURCE_DIR/$_.list", "")
};
```
this will create the parent dir and "touch" the files.

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+$directory_exists = 1;
+$sourcefiles = Set::Scalar->new(map { "$SOURCE_DIR/$_.list" } qw(source1 source2));
+
+ok($cmp->cleanup_old_sources($SOURCE_DIR, $sources), "Cleanup old source files");
+ok($sourcefiles->contains("$SOURCE_DIR/source1.list"), "Preserve allowed source file");
```
stdweird: this should then become 

```
ok($cmp->file_exists("$SOURCE_DIR/source1.list"), "Preserve allowed source file");
ok(!$cmp->file_exists("$SOURCE_DIR/source2.list"), "Remove unwanted source file");
```

ncm-spma/src/test/perl/apt-cleanup-sources.t
```diff
@@ -0,0 +1,117 @@
...
+
+In this case, the error must be reported
+
+=cut
+
+$immutable = 1;
+$directory_exists = 1;
+$sourcefiles = Set::Scalar->new(map { "$SOURCE_DIR/$_.list" } qw(source1 source2));
+
+ok(!$cmp->cleanup_old_sources($SOURCE_DIR, $sources), "Report error when an outdated repo cannot be removed");
```
stdweird: i guess you want to test a failed cleanup? this is something you can't do yet. i would open an issue to add this later, i plan to add this to maven-tools soonish

ncm-spma/src/test/perl/apt-sources-generate.t
```diff
@@ -0,0 +1,108 @@
...
+use Test::Quattor;
+use Test::More;
+use NCM::Component::spma::apt;
+use Test::MockModule;
+use Readonly;
+use CAF::Object;
+use CAF::FileWriter;
+use Cwd;
+
+$CAF::Object::NoAction = 1;
```
stdweird: if you use `Test::Quattor`, this is only needed to test noaction behaviour, not to prevent that the unittests actually write stuff.

ncm-spma/src/test/perl/apt-sources-generate.t
```diff
@@ -0,0 +1,108 @@
...
+=head2 Error handling
+
+Failures in rendering a template are reported, and nothing is written
+to disk.
+
+=cut
+
+is($cmp->generate_sources($SOURCES_DIR, $sources, "an invalid template name"), 0, "Invalid template name is detected");
+is($cmp->{ERROR}, 1, "Errors on template rendering are reported");
+$fh = get_file("$SOURCES_DIR/$name.list");
```
stdweird: you don't do anything with `$fh`

jrha:
Tests are failing due to ncm-path.

jrha:
@hpcugentbot test this please

stdweird:
@jrha conflicts due to common yum schema PR merged

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,379 @@
...
+Only a very minimal schema is implemented.
+
+Sources listed under C</software/repositories> will be configured,
+URLs should be followed by the suite and sections required e.g. C<http://example.org/debian unstable main>
+
+Packages listed under C</software/packages> will be installed, version and architecture locking (including multiarch) is fully implemented.
+
+=cut
+
+use strict;
```
stdweird: this and next line are part of the header template.

ncm-spma/src/main/perl/spma/apt.pm
stdweird: @jrha ping?

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,379 @@
...
+    $self->debug(5, 'Entered generate_sources()');
+
+    my $changes = 0;
+
+    foreach my $source (@$sources) {
+        $self->debug(5, "Generating source: $source->{name}");
+        my $prots = $source->{protocols}->[0];
+        my $rs = EDG::WP4::CCM::TextRender->new($template, $source, relpath => 'spma', log => $self);
+        if ($rs) {
+            my $fh = $rs->filewriter("$sources_dir/$source->{name}.list", log => $self);
```
stdweird: `log => $self` is default

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,379 @@
...
+
+# Write apt configuration file
+sub configure_apt
+{
+    my ($self, $config) = @_;
+    $self->debug(5, 'Entered configure_apt()');
+
+    my $tr = EDG::WP4::CCM::TextRender->new($TEMPLATE_CONFIG, $config, relpath => 'spma', log => $self);
+    if ($tr) {
+        my $fh = $tr->filewriter($FILE_CONFIG, log => $self);
```
stdweird: `log => $self` is default

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,379 @@
...
+    #   desired action  i (but anything is fine here)
+    #   package status  i (indicating currently installed)
+    #   error flag      a single space (indicating no error condition)
+    my @pkgs = map { substr $_, 4 } grep {m/^.i /} split("\n", $out);
+
+    return Set::Scalar->new(@pkgs);
+}
+
+# For a given package name, extract version and architecture from tree passed in details
+# returns a list of packages formatted with name, version and architecture for use with apt
```
stdweird: `list` ->  arrayref`

ncm-spma/src/main/perl/spma/apt.pm
```diff
@@ -0,0 +1,379 @@
...
+        }
+    } else {
+        $self->debug(5, '  Adding package ', $name, ' without version or architecture to list');
+        push(@versions, $name);
+    }
+
+    return \@versions;
+}
+
+# Apply versions from configuration tree to a bare list of packages
```
stdweird: what does it return?

jrha:
@stdweird I'll do my best to address your comments, but at this point I feel that this PR is better off being merged than not, it may not be perfect, but is probably good enough. On Thursday it will have been open for a year, which makes me sad.

stdweird:
@jrha fair enough. expect a PR to your PR soonish 😄

jrha:
I'll probably have some fixes in a few minutes, but might not find time to do anything about the test mocking.

stdweird:
ok, i can certainly take care of the test remarks. lets try to merge this by thursday.

jrha:
@stdweird I've addressed all the non-test stuff I can, rebased and generally tidied up, if you want to sort out the tests to your satisfaction please go ahead!

stdweird:
@jrha still thursday somewhere 😄

jrha:
It is in Tahiti!

PR #822, opened 2016-07-29
ncm-nfs: add cephfs share type to NFS component schema

Tom-Byrne:
Adding the ability to add CephFS network shares into fstab using ncm-nfs.

### DIFF
--- a/ncm-nfs/src/main/pan/components/nfs/schema.pan
+++ b/ncm-nfs/src/main/pan/components/nfs/schema.pan
@@ -15,7 +15,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device'     : string
     'mountpoint' : string
-    'fstype'     : string with match(SELF, '^(nfs4?|panfs|none)$')
+    'fstype'     : string with match(SELF, '^(nfs4?|(pan|ceph)fs|none)$')
     'options'    ? string
     'freq'       ? long(0..)
     'passno'     ? long(0..)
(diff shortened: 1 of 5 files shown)


jrha:
test this please

stdweird:
@Tom-Byrne this needs changes in the compoennt code as well, grep fro panfs and add cephfs 
@kwaegema @alvarosimon do you remember the issues we had with cephfs in fstab? is that working now (in recent enough el6/el7)

stdweird:
@kwaegema @alvarosimon or was that with ceph-fuse

Tom-Byrne:
@stdweird Ah, didn't think about that, done in b53ca1a (unless I've missed anything else).

stdweird:
@Tom-Byrne if you feel up to it, extend unittest
* add cephfs entry in https://github.com/quattor/configuration-modules-core/blob/master/ncm-nfs/src/test/resources/fstab_simple.pan
* hashref for cephfs simialr to https://github.com/quattor/configuration-modules-core/blob/master/ncm-nfs/src/test/perl/fstab.t#L142
* add regexp fro generated entry https://github.com/quattor/configuration-modules-core/blob/master/ncm-nfs/src/test/resources/regexps/fstab

stdweird:
ok to test

Tom-Byrne:
Finally got back around to extending the unit tests, should be all good to merge!

stdweird:
@Tom-Byrne thx a lot

PR #823, opened 2016-07-29
ncm-afsclt: use FileEditor not FileReader (for old CAF).

ajf8:
Found a small issue with @jouvin's refactoring - a missing use FileReader (it didn't use FileReader before). I think this only happens if you run ncm-afsclt alone (as we do at one point in our build) presumably because another components hasn't loaded it already.

I've changed the new use of FileReader to a FileEditor with explicit cancel so it's essentially the same, and consistent with another read in the component. Sorry, we have an old version of CAF on older builds, they will hopefully go away soon.

### DIFF
--- a/ncm-afsclt/src/main/perl/afsclt.pm
+++ b/ncm-afsclt/src/main/perl/afsclt.pm
@@ -122,7 +122,8 @@ sub Configure_Cache {
         $self->warn("Cannot determine current AFS cache size, changing only config file");
     }
 
-    my $afs_cacheinfo_fh = CAF::FileReader->new( $AFS_CACHEINFO, log => $self );
+    my $afs_cacheinfo_fh = CAF::FileEditor->new( $AFS_CACHEINFO, log => $self );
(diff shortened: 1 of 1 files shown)


stdweird:
@ajf8 what CAF release is this? adding CAF filereader to an older CAF build should be very easy.
(or even make `perl-CAF-backports` rpm wit only the filereader code?)

i'm not opposing this change, but not very happy with it either. are you aware of any other components that require older caf, ccm or ncm-ncd?

ajf8:
@stdweird - we have CAF 14.10.0 on some of our older builds which are hopefully going soon.

To be clear, the main reason for a PR was the missing use statement causing it to break when run alone. I would have simply added the use statement, but since we're one of the few users of this module and to make it consistent with the other file read, I changed it to FileEditor. 

I'm not going to be submitting PR's to change various FileReaders to FileEditors :) I think we have patched a small number of components locally for this (don't know which off the top of my head), and we can stop doing this soon.

jrha:
LGTM

PR #827, opened 2016-08-03
ncm-gpfs: reinstall gpfs update packages after base

kwaegema:
the gpfs component removes the update packages and installs the base packages, but didn't reinstall the update packages, therefore, this is not installed correctly in the ks-post-reboot phase

### DIFF
--- a/ncm-gpfs/src/main/perl/gpfs.pm
+++ b/ncm-gpfs/src/main/perl/gpfs.pm
@@ -310,8 +310,10 @@ sub remove_existing_rpms {
         } else {
             $self->error("Not removing unknown found rpm that matched gpfs.*:",
                          " $found (full: $foundfullname). \n");
-            $ok = 0;
+            $ok = 0; 
         };
+        # No need to remove other packages since we will stop after anyway
+        return 0 if !$ok;
     };
 
     $self->stopgpfs(1);
--- a/ncm-gpfs/src/main/perl/gpfs.pm
+++ b/ncm-gpfs/src/main/perl/gpfs.pm
@@ -51,6 +51,7 @@ use constant GPFSRPMS => qw(
                             ^gpfs.gskit$
                             ^gpfs.hdfs-protocol$
                             ^gpfs.hadoop-connector$
+                            ^gpfs.smb$
                            );
 
 my $compname = "NCM-gpfs";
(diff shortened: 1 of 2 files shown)


ncm-gpfs/src/main/perl/gpfs.pm
```diff
@@ -265,6 +263,26 @@ sub runcurl {
...
+sub reinstall_update_rpms {
+    my ($self, $config, $rpms) = @_;
+    my $ret = 1;
+    my $useyum =  $config->getValue("$mypath/base/useyum");
+    my @reinrpms = @$rpms;
+
+    if (scalar @reinrpms) {
+        if ($useyum) {
+            # for dependencies
+            $self->runyum($config, "install", @reinrpms) || return;
```
stdweird: don't return here. for both `runyum` and `runcmd`, do a `$ret = $self->run...`; and end this block with
```perl
if($ret) {
    $self->info("Rpms reinstalled");
} else {
   $self->error("Reinstalling rpms failed");
}
```

ncm-gpfs/src/main/perl/gpfs.pm
```diff
@@ -265,6 +263,26 @@ sub runcurl {
     return $output || 1;
 };
 
+sub reinstall_update_rpms {
+    my ($self, $config, $rpms) = @_;
+    my $ret = 1;
+    my $useyum =  $config->getValue("$mypath/base/useyum");
+    my @reinrpms = @$rpms;
+
+    if (scalar @reinrpms) {
```
stdweird: `scalar` is not needed

ncm-gpfs/src/main/perl/gpfs.pm
stdweird: maybe also fix this in `remove_existing_rpms`

stdweird:
@kwaegema some unittests would also be nice. i'm tagging this 16.10 due to feature freeze (and we're probably the only ones using this anyway)

stdweird:
@kwaegema squash the commit, also add unittests

jrha:
@kwaegema any response to @piojo's comments?

stdweird:
@kwaegema can you cleanup the commit messages? (at least prefix `ncm-gpfs`, and try to squash the remarks one if possible)

kwaegema:
This is now rebased on #911 to use same test resources. Please review/merge #911 first before this one

jrha:
@piojo any chance you can review this again now that it has been updated?

stdweird:
@kwaegema can you rebase on master? i'll review again after that

kwaegema:
@stdweird rebased

PR #828, opened 2016-08-04
ncm-aiiserver: support changing cachedir and some cleanup

stdweird:


ned21:
LGTM (not merging since I note 16.10 tag)

PR #833, opened 2016-08-09
ncm-download: code cleanup

stdweird:
Cleanup code to prepare more refactoring and new functionality


stdweird:
To avoid a new aii refactoring saga
tagged 16.10, but review asap, as i'm going to work on a new pr based on this one

ncm-download/src/main/perl/download.pm
```diff
@@ -257,64 +225,48 @@ sub download {
...
-    }
-
-    if ($capath) {
-        push @cmd, ("--capath", $capath);
-    }
-
-    if ($cert) {
-        push @cmd, ("--cert", $cert);
-    }
+    push(@cmd, map {("--$_", $opts{$_})} grep {$opts{$_}} qw(key cacert capath cert));
```
ned21: The old code was too verbose but this is too far the other way. How about this as a half way house:
```
foreach my $key (qw(key cacert capath cert)) {
   push(@cmd, "--$key ".$opts{$key});
}
```

ned21:
92cc14c does not mention the dependencies change in the commit log - is that an unintentional change from HTTP::Request::Common to LWP::Authen::Negotiate ?

ncm-download/src/main/perl/download.pm
```diff
@@ -61,113 +61,34 @@ sub Configure
...
-                    # TODO: ignoring with error?
-                    $self->error("$fn requested but no proto, ignoring");
-                    next;
-                }
-                $source = "$proto:$source";
-            }
-        }
-
-        my $min_age = $file->{min_age} || 0;
+        $file->{min_age} = 0 if ! $file->{min_age};
```
ned21: This change is not a pure refactoring - can the rationale be called out in the commit log please? (I agree it makes sense to remove the error() calls and fix them at the schema, but why not also fix the schema as part of this commit?)

ned21:
A few minor comments but mostly OK. I will try to make time in the next few days to grab the rpm generated by jenkins and install it to make sure this still works for us.

stdweird:
@ned21 `HTTP::Request::Common` was not used, and without `LWP::Authen::Negotiate` you cannot check the head timestamp using kerberos. 
`LWP::Authen::Negotiate` was added to CCM in 16.8 (https://github.com/quattor/CCM/issues/54#issuecomment-174166514 and https://github.com/quattor/CCM/pull/93/files#diff-005ad7732ca64307ff82c16c4e0884a4R38).
but it would be nice that you could verify the following on an older system: is `LWP::Authen::Negotiate` installed (via some other dependency), and if not, does checking the head work when the proxies are behind kerberos.

ned21:
It looks like the jenkins job does not generate rpms for components? I could only see a [POM file](https://jenkins1.ugent.be/job/NCM_Components-pull_requests/1452/org.quattor.cfg.module$download/artifact/org.quattor.cfg.module/download/16.6.1-SNAPSHOT/)

I ran out of time to test this today, will try again tomorrow. Regarding the dependency change I'm not so much worried about it causing an issue as much as it's not referenced in the commit log.  92cc14c only mentions the ENV changes, it does not mention the use statement changes. You should either make this a separate commit, or call it out in the log.

stdweird:
retest this please

stdweird:
retest this please

stdweird:
@ned21 i split the commit, the test now also produce rpms

ncm-download/src/main/perl/download.pm
```diff
@@ -1,247 +1,265 @@
...
+
+    my $success = 0;
+    if (@$proxyhosts && $file->{proxy}) {
+        my $attempts = scalar @$proxyhosts;
+        while ($attempts--) {
+            # take the head off the list and rotate to the end...
+            # we do this in order to avoid continually trying a dead
+            # proxy - once we've found one that works, we'll keep using it.
+            my $proxy = shift @$proxyhosts;
+            $self->debug(3, "Trying proxy $proxy for source $source and filename $fn");
```
ned21: This results in:

[DEBUG] Trying proxy ARRAY(0x3dc79c0) for source http://[redacted]/group.byname and filename /var/tmp/dsdbfiles/group.byname

which also fails.

The previous output was:
```
[DEBUG] Trying proxy [correct hostname]
[DEBUG] Processing file /var/tmp/dsdbfiles/rpc.bynumber from http://[redacted]/rpc.bynumber
[VERB] Executing command: /usr/bin/curl -s -R -f --create-dirs -o /var/tmp/dsdbfiles/rpc.bynumber http://$proxyhost/$path/rpc.bynumber
[VERB] Command options: stderr=SCALAR(0x2680ea8)
[DEBUG] We seem to have sucessfully downloaded a new copy of the file
[INFO] successfully processed file /var/tmp/dsdbfiles/rpc.bynumber
```

ned21:
I took a quick look at the generated rpm and noticed your name is missing from the developers list in pom.xml

ncm-download/src/main/perl/download.pm
ned21: $proxyhosts seems to be getting more deeply nested with each iteration

```
print Dumper($proxyhosts);
$VAR1 = [
          [
            [
              [
                [
                  [
                    [
                      'p3.x.x.x',
                      'p4.x.x.x',
                      'p5.x.x.x',
                      'p6.x.x.x'
                    ]
                  ]
                ]
              ]
            ]
          ]
        ];
```

but if I call Dumper just after this line:
```
    my $proxyhosts = $tree->{proxyhosts} || [];
```

it's nested correctly. I think the problem code is:

```
            if ($success) {
                @$proxyhosts = [$proxy, @$proxyhosts];
                last;
            } else {
                @$proxyhosts = [@$proxyhosts, $proxy];
            }
```
The move from an array to arrayref seems to have broken this construct. Changing to:
```
            if ($success) {
                #@$proxyhosts = [$proxy, @$proxyhosts];
                unshift  @$proxyhosts, $proxy;
                last;
            } else {
                #@$proxyhosts = [@$proxyhosts, $proxy];
                push @$proxyhosts, $proxy;
            }
```
works on RH6.

stdweird:
@ned21 the issue was the leading `@` on the left hand. i added preliminary improved unittests, they need some more work before this can be merged, but code seems functional from the test output logs. i'll refine the tests in a few hours

stdweird:
@ned21 unittests now check the proxyhosts.

stdweird:
@ned21 hmmmm, in `get_gss_token`, the tempdir goes out of scope, and probably is cleanup right away, thus the credential cache gets deleted immediately as well. what `CAF` version would you use on the nodes that would be updated to this? `CAF::Kerberos` can fix this.

stdweird:
@ned21 now tmpdir is kept around. i'll clean this all up by moving to `CAF::Kerberos` in later PR

ned21:
Unless there's a bug fix that means we need to run a newer component on older systems, from now on we should be able to deploy CAF+components from the same release. We've been just been running inconsistent versions while we try to "catch up" to recent releases.

I'll retest on Monday.

ned21:
Code tests fine. I guess we want to hold off merging this until after 16.8? Or should we ship this in 16.8 and then leave the changes for 16.10?

stdweird:
@ned21 both are fine by me. any future changes will be done in a new branch, which will be only for 16.10.

PR #837, opened 2016-08-11
ncm-afsclt: documentation cleanup

wdpypere:
- markdown does not like underscores in (what it thinks are) paths, escape them.
 - removed superfluous info.


stdweird:
@wdpypere when deleting old pod blocks with e.g. authors, make sure they are mentioned in the pom.xml

wdpypere:
@stdweird will do!

PR #840, opened 2016-08-12
ncm-amandaserver: documentation cleanup

wdpypere:
various consistency, punctuation and spelling fixes.


ned21:
Can you squash the two fixup commits into the first please so the commit history is kept easy to read?

wdpypere:
@ned21 done

PR #844, opened 2016-08-16
ncm-download: config-rpm is not part of the templates anymore

stdweird:
After #833, a reference to a previous template still existed


stdweird:
@ned21 this has to get merged (unittests passed because the removed template was still available via the core templates)

PR #851, opened 2016-08-16
ncm-filesystems: documentation cleanup

wdpypere:
Mainly removed the TODO list. I will create github issues for them.

### DIFF
--- a/ncm-filesystems/src/main/perl/filesystems.pod
+++ b/ncm-filesystems/src/main/perl/filesystems.pod
@@ -2,7 +2,6 @@
 # ${developer-info}
 # ${author-info}
 
-
 =pod
 
 =head1 DESCRIPTION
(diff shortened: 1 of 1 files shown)


ncm-filesystems/src/main/perl/filesystems.pod
```diff
@@ -12,33 +11,23 @@ to create and remove blockdevices without restarting or
...
+and creates new filesystems, if needed.
 
 A blockdevice is useful only for its ability to hold a
 filesystem. Blockdevices with no filesystems associated will not be
 created. If you want any such device, create a filesystem with
-"type"="none" and "mount"=false.
-
-=head2 To do list
-
-It will also remove filesystems and blockdevices that are not listed
```
ned21: I don't think this is just a TODO, it's also an important piece of functionality. The second sentence "For this to be safe, ..." can be removed. Should it be replaced by a link to the GH issue number?

wdpypere: Good point, I'll reinsert it as a warning note. I can insert the GH issue as well if you want but I don't think documentation is the correct place for that.

wdpypere:
@ned21 remark addressed.

jrha:
Could you reference this PR in the issues you created?

wdpypere:
@jrha will do

jrha:
Thanks!

PR #869, opened 2016-08-23
ncm-spma: yum: pass error_is_warn from update_pkgs_retry also to spare_dependencies

### LINKED ISSUE #868: ncm-spma: yum: missing cache dir failure
new installation with aii and spma+yum gives 

```
[WARN] dependencies of install candidates produced warnings: Could not match packages: Cannot access repository dir /var/cache/yum/x86_64/6.8/epel

[VERB] dependencies of install candidates output:
[WARN] Command output:
[ERROR] Failed dependencies of install candidates: Could not match packages: Cannot access repository dir /var/cache/yum/x86_64/6.8/epel
```

and next step (`--configure -all` is not called)

stdweird:
Provides workaround for #868

### DIFF
--- a/ncm-spma/src/main/perl/spma/yum.pm
+++ b/ncm-spma/src/main/perl/spma/yum.pm
@@ -432,11 +432,11 @@ sub expire_yum_caches
 sub apply_transaction
 {
 
-    my ($self, $tx, $tx_error_is_warn) = @_;
+    my ($self, $tx, $error_is_warn) = @_;
 
     $self->debug(5, "Running transaction: $tx");
     my $ok = $self->execute_yum_command([YUM_CMD], "running transaction", 1,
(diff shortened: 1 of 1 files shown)


jouvin:
This looks to me as a rather significant change and  I am not confortable judging it... The [issue](#868) is not providing a clear information about why this problem arised. We are installing new machines regularly and we never had this problem, including recently with CentOS 6.8...

stdweird:
@jouvin i have no idea what triggers the issue, and what actually fixes it. i changed the description a bit to reflect this

wrt the changes, they are very minimal, it just extends the yum commands that can fail without error in the multiple try update scenario.

ned21:
I think the change appears more complex because there's a refactoring (parameter rename) in the same commit as a functionality change. Please use two separate commits in future!

PR #875, opened 2016-08-29
ncm-iptables: clean up documentation

wdpypere:
- some structure work
 - cannot find extra documentation mentioned, so removed mentioning of it
 - some minor spelling fixes.

### DIFF
--- a/ncm-iptables/src/main/perl/iptables.pod
+++ b/ncm-iptables/src/main/perl/iptables.pod
@@ -10,14 +10,10 @@ iptables: Setup the IPTABLES firewall rules.
 =head1 DESCRIPTION
 
 The I<IPTABLES> component perform the setup of the
-B</etc/sysconfig/iptables> configuration file and restarts the
+C<< /etc/sysconfig/iptables >> configuration file and restarts the
 iptables service.
 
 =head1 SYNOPSIS
-
-Note: a detailed HOWTO for this component, including examples, is
-found in /usr/share/doc/ncm-iptables-2.3.13/ncm-iptables-howto.html and in the FIO twiki pages
-at https://twiki.cern.ch/twiki/bin/view/FIOgroup/Iptables.
 
 =over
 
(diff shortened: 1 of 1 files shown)


ncm-iptables/src/main/perl/iptables.pod
```diff
@@ -143,79 +137,55 @@ system for DDoS attacks it is worthwhile, for machines which are not
...
-
 =head1 FILES
 
-=head2 /etc/sysconfig/iptables:
+=head2 C<< /etc/sysconfig/iptables >>:
 
 I<IPTABLES> firewall configuration file policy.
 
-=head2 pro_declaration_component_iptables.tpl:
+=head2 pro_declaration_component_iptables.pan:
```
ned21: These two files also no longer exist?

ncm-iptables/src/main/perl/iptables.pod
```diff
@@ -10,15 +10,11 @@ iptables: Setup the IPTABLES firewall rules.
...
 
 The I<IPTABLES> component perform the setup of the
-B</etc/sysconfig/iptables> configuration file and restarts the
+C<< /etc/sysconfig/iptables >> configuration file and restarts the
 iptables service.
 
 =head1 SYNOPSIS
 
-Note: a detailed HOWTO for this component, including examples, is
-found in /usr/share/doc/ncm-iptables-2.3.13/ncm-iptables-howto.html and in the FIO twiki pages
```
ned21: Is there a version of the 2.3.13 rpm around anywhere? Perhaps we could check if that contains this file? Perhaps the source was lost in the migration to maven?

ncm-iptables/src/main/perl/iptables.pod
wdpypere: indeed, I'll remove them as well.

ncm-iptables/src/main/perl/iptables.pod
wdpypere: if we can get the old rpm, I'm happy to read anything that is in it. maybe it also still contains the 2 missing pan/tpl files.

ncm-iptables/src/main/perl/iptables.pod
```diff
@@ -31,25 +27,21 @@ default targets are supported.
...
-with multiple chains E.g. chain "LocalRules" is not the same as
+needs to - N.B. This means that you need to spell target names
+consistently and with identical capitalisation otherwise you will end up
+with multiple chains. E.g. chain "LocalRules" is not the same as
 "localrules".
 
 Duplicated entries in the component resource declaration are
 ignored. For each configured table, the chains are added to the
-B</etc/sysconfig/iptable> in order, the relative order among the rules
+C<< /etc/sysconfig/iptable >> in order, the relative order among the rules
```
stdweird: `s` missing?

ncm-iptables/src/main/perl/iptables.pod
stdweird: maybe ask around on the mailing list (we don't have these)

wdpypere: the oldest I can find in our repos is 13.2.0 which does not contain any of the mentioned files.

wdpypere: OK, I'll ask on the mailing list.

stdweird: the declaration file is now schema, but i have no clue what the functions are about.

ncm-iptables/src/main/perl/iptables.pod
wdpypere: Yes! thx. I'll fix it.

wdpypere:
@ned21 the old examples were recovered and added again.

ned21:
Thanks, @wdpypere. I don't think it really makes sense to split these two changes across two commits so would you be OK to squash them into a single commit?

wdpypere:
@ned21 squashed the commits.

PR #881, opened 2016-09-01
ncm-metaconfig: clean up documentation

wdpypere:
minor clean up

### DIFF
--- a/ncm-metaconfig/src/main/perl/metaconfig.pod
+++ b/ncm-metaconfig/src/main/perl/metaconfig.pod
@@ -1,9 +1,6 @@
-=encoding utf8
-
 # ${license-info}
 # ${developer-info}
 # ${author-info}
-# ${build-info}
 
 =head1 NAME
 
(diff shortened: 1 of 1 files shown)


ncm-metaconfig/src/main/perl/metaconfig.pod
```diff
@@ -66,40 +63,40 @@ it uses will generate them. The preamble attribute will be written out
...
 A free-form structure describing the valid entries for the
 configuration file. It is recommended to define another type for each
 config file, and bind it to these contents, to get the best
 validation.
 
-=item * element
+=item * C<element>
 
-Predefined conversions from L<EDG::WP4::CCM::TextRender>:
+Predefined conversions from C<< EDG::WP4::CCM::TextRender >>:
```
ned21: I thought L<> was supposed to link to the EDG::WP4::CCM::TextRender page - does it not work? Or does C<<>> do the same thing?

wdpypere: L<> indeed makes a link, but not as we would expect it. 
 - for man pages this is just stripped out and becomes text  (see ```man NCM::Component::metaconfig```)
 - if people would use perldoc to look at this  (```perldoc NCM::Component::metaconfig```) this should work as you expect (I think) but we don't ship the pod files to the system so it means nothing. (if this is supposed to happen then it is a bug).
```
[wdpypere@gastly ~]$ perldoc NCM::Component::metaconfig
No documentation found for "NCM::Component::metaconfig".
```
 - for web related content this gets translated to https://metacpan.org/pod/EDG::WP4::CCM::TextRender which does not exist and fails the htmlproofer tests as it should.

So putting it between C<<>> will make sure it stands out in a page and render correctly without any regression afaik.
In the future it would be good that we link correctly but that needs some decent work to the script that generates the docs.

PR #884, opened 2016-09-01
ncm-metaconfig: nginx fix ssl conf

### LINKED ISSUE #876: ncm-metaconfig: nginx: ssl.tt file generates an empty ssl_client_certificate value
`ca` is not mandatory in the nginx ssl schema:

>...
>"ca"  ? string
>...

But it is included by the ssl.tt file in any case and it generates a malformed nginx.conf file:

>...
> ssl_certificate /etc/ipa/quattor/certs/host.pem;
> ssl_client_certificate ;
> ssl_certificate_key /etc/ipa/quattor/keys/host.key;
>...

From systemd:
>nginx: [emerg] invalid number of arguments in "ssl_client_certificate" directive in /etc/nginx/nginx.conf:36

`ssl_client_certificate`  value should be included only if the ca is set.

alvarosimon:
* Fix #876 : ssl.tt file generates an empty ssl_client_certificate value

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/nginx/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/nginx/pan/schema.pan
@@ -26,9 +26,9 @@ type httpd_ssl = {
     "protocol" : cipherstring[] = list("TLSv1")
     "certificate" : string
     "key" : string
-    @{ssl_client_certificate specifies a file with trusted CA certificates in
-    the PEM format used to verify client certificates and OCSP responses if
-    ssl_stapling is enabled}
+    @{ca sets ssl_client_certificate which specifies a file with trusted
+    CA certificates in the PEM format used to verify client certificates and OCSP
+    responses if ssl_stapling is enabled}
     "ca"  ? string
     "certificate_chain_file" ? string
     "revocation_file" ? string
--- a/ncm-metaconfig/src/main/metaconfig/nginx/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/nginx/pan/schema.pan
@@ -16,6 +16,9 @@ type basic_ssl = {
     "require" ? string
 };
 
+@{
+    SSL nginx configuration
+}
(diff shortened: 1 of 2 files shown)


stdweird:
@ned21 add a unittest for the ca (updating a template with the  and the regex).
also why is the `ssl_client_certificate` configured with a `ssl.ca` attribute? (what does the CA have to do with the "client"? at least add some annotiation to the schema to clarify this)

ned21:
@stdweird Did you mean to address those remarks to @alvarosimon instead of me?

stdweird:
@ned21 oops, no, ofcourse not
@alvarosimon see comment above

alvarosimon:
@stdweird no problem :+1:

alvarosimon:
@stdweird the ca (`ssl_client_certificate`)  unittest and regexp is already there, but taking a closer look to the schema and tt files it seems that some ssl values from the schema are not used at all from the tt files or the names have been changed, so yes it's a bit confusing:

http://nginx.org/en/docs/http/ngx_http_ssl_module.html

So I don't know if we should also fix that in this PR or open a new one...

ncm-metaconfig/src/main/metaconfig/nginx/pan/schema.pan
```diff
@@ -16,13 +16,19 @@ type basic_ssl = {
...
+    SSL nginx configuration
+}
 type httpd_ssl = {
     include basic_ssl
     "active" : boolean = true
     "ciphersuite" : cipherstring[] = list("TLSv1")
     "protocol" : cipherstring[] = list("TLSv1")
     "certificate" : string
     "key" : string
+    @{ssl_client_certificate specifies a file with trusted CA certificates in
```
stdweird: should be `ca sets ssl_client_certificate which specifies ...`

piojo:
Unit test plz?

alvarosimon:
@piojo the current base regex is already checking `ssl_client_certificate` output

PR #885, opened 2016-09-02
ncm-ccm: add missing "use CAF::FileReader"

ajf8:
ncm-ccm: add missing "use CAF::FileReader"

Component fails if the noquattor file is present, and some other component has not already loaded FileReader (eg. only this compoent was run).

None.


ned21:
@jrha - Trivial bug fix, feel free to push back to 16.10 is it's too late for 16.8

stdweird:
@ajf8 thanks for catching this.

PR #888, opened 2016-09-05
ncm-opennebula: Use resource dictionaries instead of lists

alvarosimon:
This PR changes OpenNebula users/groups/vnets/datastores resources schema to use dictionaries

* Fix: #834 : Change some schema resource types to dictionaries instead lists

PR #890, opened 2016-09-06
ncm-opennebula: Include OneFlow support

### LINKED ISSUE #755: ncm-opennebula: Include OneFlow support
At this moment `oneflow-server.conf` file is provided as a default conf file from  OneFlow package. We should manage this conf file from Quattor as we did for `oned.conf`and `sunstone-server.con` services, to change for example these default values:

>:host: 127.0.0.1
>:port: 2474

alvarosimon:
This PR includes OneFlow configuration support and unittest.

 * Fix #755 : Include OneFlow support

**NOTE: This PR depends on #888 that should be reviewed/merged first**


alvarosimon:
It was tested in altaria

PR #895, opened 2016-09-06
ncm-gpfs: add new hdfs-protocol package to GPFSRPMS

kwaegema:
ncm-gpfs does not need to fail because of unrecognized package gpfs.hdfs-protocol


stdweird:
LGTM

PR #907, opened 2016-09-07
ncm-resolver: clean up documentation

wdpypere:
Did some minor clean up.

@jrha @ned21 this documentation contains a disclaimer (which I probably can't remove) and a reference to [http://eu-datagrid.web.cern.ch/eu-datagrid/license.html](http://eu-datagrid.web.cern.ch/eu-datagrid/license.html).
to adhere to point 2 I can add an acknowledgement in the pod stating:
```
"This product includes software developed by the EU DataGrid (http://www.eu-datagrid.org/)."
```

But I don't know if this actually is enough to be compliant as most of the reference and disclaimer will never show up on a website or documentation as they are in comments.

Thoughts?

### DIFF
--- a/ncm-resolver/src/main/perl/resolver.pod
+++ b/ncm-resolver/src/main/perl/resolver.pod
@@ -2,32 +2,6 @@
 # ${developer-info}
 # ${author-info}
 
-#
-# This program is free software; you can redistribute it and/or modify
-# it under the terms of the EU DataGrid Software License.  You should
-# have received a copy of the license with this program, and the license
-# is published at http://eu-datagrid.web.cern.ch/eu-datagrid/license.html.
-#
-# THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS
-# CONTRIBUTED IN CONNECTION WITH THIS PROGRAM.
-#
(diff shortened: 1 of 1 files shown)


jrha:
AFAIK the disclaimer could actually be removed, as it applies to contributions rather than the code base as a whole.

wdpypere:
OK. If nobody disagrees before next week or so, I'll remove it.

ncm-resolver/src/main/perl/resolver.pod
```diff
@@ -41,41 +41,35 @@ NCM::resolver - NCM resolver configuration component
...
 If dnscache is used, then dnscache will be restarted on any change.
 If DNS resolution fails after making the change, then resolv.conf
-is left in it's previous state.
-Returns
-error in case of failure.
-
-=item Unconfigure()
-
-not available.
+is left in it's previous state. Returns error in case of failure.
```
ned21: Remove the sentence "Returns error in case of failure" because that does not make sense and a quick look at the code shows it's not true. :( In most, but not all, cases it returns 0 after an error, and >0 for success... I think the component needs some modernisation.

wdpypere:
@jrha remarks addressed.

PR #921, opened 2016-09-26
ncm-download: fix GSSAPI on GET requests.

ajf8:

The KRB5CCNAME environment variable is set using local scope inside an
if block. This means the environment variable is not set for the curl
invocation outside, and as such GSSAPI fails.

Declare $ENV{KRB5CCNAME} as local variable outside in the right scope.

None.


stdweird:
@ajf8 thx for this fix

PR #922, opened 2016-09-27
ncm-ccm: support purge_time config setting.

ajf8:

We would like to start using the purge_time config setting but it's missing from the schema.

None.


stdweird:
@ajf8 can you check if there are any other missing options (or confirm there aren't?)
LGTM otherwise

PR #750, opened 2016-04-11
ncm-freeipa: initial support

stdweird:
ncm-freeipa adds support for configuring (parts of) FreeIPA (http://www.freeipa.org) server and client

PR #925, opened 2016-09-30
[WIP] ncm-download: support default Net::HTTPS SSL class

stdweird:
The default on EL5/6/7 is `IO::Socket::SSL`, and on EL6/7 you can also start using `verify_hostname`

### DIFF
--- a/ncm-download/src/main/perl/download.pm
+++ b/ncm-download/src/main/perl/download.pm
@@ -245,7 +245,13 @@ sub get_remote_timestamp
     $lwp->timeout($opts{head_timeout}) if (defined($opts{head_timeout}));
     my $head = $lwp->head($source);
 
-    return $head->headers->last_modified();
+    if ($head->is_success()) {
+        return $head->headers->last_modified();
+    } else {
+        # Status line starts with code
+        $self->debug(1, "head for $source failed with ", $head->status_line());
+        return;
+    }
 }
 
 
--- a/ncm-download/src/main/perl/download.pm
+++ b/ncm-download/src/main/perl/download.pm
@@ -7,8 +7,49 @@ our $EC = LC::Exception::Context->new->will_store_all;
 use File::Temp qw(tempdir);
 use CAF::Process;
 use POSIX;
+use Readonly;
 
-use LWP::UserAgent;
+# Lexical scope for Readonly set in BEGIN{}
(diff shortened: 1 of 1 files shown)


stdweird:
This is WIP, and needs more testing. It's a requirement for #750

ncm-download/src/main/perl/download.pm
```diff
@@ -20,6 +61,126 @@ $NCM::Component::download::NoActionSupported = 1;
...
+
+sub _lwp_ua
+{
+    my ($self, $method, $args, %opts) = @_;
+
+    # This is a mess.
+
+    # Set this again; very old Net::HTTPS (like in EL5) does not set the class
+    # on the initial import in the BEGIN{} section
+    $ENV{PERL_NET_HTTPS_SSL_SOCKET_CLASS} = $_default_https_class;
```
piojo: Perhaps these should be `local` maybe? So that other incantations and aliases are not affected by this environment variable?

I know, it would be silly to have different classes and CAs for different downloading portions, but just in case...

ncm-download/src/main/perl/download.pm
```diff
@@ -20,6 +61,126 @@ $NCM::Component::download::NoActionSupported = 1;
...
+        my $ssl_opts = {
+            verify_hostname => $verify_hostname,
+        };
+
+        $ssl_opts->{SSL_ca_file} = $opts{cacert} if $opts{cacert};
+        $ssl_opts->{SSL_ca_path} = $opts{cadir} if $opts{cadir};
+        $ssl_opts->{SSL_cert_file} = $opts{cert} if $opts{cert};
+        $ssl_opts->{SSL_key_file} = $opts{key} if $opts{key};
+
+        $self->debug(3, "Using LWP::UserAgent ssl_opts ", join(" ", map {"$_: ".$ssl_opts->{$_}} sort keys %$ssl_opts));
```
piojo: String interpolation plz.

```perl
join(" ", map{"$_: $ssl_opts->{$_}"} sort...);
```

ncm-download/src/main/perl/download.pm
stdweird: environment protection is added in https://github.com/quattor/ncm-ncd/pull/88/commits/20b16fa1617bb87a83b943ea999459390619834a, part of the https://github.com/quattor/ncm-ncd/pull/88

jrha:
Replaced by #984.

PR #933, opened 2016-10-04
ncm-useraccess: clean up documentation

wdpypere:
minor fixing.

### DIFF
--- a/ncm-useraccess/src/main/perl/useraccess.pod
+++ b/ncm-useraccess/src/main/perl/useraccess.pod
@@ -2,7 +2,6 @@
 # ${developer-info}
 # ${author-info}
 
-
 =pod
 
 =head1 DESCRIPTION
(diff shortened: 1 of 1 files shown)


ncm-useraccess/src/main/perl/useraccess.pod
```diff
@@ -163,10 +153,7 @@ may consume lots of disk space.
 
 =head1 EXAMPLES
 
-You still have questions? I hope the following examples will make it
-clearer.
-
-=head2 One user who can be sued
+=head2 One user who can be sued:
 
 This example illustrates how tricky ACLs can get:
```
piojo: While you are at that, remove this section.  PAM ACLs should be avoided, and the only user has long left the community.

Or, at least, put a bold text mentioning that this feature should be avoided at all costs.

ncm-useraccess/src/main/perl/useraccess.pod
```diff
@@ -233,37 +220,37 @@ We all know how evil Mr Burns is. So, let's say he wants full control
...
-	"acls", list ("system-auth") # Woops! now Homer can't log-in!
-	);
+    "kerberos4", list (nlist (
+        "realm", "SPRINGFIELD.COM",
+        "principal", "homer",
+        "instance", "another_silly_project")),
+    "acls", list ("system-auth") # Woops! now Homer can't log-in!
+    );
 
  "/software/components/@COMPS@/users/marge/roles" = list (
```
piojo: `@COMPS@` is not expanded anymore.  Can you please replace it with `useraccess` here?

wdpypere:
@piojo remarks addressed.

PR #935, opened 2016-10-05
ncm-accounts: Add users and groups used by RabbitMQ, Memcached and MongoDB

apdibbo:
Add users and groups required by RabbitMQ, Memcached and MongoDB to these lists. 

This is in response to https://github.com/quattor/template-library-openstack/pull/30#issuecomment-251671061

I have only added the users and groups for the general tools rather than anything OpenStack specific.


ned21:
Is it scalable to have every application add their users to these "standard" lists? I suppose if memcached, rabbitmq and mongodb are shipped by the OS then it would make sense to put them here, but otherwise should they not be added to the template that configures those components?

apdibbo:
This PR was at the request of @jouvin as these are "standard" components which are available through the standard OS repos

apdibbo:
I have made the changes requested

ned21:
@apdibbo - makes sense, thanks for clarifying

PR #956, opened 2016-10-24
ncm-accounts, ncm-amandaserver, ncm-cups, ncm-iptables, ncm-nscd: Use legacy type for simple yes/no strings

### LINKED ISSUE #752: ncm-*: Migrate remaining yes/no strings to legacy_binary_affirmation_string type
As discussed in #749, there are several components that use a similar string but defined directly as a string and match rather than a type, e.g.
```pan
'create_home' ? string with match (SELF,'yes|no')
```
These should be migrated to `legacy_binary_affirmation_string`.

jrha:
Fixes #752 as described therein.


stdweird:
@jrha not sure my remarks are worth the trouble...

jrha:
What do you mean?

jouvin:
Thanks for this cleanup!

PR #911, opened 2016-09-08
ncm-gpfs: add ccr support

kwaegema:
gpfs with ccr needs some more files.
Will add some test for this, already made PR for qmmpack

PR #957, opened 2016-10-26
ncm-gpfs: allow to run mmsdrfrestore after each configfetch

stdweird:
Requires #911


stdweird:
I've assigned this to @kwaegema. Should cherry pick this in #911 when he's back from holiday
We need this hotfix and it has to be a PR to be able to make rpms.

kwaegema:
included in #911

PR #965, opened 2016-11-15
ncm-systemd: unmask masked units before state change

### LINKED ISSUE #962: ncm-systemd 16.10-rc1: a masked service cannot be reenabled
As of ncm-systemd 16.10.0-rc1, `masked` is considered as a service state (like `enabled` and `disabled`) when they are different things in `systemd` (an `enabled` or `started` service can be masked without impacting the enabled or started state). IMO, I think the choice made in `ncm-systemd` is good: in particular if you defined the `state` to `masked` and `startstop` to `true`,  the service is stopped and disabled in addition to be masked. The problem is that to unmask a masked service, the only possibility in `ncm-systemd` is to define its state to `enabled`. But `systemctl` doesn't allow to enable a masked service   (it does allow to disable it). 

If we want to keep the current `ncm-systemd` behaviour where we consider `masked` as a state, setting the `state` to `enabled` should unmask the service if necessary. Currently, the only way to reeanble the service requires to execute ("manually") `systemctl unmask` on the machine.

### LINKED ISSUE #964: ncm-systemd: no error for possible missing units
altough the component reports 'no error will be logged' , an error from failing `systemctl` call is reported
```
[DEBUG] systemctl_show for name ipsec.service
[VERB] Getting output of command: /usr/bin/systemctl --no-pager --all show -- ipsec.service
[ERROR] systemctl show failed (cmd /usr/bin/systemctl --no-pager --all show -- ipsec.service; ec 256) with output Unit ipsec.service could not be found.

[DEBUG] Unit ipsec.service not listed and has no show data.
[DEBUG] Unit ipsec.service is in possible_missing (and not found). No error will be logged.
[VERB] Trying to add details of unit ipsec.service but no entry in cache after adding all units and unitfiles. Ignoring this unit.
```

stdweird:
Fixes #962 (and also fixes #964)


stdweird:
@jouvin feel free to try https://jenkins1.ugent.be/job/NCM_Components-pull_requests/1640/org.quattor.cfg.module$systemd/artifact/org.quattor.cfg.module/systemd/16.10.0-rc2-SNAPSHOT/systemd-16.10.0-rc2-SNAPSHOT.rpm

stdweird:
tested the handling of errors and unmasking, works as expected

stdweird:
@jrha on a rhel73 beta system, this PR seems to be needed. i'll wait to try on centos7.3, but be prepared to make  a 16.10.1 with this PR for el7.3

jouvin:
@stdweird @jrha I had not time yet to test this PR but if there is a change that it will be required for EL 7.3, why not to merge it in 16.10? Better not to create a problem, I had a quick look to the change and it is pretty limited. And there is a unit test!

PR #969, opened 2016-11-23
ncm-network: Remove useless log message

jrha:
Removes the utterly useless "Some more debug info" message that irritates me so much.

We should probably do something about the multi line message after it as well, but this will do for now.


stdweird:
milestone?

stdweird:
retest this please

PR #971, opened 2016-11-24
ncm-aiiserver: do not add (empty) attributes for unused CCM properties

### LINKED ISSUE #970: aii-server 16.10.0-rc3: keyword without values added
In 16.10.0 release candidates (including RC3), ncm-aiiserver produces a `aii-shellfe.conf` file with keywords without a value, causing aii-shellfe to fail. This includes: 
```
ca_dir =
ca_file =
cert_file =
key_file =
``` 

This is a showstopper for release [16.10.0](https://github.com/quattor/release/issues/269).

stdweird:
Fixes #970


stdweird:
@jouvin this should work

jouvin:
Thanks. I'm testing it...

jouvin:
Thanks for the quick fix. I confirm that it works.

jrha:
Awesome.

jrha:
Tests just failed on the OpenNebula component, I suspect it is the same error as I was seeing with the Perl library version.

PR #967, opened 2016-11-18
ncm-opennebula: Include OpenNebula labels support

alvarosimon:
**NOTE:** This PR requires `Net-OpenNebula-0.309.0` to work

* Fix #959 : Include labels support
* Fix #966 : fix `Net::OpenNebula` `0.309.0`  host rpc_history

PR #976, opened 2016-11-29
ncm-download: download to temporary file.

### LINKED ISSUE #975: ncm-download: can leave partial files
I'm looking to an issue we encountered involving ncm-download. It appears that a rather important file which we fetch with ncm-download did not download successfully, and a partial file was left. Although ncm-download reported the error, another process came along and blindly made use of the file, resulting in bad things happening.

While we will probably improve the process which consumes that file, it seems like some protection in ncm-download would be useful too, and I thought perhaps we could discuss that here before I send a pull request.

As we are using it currently, curl truncates the file it is writing to if it already exists, so I guess this could be problem if a process tries to use the file while a successful download is taking place, also.

How about downloading .$filename.part (or some other variant... should this be configurable?) in the same directory, and doing rename()? Similar to a web browser.

I thought about using the tempdir we use for kerberos credentials currently, but then it may have to move() across filesystems.

Should this be configurable and what should the default be?

ajf8:

See issue #975 for discussion. Summary: downloads should be atomic by downloading to a temporary file and renaming. An unsuccessful/partial download could cause a problem for a process that tries to use it later or during the download even in the successful case.

.$filename.part files will be created in the target directory temporarily. This could potentially cause a problem for something which finds these files and reacts badly. It also means more disk is used temporarily during the download, if the file already exists.

Fixes #975.

### DIFF
--- a/ncm-download/src/main/perl/download.pm
+++ b/ncm-download/src/main/perl/download.pm
@@ -5,6 +5,7 @@ use parent qw(NCM::Component CAF::Path);
 our $EC = LC::Exception::Context->new->will_store_all;
 
 use File::Temp qw(tempdir);
+use File::Basename;
 use CAF::Process;
 use POSIX;
 
@@ -248,11 +249,30 @@ sub get_remote_timestamp
     return $head->headers->last_modified();
 }
 
+sub _cleanup_tmpfile
+{
+    my ($self, $tmpfn) = @_;
+    if (!defined($self->cleanup($tmpfn))) {
+        $self->error("Unable to delete temporary file $tmpfn: $self->{fail}");
+        return;
+    }
+    return 1;
+}
+
+# the temporary file is created in the same directory that the real file should
+# be in, with a dot prefix and .part suffix, so CAF::Path::move can move into
+# place atomically (because it is on the same filesystem).
+sub _tempfile
+{
+    my ($self, $fn) = @_;
+    return dirname($fn) . "/." . basename($fn) . ".part";
+}
 
 # Single attempt, actual transfer
 sub retrieve
 {
     my ($self, $fn, %opts) = @_;
+    my $tmpfn = $self->_tempfile($fn);
 
     my $source  = $opts{href};
     my $timeout = $opts{timeout};
@@ -265,7 +285,7 @@ sub retrieve
     $self->debug(1, "Retrieving file $fn from $source");
 
     local $ENV{KRB5CCNAME};
-    my @cmd = (qw(/usr/bin/curl -s -R -f --create-dirs -o), $fn);
+    my @cmd = (qw(/usr/bin/curl -s -R -f --create-dirs -o), $tmpfn);
 
     if ($proxy) {
         $source =~ s{^([a-z]+)://([^/]+)/}{$1://$proxy/};
@@ -291,7 +311,7 @@ sub retrieve
     # Get timestamp of any existing file, defaulting to zero if the
     # file doesn't exist
     my $timestamp_existing = 0;
-    if (-e $fn) {
+    if ($self->file_exists($fn)) {
         $timestamp_existing  = (stat($fn))[9];
     }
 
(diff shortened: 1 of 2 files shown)


ncm-download/src/main/perl/download.pm
```diff
@@ -253,6 +254,7 @@ sub get_remote_timestamp
 sub retrieve
 {
     my ($self, $fn, %opts) = @_;
+    my $tmpfn = dirname($fn) . "/." . basename($fn) . ".part";
```
jrha: Is it better to join file paths with the functionality provided by `File::Spec`?

ncm-download/src/main/perl/download.pm
```diff
@@ -306,12 +308,25 @@ sub retrieve
                 if (!POSIX::WIFEXITED($?) || POSIX::WEXITSTATUS($?) != 0) {
                     # Only warn as there may be attempts from multiple proxies.
                     $self->warn("curl failed (". ($?>>8) ."): $errs\nCommand = $proc");
+                    if (-e $tmpfn) {
```
stdweird: can you use `CAF::Path`? so we can unittest it?
also, you should also do this before the download

ncm-download/src/main/perl/download.pm
```diff
@@ -306,12 +308,25 @@ sub retrieve
...
                     if ($CAF::Object::NoAction) {
                         $self->debug(1, "We have sucessfully pretended to retrieve a new copy of the file $fn");
                     } else {
-                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn");
+                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn to $tmpfn");
+                        if (rename($tmpfn, $fn)) {
+                            $self->debug(1, "Succesfully renamed temporary file $tmpfn to $fn");
+                        } else {
+                            $self->error("Unable to rename temporary file $tmpfn to $fn: $!");
+                            unless (unlink($tmpfn)) {
```
stdweird: use the same `if (-e ...)` logic as above; and even better, make a small private method for it

ncm-download/src/main/perl/download.pm
ajf8: I think (maybe wrong) I've seen this done in a lot of places, so didn't see much point, and it doesn't look like File::Spec is used currently. Happy to add it, though.

ncm-download/src/main/perl/download.pm
ajf8: Why before? The file should not be there, but if it is, curl will truncate it. I can't think why we would want to remove it before.

stdweird:
@ajf8 lets not assume that the component crashes in the middle or people interrupt it or powercycle etc etc....
the temp file should exist prior to donwloading it, so why not make sure of it? (esp now that you have a easy method to use).

ajf8:
@stdweird it would work fine if the component crashes.

the temp file should *not* exist prior to the download, but I can't think why it matters if it is, because curl will truncate it. I'm happy to add it though if you think it's useful to check.

stdweird:
@ajf8 please do
@jrha tests are failing waiting for #967 to be merged, can you do the honours?

ajf8:
@stdweird done

ncm-download/src/main/perl/download.pm
```diff
@@ -248,11 +249,29 @@ sub get_remote_timestamp
     return $head->headers->last_modified();
 }
 
+sub _cleanup_tmpfile
```
stdweird: you could use `$self->cleanup` from `CAF::Path` instead of the method (i wasn't aware `CAF::Path` was already assumed).
if you want an error you can do
```perl
$self->error("Unable to delete temporary file $tmpfn: $self->{fail}") if (! defined($self->cleanup($tmpfn)));
```
instead this method

ncm-download/src/main/perl/download.pm
```diff
@@ -301,19 +320,28 @@ sub retrieve
...
                         $self->debug(1, "We have sucessfully pretended to retrieve a new copy of the file $fn");
                     } else {
-                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn");
+                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn to $tmpfn");
+                    }
+                    if ($self->move($tmpfn, $fn)) {
+                        $self->debug(1, "Succesfully renamed temporary file $tmpfn to $fn");
+                        return 1;
+                    } else {
+                        $self->error("Unable to rename temporary file $tmpfn to $fn: $!");
```
stdweird: use `$self->{fail}` instead of `$!`

ncm-download/src/main/perl/download.pm
```diff
@@ -301,19 +320,28 @@ sub retrieve
...
+                    $self->_cleanup_tmpfile($tmpfn);
                     return 0; # Curl barfed, so we return.
                 } else {
                     if ($CAF::Object::NoAction) {
                         $self->debug(1, "We have sucessfully pretended to retrieve a new copy of the file $fn");
                     } else {
-                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn");
+                        $self->debug(1, "We seem to have sucessfully retrieved a new copy of the file $fn to $tmpfn");
+                    }
+                    if ($self->move($tmpfn, $fn)) {
```
stdweird: we should document here (and possibly in CAF::Path) that `$self->move` use `FIle::Copy::move` which uses a `rename`, which is atomic within the same filesystem (which is the case for file in the same directory) so this does what we want safely

ncm-download/src/main/perl/download.pm
```diff
@@ -248,11 +249,29 @@ sub get_remote_timestamp
...
+            return 0;
+        }
+    }
+    return 1;
+}
+
+sub _tempfile
+{
+    my ($self, $fn) = @_;
+    return dirname($fn) . "/." . basename($fn) . ".part";
```
stdweird: add a comment that the tempfile must be in the same dir as the destination (so the move/rename is atomic)
it's to prevent other devs from causing serious issues when they want to rename the tempfile to something else

jrha:
test this please

ncm-download/src/main/perl/download.pm
ajf8: Done, put this inside the existing method to avoid repetition.

Issue #434, opened 2015-03-04
Configuration modules that can be replaced by usage of metaconfig


jrha:
At the 19th (Grenoble) workshop we reviewed the core modules to see which could be trivially replaced by metaconfig, the table below summarises our decisions.

This issue should be closed when all the decisions below are implemented (this may take a long time!).

Module | Descision
----- | -----
ncm-accounts | keep
ncm-afsclt | keep
ncm-aiiserver | metaconfig
ncm-altlogrotate | textrender
ncm-amandaserver | keep
ncm-authconfig | keep - very complex
ncm-autofs | keep - very complex
ncm-ccm | keep
ncm-cdp | metaconfig
ncm-ceph | keep - runs ceph commands
ncm-chkconfig | keep - changes state
ncm-cron | keep - cleans up
ncm-cups | keep - edits files
ncm-directoryservices | archive
ncm-dirperm | keep - changes permissions
ncm-download | keep - download logic
ncm-etcservices | keep - edits files - migrate to CAF::FileEditor
ncm-filecopy | keep
ncm-filesystems | keep
ncm-fmonagent | maybe archive - check if used by MS, if yes then metaconfig
ncm-fsprobe | archive - used metaconfig
ncm-fstab | keep - edits files
ncm-ganglia | metaconfig
ncm-gmetad | metaconfig
ncm-gmond | metaconfig
ncm-gpfs | keep
ncm-grub | keep
ncm-hostsaccess | metaconfig
ncm-hostsfile | keep - edits files
ncm-icinga | possibly metaconfig
ncm-interactivelimits | keep - edits files
ncm-ipmi | keep - communicates with bmc units
ncm-iptables | keep - very complex
ncm-ldconf | textrender - reloads ld cache
ncm-libvirtd | metaconfig
ncm-mcx | archive
ncm-metaconfig | keep - infinite recursion
ncm-modprobe | keep - runs mkinitrd (hard coded support for 2.6 kernels)
ncm-mysql | keep - complex
ncm-nagios | possibly metaconfig
ncm-named | keep - edits files, but remove support for resolv.conf
ncm-network | keep - very complex
ncm-nfs | keep - edits files
ncm-nrpe | metaconfig
ncm-nsca | metaconfig - would need update for EL7
ncm-nscd | keep - edits files and rolls back
ncm-nss | keep - edits files, but needs documentation
ncm-ntpd | metaconfig - pay attention to service name on linux/solaris
ncm-ofed | metaconfig
[...]

PR #983, opened 2016-12-07
ncm-ssh, ncm-metaconfig httpd: stricter checking of ssh config opts

JensTimmerman:
currently every string is allowed as an option for `ciphers`, `MACs` and `KexAlgorithms`
Here I've added some stricter checking  (and only allow recommended ciphers, macs and algorithms, so no CBC and sha1 (for non mac))


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -30,6 +75,7 @@ type ssh_core_options_type = {
     "SendEnv"                           ? legacy_binary_affirmation_string
     "TCPKeepAlive"                      ? legacy_binary_affirmation_string
     "XAuthLocation"                     ? string
+    "KexAlgorithms"						? ssh_kexalgorithms[]
```
stdweird: fix whitespace

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -10,26 +10,72 @@ include 'pan/types';
...
 type ssh_preferred_authentication = string with match(SELF, '^(gssapi-with-mic|hostbased|publickey|keyboard-interactive|password)$');
 
+type ssh_ciphers = string with is_valid_ssh_cipher(SELF);
+type ssh_hostkeyalgorithms = string with match(SELF, "^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521)|(ssh-rsa-cert-v01|ssh-dss-cert-v01|ecdsa-sha2-nistp256-cert-v01|ecdsa-sha2-nistp384-cert-v01|ecdsa-sha2-nistp521-cert-v01|ssh-rsa-cert-v00|ssh-dss-cert-v00|ssh-ed25519-cert-v01)@openssh.com)$");
+type ssh_kbdinteractivedevices = string with match (SELF, "^(bsdauth|pam|skey)$");
+type ssh_kexalgorithms = string with match (SELF, "^(diffie-hellman-group-exchange-sha256|ecdh-sha2-nistp(256|384|521)|curve25519-sha256@libssh.org)$");
+type ssh_MACs = string with is_valid_ssh_MAC(SELF);
+
+function is_valid_ssh_MAC = {
+    match(ARGV[0], "^(hmac-(sha1|sha2-256|sha2-512|ripemd160)|(hmac-ripemd160|umac-64|umac-128|hmac-sha1-etm|hmac-sha2-256-etm|hmac-sha2-512-etm|hmac-ripemd160-etm|umac-64-etm|umac-128-etm)@openssh.com)$");
```
kwaegema: no need to specify hmac-ripemd160, already matched by hmac-(..|ripemd160)
also add ``(-etm)?`` to the first part of the regex to get rid of all the seperate -etm entries.
Same can be done for the umac entries

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -10,26 +10,72 @@ include 'pan/types';
 
 type ssh_preferred_authentication = string with match(SELF, '^(gssapi-with-mic|hostbased|publickey|keyboard-interactive|password)$');
 
+type ssh_ciphers = string with is_valid_ssh_cipher(SELF);
+type ssh_hostkeyalgorithms = string with match(SELF, "^(ssh-(rsa|dss|ed25519)|ecdsa-sha2-nistp(256|384|521)|(ssh-rsa-cert-v01|ssh-dss-cert-v01|ecdsa-sha2-nistp256-cert-v01|ecdsa-sha2-nistp384-cert-v01|ecdsa-sha2-nistp521-cert-v01|ssh-rsa-cert-v00|ssh-dss-cert-v00|ssh-ed25519-cert-v01)@openssh.com)$");
```
kwaegema: you can also group this more by adding ``(-cert-v01)?`` to the ecdsa-sha2-nistp part
ssh-ed25519-cert-v01 does not match or is this by design?

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -10,18 +10,63 @@ include 'pan/types';
...
+        };
+    };
+    true;
+};
+
 type ssh_core_options_type = {
     "AddressFamily" ? string with match (SELF, '^(any|inet6?)$')
     "ChallengeResponseAuthentication" ? legacy_binary_affirmation_string
-    "Ciphers" ? string
+    "Ciphers" ? legacy_ssh_ciphers
```
kwaegema: Ideally, this should be legacy_ssh_cipher[], but the component should be able to handle this.. As I see the component is not using Textrender, i doubt it. 
I guess we should open an issue to have ncm-ssh using Textrender , and review this types then..

kwaegema: Updated issue #434

ncm-ssh/src/main/pan/components/ssh/schema.pan
JensTimmerman: `ssh-ed25519-cert-v01` doesn't match because it needs to be `ssh-ed25519-cert-v01@openssh.com` so that's by design

ncm-ssh/src/main/pan/components/ssh/schema.pan
JensTimmerman: no, `hmac-ripemd160@openssh.com` isn't matched by `hmac-ripemd160`

ncm-ssh/src/main/pan/components/ssh/schema.pan
JensTimmerman: wel, it used to be string, so we can't just change it into a list? that's why I had to create the legacy string with...

JensTimmerman:
@kwaegema I don't think the shorter regexes make things more readable, I actually prefered the longer ones, they were more readable.

stdweird:
please squash the commits

jrha:
This also modified ncm-metaconfig's httpd schema, is that intentional and could you split it into a separate PR?

JensTimmerman:
that was intentional

JensTimmerman:
split into https://github.com/quattor/configuration-modules-core/pull/1059 and https://github.com/quattor/configuration-modules-core/pull/1058

jrha:
Thanks

PR #985, opened 2016-12-08
ncm-filesystems: rewrite to reuse ncm-fstab code

kwaegema:
Use fstab methods for filesystems
Filesystems is written to also include ncm-fstab functionality (So if you want to use filesystems somewhere, you don't need fstab) .
 If you would set manage_blockdevs to false, it would do the same as ncm-fstab

### DIFF
--- a/ncm-filesystems/src/main/pan/components/filesystems/schema.pan
+++ b/ncm-filesystems/src/main/pan/components/filesystems/schema.pan
@@ -8,10 +8,14 @@ include 'quattor/schema';
 include 'quattor/blockdevices';
 include 'quattor/filesystems';
 
+@documentation{
+when manage_blockdevs is false, filesystems does same as fstab
+No other resources here: this component takes its configuration
+from fstab component, "/system/filesystems" and "/system/blockdevices"
+}
 type structure_component_filesystems = {
     include structure_component
-    # No resources here: this component takes its configuration
-    # from "/system/filesystems" and "/system/blockdevices"
+    'manage_blockdevs' : boolean = true
 };
 
 bind "/software/components/filesystems" = structure_component_filesystems;
--- a/ncm-filesystems/src/main/perl/filesystems.pm
+++ b/ncm-filesystems/src/main/perl/filesystems.pm
@@ -1,116 +1,130 @@
 #${PMpre} NCM::Component::filesystems${PMpost}
 
-use LC::Exception qw (throw_error);
-use LC::Process qw (output);
-use Cwd qw(abs_path);
+use CAF::Process;
+use CAF::FileEditor;
+use CAF::FileReader;
 use NCM::Filesystem;
 use NCM::Partition 16.12.1 qw (partition_sort);
 use CAF::Object;
+use Fcntl qw(:seek);
+use Readonly;
+use Cwd qw(abs_path);
 
-use constant PROTECTED_PATH => "/software/components/fstab/protected_mounts";
+Readonly my $UMOUNT => "/bin/umount";
+Readonly my $FSTAB_CMP_PATH => '/software/components/fstab';
+Readonly my $FS_TREE => '/system/filesystems';
+Readonly my $PARTITIONS_TREE => '/system/blockdevices/partitions';
+use parent qw(NCM::Component::fstab);
 
-use parent qw(NCM::Component);
-our $EC = LC::Exception::Context->new->will_store_all;
+our $EC = LC::Exception::Context->new()->will_store_all();
+our $NoActionSupported = 0; # First needs check/fixing in ncm-lib-blockdevices issue #74
 
-# Returns a hash with the canonical paths to the protected mountpoints
-# as its keys. I This small overhead may simplify the code later.
-sub protected_hash
+#  Removes filesystem and deletes stuff that is not present in the %mounts 
+#  argument from fstab.
+sub delete_outdated
 {
-	my $config = shift;
-	my %ph;
+    my ($self, $fstab, $mountsr, $config, $remove) = @_;
 
-	my $pl = $config->getElement (PROTECTED_PATH)->getTree;
+    my @rm;
+    my %mounts = %$mountsr;
+    $fstab->seek_begin();
 
-	foreach my $i (@$pl) {
-		$ph{$i} = 0;
-		my $path = abs_path("$i/");
-		$ph{$path} = 0 if $path;
-	}
-	return %ph;
+    while (my $f = <$fstab>) {
+        my $mount = $self->mount_from_entry($f) or next;
+        if (!exists ($mounts{$mount}) && !exists($mounts{abs_path($mount)})) {
+            CAF::Process->new ([$UMOUNT, $mount],
+                log => $self)->run();
+            push (@rm, $f);
+            $self->verbose ("Scheduling for removal from fstab: $mount");
+            $self->verbose ("Scheduling for removal filesystem $mount") if $remove;
+        }
+    }
+
+    foreach my $outdated (@rm) {
+        if ($remove) {
+            my $fsrm = NCM::Filesystem->new_from_fstab ($outdated, $config, log => $self);
+            $self->info("Removing filesystem for $fsrm->{mountpoint}");
+            if ($fsrm->remove_if_needed) {
+                $self->error("error $fsrm->{mountpoint}");
+                return;
+            } else {
+                $self->info("removed $fsrm->{mountpoint}")
+            }
+        }
+        $self->info("Removing line $outdated");
+        $fstab->replace_lines (qr{$outdated}, qr{^$}, "");
+    }
+    return 1;
 }
 
-# Returns a hash with the mountpoints of the filesystems defined on
-# the profile as its keys, and the filesystem objects as its values.
-sub fshash
+
+sub create_blockdevices
 {
-	my $fsl = shift;
-	my %fsh;
+    my ($self, $config, $fs) = @_;
+    # Partitions must be created first
+    my $parttree = $config->getElement ($PARTITIONS_TREE);
+    my @part = ();
+    $self->info ("Checking whether partitions need to be created");
 
-	$fsh{$_->{mountpoint}} = $_ foreach @$fsl;
-	return %fsh;
-}
+    while ($parttree && $parttree->hasNextElement) {
+        my $partel = $parttree->getNextElement;
+        push (@part, NCM::Partition->new ($partel->getPath->toString, $config, log => $self));
+    }
+    foreach my $partition (partition_sort(@part)) {
+        if ($partition->create != 0) {
+            $self->error("Couldn't create partition: " . $partition->devpath);
+            return 0;
+        }
+    }
+    foreach my $filesystem (@$fs) {
+        if ($filesystem->create_if_needed != 0) {
+            $self->error("Failed to create filesystem:  $filesystem->{mountpoint}");
+            return 0;
+        }
+    }
+    return 1;
+};
 
-# Frees space on the system. It removes filesystems that are not
-# present in the profile anymore, and are *not* present in the
-# protected hash either. Returns 0 on success, -1 on error.
-sub free_space
-{
-	my ($self, $cfg, $protected, @fs) = @_;
-
-	my %ph = %$protected;
-	my %fsh = fshash (\@fs);
-
-	my $fl = output ("grep", "^[[:space:]]*#", "/etc/fstab", "-v");
-
-	$self->info ("Checking filesystems not defined in the profile");
-	my @fstab = split ("\n+", $fl);
-	foreach my $l (@fstab) {
-		$self->debug (5,"Fstab line: $l");
-		$l =~ m{^\S+\s+(\S+)\s} or next;
-		my $mount = $1;
-		next if(exists $fsh{$mount} || exists $ph{$mount}
-			    || exists($ph{abs_path($mount)}));
-		$self->verbose("Removing filesystem $mount");
-		my $f = NCM::Filesystem->new_from_fstab ($l, log => $self);
-		$f->remove_if_needed==0 or return -1;
-
-	}
-	return 0;
-}
 
 sub Configure
 {
-	my ($self, $config) = @_;
+    my ($self, $config) = @_;
 
-	my @fs = ();
+    my $tree = $config->getTree($self->prefix());
+    my $fstab_tree = $config->getTree($FSTAB_CMP_PATH);
 
-	if ($CAF::Object::NoAction) {
-		$self->warn ("--noaction not supported. Leaving.");
-		return 1;
-	}
+    my $manage = $tree->{manage_blockdevs};
+    $self->info('Managing blockdevices: ', ($manage) ? 'yes' : 'no');
+    my $fstab = CAF::FileEditor->new (NCM::Filesystem::FSTAB, log => $self,
+				      backup => '.old');
 
-	my $el = $config->getElement ("/system/filesystems");
-	while ($el->hasNextElement) {
-		my $el2 = $el->getNextElement;
-		push (@fs, NCM::Filesystem->new ($el2->getPath->toString, $config, log => $self));
-	}
-	my %protected = protected_hash ($config);
-	$self->free_space ($config, \%protected, @fs)==0 or return 0;
-	# Partitions must be created first, see bug #26137
-	$el = $config->getElement ("/system/blockdevices/partitions");
-	my @part = ();
-	$self->info ("Checking whether partitions need to be created");
+    my $fs = [];
+    my $fstree = $config->getElement ($FS_TREE);
+    while ($fstree->hasNextElement) {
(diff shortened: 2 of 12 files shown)


ncm-filesystems/src/main/pan/components/filesystems/schema.pan
```diff
@@ -9,9 +9,10 @@ include {'quattor/blockdevices'};
 include {'quattor/filesystems'};
 
 type structure_component_filesystems = {
-	include structure_component
-	# No resources here: this component takes its configuration
-	# from "/system/filesystems" and "/system/blockdevices"
+    include structure_component
+    'manage_blockdevs' : boolean = true # when this is false, filesystems does same as fstab
+    # No other resources here: this component takes its configuration
```
stdweird: put this as annotation above the type

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
-#${PMpre} NCM::Component::filesystems${PMpost}
```
stdweird: please keep this

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
-
-# Returns a hash with the canonical paths to the protected mountpoints
-# as its keys. I This small overhead may simplify the code later.
-sub protected_hash
-{
-	my $config = shift;
-	my %ph;
+use Cwd qw(abs_path);
+ 
+use constant UMOUNT => "/bin/umount";
```
stdweird: readonly pls

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+
+    }
+    return 1;
 }
 
+
+sub create_blockdevices
+{
+    my ($self, $config, $fs) = @_;
+    # Partitions must be created first, see bug #26137
```
stdweird: did you find any reference to this bug?

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+    my $el = $config->getElement ("/system/blockdevices/partitions");
+    my @part = ();
+    $self->info ("Checking whether partitions need to be created");
+
+    while ($el && $el->hasNextElement) {
+        my $el2 = $el->getNextElement;
+        push (@part, NCM::Partition->new ($el2->getPath->toString, $config, log => $self));
+    }
+    foreach (sort partition_compare @part) {
```
stdweird: use a variabel instead of relying on magical `$_`

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+        if ($_->create != 0) {
+            $self->error("Couldn't create partition: " . $_->devpath);
+            return 0;
+        }
+    }
+    foreach (@$fs) {
```
stdweird: same, `$_` is pure magic

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
-		if ($_->format_if_needed (%protected) != 0) {
-			$self->warn ("Failed to format filesystem: ".
-				      $_->{mountpoint});
-		}
-	}
-	return 1;
+    my ($self, $config) = @_;
+
+    my $tree = $config->getTree($self->prefix());
+    my $fstab_tree = $config->getTree('/software/components/fstab');
```
stdweird: fstab path in readonly

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+
+    my $manage = $tree->{manage_blockdevs};
+    if ($manage) {
+        $self->info('Managing blockdevices');
```
stdweird: always report if you are managing or not

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+    my $fstab = CAF::FileEditor->new (NCM::Filesystem::FSTAB, log => $self,
+				      backup => '.old');
+
+    if ($NoAction) {
+        $self->warn ("--noaction not supported. Leaving.");
+        return 1;
+    }
+
+    my $fs = [];
+    my $el = $config->getElement ("/system/filesystems");
```
stdweird: use readonly

kwaegema:
@stdweird remarks fixed

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
+
+    my $tree = $config->getTree($self->prefix());
+    my $fstab_tree = $config->getTree($FSTAB_CMP_PATH);
+
+    my $manage = $tree->{manage_blockdevs};
+    $self->info('Managing blockdevices: ', ($manage) ? 'yes' : 'no');
+    my $fstab = CAF::FileEditor->new (NCM::Filesystem::FSTAB, log => $self,
+				      backup => '.old');
+
+    if ($NoAction) {
```
ned21: If no action is not supported then why enable it? Or is it that NoAction is only partially supported? What isn't done under NoAction? I see a subsequent commit to the one that added this block disables NoAction but is there a plan to partially re-enable it in the future or should this check also be removed and replaced with another comment indicating that this is the section that's not safe?

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
-
-	foreach my $i (@$pl) {
-		$ph{$i} = 0;
-		my $path = abs_path("$i/");
-		$ph{$path} = 0 if $path;
-	}
-	return %ph;
-}
+our $EC = LC::Exception::Context->new()->will_store_all();
+our $NoActionSupported = 0; # First needs check/fixing in ncm-lib-blockdevices
```
ned21: Please add the GH issue number for the ncm-lib-blockdevices fixes.

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,148 @@
...
-		}
-	}
-	return 1;
+    my ($self, $config) = @_;
+
+    my $tree = $config->getTree($self->prefix());
+    my $fstab_tree = $config->getTree($FSTAB_CMP_PATH);
+
+    my $manage = $tree->{manage_blockdevs};
+    $self->info('Managing blockdevices: ', ($manage) ? 'yes' : 'no');
```
ned21: Can you add a note about the manage_blockdevs attribute to the .POD since it's not referenced. Also I'm not sure I know why this setting is useful: if you want this behaviour, why not set this component to active = false and use ncm-fstab?

ned21:
The new PAN lint script that's causing the build failure is very impressive!

stdweird:
@kwaegema please rebase on master. it might fix the the panlint failures.

ncm-filesystems/src/main/perl/filesystems.pm
kwaegema: this was some existing code, can be removed in favor for disabling it until ncm-lib-blockdevices is checked

kwaegema:
rebased, squashed and fixed remarks

ncm-filesystems/src/main/perl/filesystems.pod
```diff
@@ -16,6 +16,8 @@ from C<< /etc/fstab >>, using ncm-fstab.
 The component doesn't provide any special resources at the moment. It
 just watches for changes on C<< /system/filesystems >> and C<< /system/blockdevices >>
 and creates new filesystems, if needed.
+You can also use ncm-filesystems to replace ncm-fstab :
+If C<< manage_blockdevs >> is set to false, only the ncm-fstab code will run.
```
ned21: Is this functionality actually useful though? Wouldn't it be less code to maintain if we just dropped this feature? Apologies if I am being dense and missing an obvious use case!

kwaegema: It is not a lot of code, but probably most useful for testing: With the flag false, I can use tests of ncm-fstab to check ncm-filesystems does at least the same. But I can remove it quite simple :smile:

ned21: hmm.. Testing functionality sounds useful but perhaps that should be enabled some other way?What do other people think?

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+            $self->verbose ("Scheduling for removal of fstab: $mount");
+            $self->verbose ("Scheduling for removal filesystem $mount") if $remove;
+    	}
+    }
+
+    foreach my $outdated (@rm) {
+        if ($remove) {
+            my $fsrm = NCM::Filesystem->new_from_fstab ($outdated, $config, log => $self);
+            $self->info("Removing filesystem for $fsrm->{mountpoint}");
+            $fsrm->remove_if_needed==0 or return 0;
```
stdweird: what? (also, just `return` on failure)

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+        }
+        $self->info("Removing line $outdated");
+        $fstab->replace_lines (qr{$outdated}, qr{^$}, "");
+
```
stdweird: remvoe newline

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+    }
+    return 1;
 }
 
+
+sub create_blockdevices
+{
+    my ($self, $config, $fs) = @_;
+    # Partitions must be created first
+    my $el = $config->getElement ("/system/blockdevices/partitions");
```
stdweird: path should be a `Readonly`

stdweird: or imported from `fstab` component if used there

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+    my @part = ();
+    $self->info ("Checking whether partitions need to be created");
+
+    while ($el && $el->hasNextElement) {
+        my $el2 = $el->getNextElement;
```
stdweird: i probably wrote this code, but can we have decent variable names instead of el and el2 😛

ncm-filesystems/src/main/perl/filesystems.pm
stdweird: @ned21 see my proposal on using an alias above

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+    my $tree = $config->getTree($self->prefix());
+    my $fstab_tree = $config->getTree($FSTAB_CMP_PATH);
+
+    my $manage = $tree->{manage_blockdevs};
+    $self->info('Managing blockdevices: ', ($manage) ? 'yes' : 'no');
+    my $fstab = CAF::FileEditor->new (NCM::Filesystem::FSTAB, log => $self,
+				      backup => '.old');
+
+    my $fs = [];
+    my $el = $config->getElement ($FS_TREE);
```
stdweird: same remark wrt el/el2 as above.
can you also open an issue in CCM that we need a `config->getPathTree` to return a tree with keys the path and value the elements; and a `path` option to `getTree` to return a tree with usual keys and the path as values. (this code needs the 2nd one).

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+    while ($el->hasNextElement) {
+        my $el2 = $el->getNextElement;
+        push (@$fs, NCM::Filesystem->new ($el2->getPath->toString, $config, log => $self));
+    }
+
+    my $protected = $self->protected_hash($fstab_tree);
+    my %mounts = $self->fshash ($fs);
```
stdweird: if this is the only place where fshash is used, use the `map` code directly in the last argument on the next line. this does not deserve its own method 😄

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
+        if ($filesystem->create_if_needed != 0) {
+            $self->error("Failed to create filesystem:  $filesystem->{mountpoint}");
+            return 0;
+        }
+    }
+    return 1;
+};
+
+
 sub Configure
```
stdweird: how different is this code from the fstab one? if we go for aliases, i propose to make the `Configure` code equal, by adding noop subs to the fstab code or or using `$self->can()` to tets if th emethod exists.

ncm-filesystems/src/main/pan/components/filesystems/schema.pan
```diff
@@ -8,10 +8,14 @@ include 'quattor/schema';
 include 'quattor/blockdevices';
 include 'quattor/filesystems';
 
+@documentation{
+when manage_blockdevs is false, filesystems does same as fstab
```
kwaegema: well, not having the code in ncm-fstab was the reason why I closed #600 : 'Agreement from the workshop: refactoring the duplicated code into ncm-lib-blockdevices or somewhere else is a better and safer way. RAL and Morgan Stanley still want to keep the dangerous code isolated and if possible not even installed.'
running both fstab and filesystems will first do filesystems, then run fstab (so fstab code will run twice, as if you run the component twice)

ncm-filesystems/src/main/perl/filesystems.pm
```diff
@@ -1,116 +1,143 @@
...
-
-	foreach my $i (@$pl) {
-		$ph{$i} = 0;
-		my $path = abs_path("$i/");
-		$ph{$path} = 0 if $path;
-	}
-	return %ph;
-}
+our $EC = LC::Exception::Context->new()->will_store_all();
+our $NoActionSupported = 0; # First needs check/fixing in ncm-lib-blockdevices issue #74
```
kwaegema: issue #74 is created for that. it is not really a requirement for this PR (I moved the NoAction test, but filesystems NoAction behaviour isn't changed here).

ncm-filesystems/src/main/perl/filesystems.pm
kwaegema: what do you mean by 'what' ? ncm-lib-blockdevices returns 0 on success..

ncm-filesystems/src/main/pan/components/filesystems/schema.pan
stdweird: with an alias, you run fstab component, but either with the fstab.pm or the filesystems.pm code.

that way the code is not in the same module, but the relation is much clearer.

ncm-filesystems/src/main/perl/filesystems.pm
stdweird: i mean it's ver very cryptic. especially if ncm-lib-blockdevices does not report an error. it would be better to explcitly,

```perl
if ($fsrm->remove_if_needed) {
    $self->error("error $fsrm->{mountpoint}");
} else {
    $self->info("removed $fsrm->{mountpoint}")
}
```

you are after all deteing a filesystem, it's nice we have this reporter/logged

ncm-filesystems/src/main/pan/components/filesystems/schema.pan
ned21: I don't like the alias idea - that makes it the same code that can destroy data and is only protected by the name through which the code is invoked. ncm-filesystems should be a separate component (a subclass to avoid code duplication) and people that want to use it should have to disable ncm-fstab to prevent it running twice (we can enforce that with templates).

stdweird: the whole `dangerous code isolated` is FUD unless you also don't use `ncm-symlink` and probably some other components that can easily do inrepairable damage. i tought that it was sufficient that the code was not in `fstab.pm`; i never understood it as `not in ncm-fstab`.
anyway, if this is the agreement, so be it.

kwaegema:
remarks addressed

kwaegema:
Had to fix and resolve conflicts due to rebase on master

ned21:
Is everyone happy with the code now? It looks to me like all the requirements and preferences have been met. In which case there be one final rebase and re-word of the one-liner logs to provide more detail as to what actually changed in each commit. e.g.

	more remarks processed => describe actual changes or squash
        please travis => modernise PAN files to comply with panlint
        fix partition_sort => fix up commit should be squashed?

kwaegema:
@ned21 rebased and squashed

PR #986, opened 2016-12-08
ncm-fstab: pass tree instead of config

kwaegema:
Some changes needed to subclass ncm-filesystems from ncm-fstab. So also needed for the tests to run for #985


stdweird:
add in the description that this is to support subclassing, LGTM otherwise

kwaegema:
done

PR #989, opened 2016-12-11
ncm-freeipa: mock_rpc mocks JSON::XS which should be unmocked before Devel::Cover produces the digests

stdweird:


stdweird:
Should handle jenkins master failures like https://jenkins1.ugent.be/job/NCM_components/1722/

stdweird:
retest this please

PR #990, opened 2016-12-12
ncm-useraccess: close dangling FileWriter instance before DESTROY

stdweird:
Unittest fail with usual 'throw_error on DESTROY'


stdweird:
@ned21 i removed the whitespace changes

PR #988, opened 2016-12-08
ncm-*: Trim whitespace from Pan templates

jrha:
* Convert all tabs to four spaces.
* Remove trailing whitespace.
* Collapse multiple spaces within type declarations.
* Add missing newlines to end of files.

Also simplifies one of the pieces of test data used by ncm-named to make it not depend on whitespace.

When merged, this will likely cause all other open PRs on this repository to have conflicts, so it should probably get merged immediately following the release of 16.12.

PR #995, opened 2016-12-14
ncm-*: Reflow indentation in pan templates

jrha:
Correct many blocks of incorrect indentation in component pan templates.
This was done in a semi-automated fashion with lots of manual help, so it is worth looking over before merging.

Based on and hence requires #988.


stdweird:
retest this please

PR #1006, opened 2017-01-04
ncm-authconfig: sssd: use sensible default debug_level

stdweird:
previous default `0x0210` loglevel makes no sense (fatal failure and function data). it is probably some miscaclculation/misconversion.
the new default logs fatal/critical/serious/minor failures and configuration data on startup. it's not too verbose, log files stay reasonably sized

### DIFF
--- a/ncm-authconfig/src/main/pan/components/authconfig/sssd.pan
+++ b/ncm-authconfig/src/main/pan/components/authconfig/sssd.pan
@@ -104,6 +104,7 @@ type authconfig_sssd_domain  = {
     "min_id" : long = 1
     "max_id" : long = 0
     "enumerate" : boolean = false
+    "timeout" : long = 10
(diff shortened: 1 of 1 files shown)


jrha:
Is there any explanation of the magical hex values?

jrha:
I would like to see a description on the PR, an explanation of the hex magic and some consideration of validation for these parameters.

jrha:
@stdweird ping

stdweird:
@jrha is this sufficient

jrha:
It's better, I still don't understand where the values come from, are they documented somewhere?

jrha:
I just found https://linux.die.net/man/5/sssd.conf

> This is new format of debug levels introduced in 1.7.0. Older format (numbers from 0-10) is compatible but deprecated.

PR #1022, opened 2017-01-26
ncm-grub: support settings passwords 

ajf8:
@stdweird's nice refactor of ncm-grub, mostly uses grubby commands now, but as pointed out the password support still manipulates grub.conf directly, which will not work with grub2.

This is a simple change to use /boot/grub2/user.cfg and the different syntax if /boot/grub2 is present (other suggestions welcome).

The serial console stuff is still todo.

### DIFF
--- a/ncm-grub/src/main/perl/grub.pm
+++ b/ncm-grub/src/main/perl/grub.pm
@@ -1,6 +1,6 @@
 #${PMpre} NCM::Component::grub${PMpost}
 
-use Fcntl qw(SEEK_SET);
+use Fcntl qw(SEEK_SET SEEK_END);
 use CAF::Object qw(SUCCESS);
 use CAF::FileEditor;
 use CAF::FileWriter;
@@ -14,6 +14,8 @@ our $NoActionSupported = 1;
 Readonly my $PATH_KERNEL_VERSION => '/system/kernel/version';
 Readonly my $PATH_CONSOLE_SERIAL => '/hardware/console/serial';
 Readonly my $GRUB_CONF           => '/boot/grub/grub.conf';
+Readonly my $GRUB2_DIR           => '/boot/grub2';
+Readonly my $GRUB2_USER_CFG      => "$GRUB2_DIR/user.cfg";
 Readonly my $GRUBBY              => '/sbin/grubby';
 Readonly my $PREFIX              => '/boot';
 Readonly::Hash my %SERIAL_CONSOLE_DEFAULTS => {
@@ -22,6 +24,8 @@ Readonly::Hash my %SERIAL_CONSOLE_DEFAULTS => {
     word => 8,
     parity => "n",
 };
+
+our $GRUB_MAJOR = -d $GRUB2_DIR ? 2 : 1;
 
 =pod
 
@@ -191,11 +195,25 @@ sub password
     my $val = $tree->{option} ? "--$tree->{option} " : "";
     $val .= $password;
 
-    $grub_fh->add_or_replace_lines(qr/^password\s+/,
-                                   qr/^password $val$/,
-                                   "password $val\n",
-                                   $self->main_section_offset($grub_fh),
-                                   );
+    if ( $GRUB_MAJOR == 2 ) {
+        my $usercfg_fh = CAF::FileEditor->new( $GRUB2_USER_CFG,
+            owner => "root",
+            group => "root",
+            mode  => 0400,
+            log   => $self );
+        
+        $usercfg_fh->add_or_replace_lines(qr/^GRUB2_PASSWORD=/, qr/^GRUB2_PASSWORD=$password$/, "GRUB2_PASSWORD=$password\n", SEEK_END);
+        $usercfg_fh->close();
+    } else {
+        my $val = $tree->{option} ? "--$tree->{option} " : "";
+        $val .= $password;
+
+        $grub_fh->add_or_replace_lines(qr/^password\s+/,
+                                  qr/^password $val$/,
+                                  "password $val\n",
+                                  $self->main_section_offset($grub_fh),
+                                  SEEK_END);
+    }
 
     return SUCCESS;
 }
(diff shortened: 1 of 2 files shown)


ncm-grub/src/main/perl/grub.pm
```diff
@@ -191,11 +193,25 @@ sub password
...
-                                   );
+    if ( -d $GRUB2_DIR ) {
+        my $usercfg_fh = CAF::FileEditor->new( $GRUB2_USER_CFG,
+            owner => "root",
+            group => "root",
+            mode  => 0400,
+            log   => $self );
+        
+        $usercfg_fh->add_or_replace_lines(qr/^GRUB_PASSWORD=/, qr/^GRUB_PASSWORD=$password$/, "GRUB_PASSWORD=$password\n");
+        $usercfg_fh->close();
```
stdweird: this is not suficient on it's own i think. doesn't this need a `grub-update`?
btw, for serial console support, you can modifiy `/etc/default/grub` (https://wiki.archlinux.org/index.php/working_with_the_serial_console) and do an update.

ncm-grub/src/main/perl/grub.pm
```diff
@@ -191,11 +193,25 @@ sub password
     my $val = $tree->{option} ? "--$tree->{option} " : "";
     $val .= $password;
 
-    $grub_fh->add_or_replace_lines(qr/^password\s+/,
-                                   qr/^password $val$/,
-                                   "password $val\n",
-                                   $self->main_section_offset($grub_fh),
-                                   );
+    if ( -d $GRUB2_DIR ) {
```
stdweird: this is probably not sufficient. check `man grubby` on eg el7 and look for the location of files on certain arch/bios combos.
`/etc/grub2.cfg` is an EL7-ism that is a symlink to whatever arch/bios combo; so you might also check that (or only check that).

ajf8: Good point about grub2 x86_64 UEFI using /boot/efi/EIF/redhat/grub.cfg. Am I right that we do not support any of the other architectures?

But grub2-efi depends on grub2-tools, which does provide this directory.

# rpm --requires -qip /root/grub2-efi-2.02-0.44.el7.x86_64.rpm
grub2-tools = 1:2.02-0.44.el7

# rpm -ql grub2-tools|grep /boot/grub2
/boot/grub2

The key part is /etc/grub.d/01_users, which is also in grub2-tools.

[root@izvmn08azgh7 tmp]# rpm -ql grub2-tools| grep users
/etc/grub.d/01_users
[root@izvmn08azgh7 tmp]# cat /etc/grub.d/01_users
#!/bin/sh -e
cat << EOF
if [ -f \${prefix}/user.cfg ]; then
  source \${prefix}/user.cfg
  if [ -n "\${GRUB2_PASSWORD}" ]; then
    set superusers="root"
    export superusers
    password_pbkdf2 root \${GRUB2_PASSWORD}
  fi
fi
EOF

It sources the user.cfg file, meaning you don't have to regenerate the grub config. I think it should work for both UEFI and BIOS.

stdweird: i don't think so. all the shell scripts in `/etc/grub.d` are used during grub2 update (it's not that during boot, grub2 reads/parses `/etc/grub.d`, it only reads the grub.cfg from /boot

ajf8: I get that, but my point is that /etc/grub.d/01_users seems to be common across UEFI and BIOS as part of grub2-tools, so wherever the grub.cfg is it should contain the same block which reads $prefix/user.cfg. And /boot/grub2 will be there for both too.

stdweird: @ajf i agree, but you still need to run `grub-mkconfig`, changing the files is not enough

ajf8: Hmm, I'm not sure.  If you follow the method on the ubuntu wiki, and some methods I've read for el7, where you add "set superusers ...", "password_pbkdf2 ..." etc to a grub.d config, then sure you must run grub2-mkconfig.

But if you are making use of this standard grub.d, since it sources user.cfg itself, so then I don't think it's necessary. Do you see what I mean, or am I missing something?

# grub2-mkconfig >/tmp/grub.cfg.old
/usr/bin/grub2-editenv: error: cannot rename the file /boot/grub2/grubenv.new to /boot/grub2/grubenv: No such file or directory.
Generating grub configuration file ...
Found linux image: /boot/vmlinuz-3.10.0-514.6.1.el7.x86_64
Found initrd image: /boot/initramfs-3.10.0-514.6.1.el7.x86_64.img
Found linux image: /boot/vmlinuz-0-rescue-3a304c440aca4b8bbbd43fe3b4db7297
Found initrd image: /boot/initramfs-0-rescue-3a304c440aca4b8bbbd43fe3b4db7297.img
done
# mv user.cfg user.cfg.bak
# grub2-mkconfig >/tmp/grub.cfg.new
/usr/bin/grub2-editenv: error: cannot rename the file /boot/grub2/grubenv.new to /boot/grub2/grubenv: No such file or directory.
Generating grub configuration file ...
Found linux image: /boot/vmlinuz-3.10.0-514.6.1.el7.x86_64
Found initrd image: /boot/initramfs-3.10.0-514.6.1.el7.x86_64.img
Found linux image: /boot/vmlinuz-0-rescue-3a304c440aca4b8bbbd43fe3b4db7297
Found initrd image: /boot/initramfs-0-rescue-3a304c440aca4b8bbbd43fe3b4db7297.img
done
# diff -Naur /tmp/grub.cfg.old /tmp/grub.cfg.new
#

ajf8: BTW el7 provides a utility grub2-setpassword which I verified with strace only updates user.cfg and does nothing else. The only reason I'm not using it is that it prompts for the plaintext password.

stdweird: @ajf8 if it works, that's what counts ;) i obviously misunderstand some of the internals wrt setting the user password.

ajf8: Great :) I tested it and it did seem to work. Is there any other points to be addressed?

stdweird: unittests?

ajf8: Thanks, just added a test

Issue #1023, opened 2017-01-27
ncm-syslog: bugs


stdweird:
I'm rewriting the `ncm-syslog` code, and discovered following bugs
1. explicit `fullcontrol = false` is treated by code as `fullcontrol = true`
2. the `template` part of a rule is ignored
3. rules without selectors are ignored when fullcontrol is not set

the rewrite targets `fullcontrol = true` usage
1. will be fixed (`->getTree` is the only thing you should ever use)
2. will be fixed for `fullcontrol = true`
3. will not be fixed (this is an issue with multiline editing; and a more fundamental issue with the primitive schema when trying to express multiple actions for a single rule)

people using ncm-syslog with `fullcontrol = false` are welcome to suggest fixes for 2 and 3

PR #1024, opened 2017-01-27
ncm-syslog: cleanup and refactor

stdweird:
Addresses some bugs from #1023 
Requires https://github.com/quattor/CAF/pull/211 and https://github.com/quattor/CAF/pull/206


stdweird:
Needs unittests.
There will be a follow up PR to use TT and support rules/rulesets as rsyslog supports them; possibly in rainerscript only.

stdweird:
unittests added

ncm-syslog/src/main/pan/components/syslog/schema.pan
```diff
@@ -5,29 +5,32 @@
...
+    "daemontype" ? string with match (SELF, '^(syslog|rsyslog)$')
     "file" ? string
+    @{Options for syslogd /etc/sysconfig/(r)syslog (will be wrapped in double quotes if needed)}
     "syslogdoptions" ? string
+    @{Options for the klogd /etc/sysconfig/(r)syslog (will be wrapped in double quotes if needed)}
     "klogdoptions" ? string
+    @{Determines whether component has full control over the configuration file,
+      eventually erasing entries from other sources. If false, entries
+      from other sources are kept and configuration entries are added.}
     "fullcontrol" ? boolean
```
ned21: What happens if this is unset? If this is truly binary and not tri-state then would it be better to make this mandatory and set a default to make it explicit what the default behaviour is?

ncm-syslog/src/main/perl/syslog.pm
```diff
@@ -1,253 +1,243 @@
...
+=cut
+
+use parent qw(NCM::Component);
+
+use CAF::FileWriter;
+use CAF::FileEditor 16.12.1;
+use CAF::Service;
+use Readonly;
+
+Readonly my $SYSLOG_DEFAULT_DAEMONTYPE => 'syslog';
```
ned21: These should be defaulted in the PAN not the Perl.

ncm-syslog/src/main/pan/components/syslog/schema.pan
stdweird: should be `false/unset`; i'll fix the docs. this didn't used to be the case. the old code did `if set` instead of `if true` 😄

ncm-syslog/src/main/perl/syslog.pm
stdweird: i'll also default this in pan; but i have to keep this in the code as well, as we don't enforce schema and component versions.

stdweird:
@ned21 remarks addressed

stdweird:
@ned21 @jrha suggestion on how to fix the panlint tests are welcome, esp if we can move to panc 10.4 and the choice type...

ned21:
re: 10.4, first we would need a [release](https://github.com/quattor/pan/releases)... (including panlint would be nice too, I had to recently cut an internal release for us to access it).

re: line length - I was able to wrap a string over multiple lines like this:
```
bind '/test' = string with match(SELF, '^(foo|bar|'+
                           'abc|cba)$');
```

stdweird:
@ned21 wrt panlint: if that will pass review, i'll go for it 😄 
wrt 10.4, it would be nice if MS could do a comparison based on full aquilon rebuild wrt correctness, speed and mem usage. jars from https://jenkins1.ugent.be/view/Quattor/job/panc/lastBuild/org.quattor.pan$panc/ should work.

jrha:
I wouldn't get hung up on line length, but rather than introduce extra concatenation, I'd wrap long lines like this...
```pan
bind '/test' = string with match(SELF,
    '^(foo|bar|abc|cba)$'
);
```

stdweird:
unittests fail due to missing CAF versioning.

jrha:
Needs re-basing to enable allow of mvn templates in Panlint.

jrha:
#1031, please rebase

stdweird:
retest this please

jrha:
@hpcugentbot test this please

stdweird:
test this please

stdweird:
test this please

PR #1030, opened 2017-02-09
ncm-*: Update build tools to version 1.51

jrha:


jrha:
:see_no_evil:

stdweird:
i propose to do to move to the tqu unittest component per component, incl pan template cleanup. i'll make an example PR for a single component; but we should avoid fixing components that have open PRs.

stdweird:
proposal to cleanup in https://github.com/quattor/quattor.github.com/pull/199

stdweird:
https://github.com/quattor/configuration-modules-core/pull/1031 is an example of such cleanup. should be more or less in line with the post

stdweird:
@jrha crap, the panlinter doesn't like maven templates.

jrha:
Superseded by 1.52.

PR #1037, opened 2017-02-21
ncm-ceph: Correct and update ncm-ceph schema.

brunocanning:
This change corrects an erroneous setting from a previous commit (op_queue) and adds a number of settings useful when operating large clusters.

I do not expect this to cause any backwards incompatibility as ceph ignores the erroneous config setting, so it has no affect and removing it should have no consequence.


jrha:
@hpcugentbot please test this

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
     # the component sets this to false when generating crushmap itself, and true when crushmap is generated by ceph
     'osd_crush_update_on_start' ? boolean
+    'osd_disk_thread_ioprio_class' ? string with match(SELF, '^(idle|be|rt)$')
+    'osd_disk_thread_ioprio_priority' ? long(0..)
+    'osd_disk_threads' ? long(0..)
+    'osd_heartbeat_grace' ? long(0..)
+    'osd_heartbeat_interval' ? long(0..)
     'osd_journal_size' : long(0..) = 10240
+    'osd_max_backfills' ? long(0..)
+    'osd_max_scrubs' ? long(0..)
```
kwaegema: Some of these new options are already defined at https://github.com/quattor/configuration-modules-core/blob/master/ncm-ceph/src/main/pan/components/ceph/schema-osd.pan. I guess we shouldn't add non-global options here, but in the actual schema-<daemon> template. If you would need them as global config options, we can include the options here alltogether.

brunocanning:
@kwaegema Kenneth, good to hear from you! Sorry, I didn't check the `components/ceph/schema-osd.pan` file. I'll refactor this in due course.

stdweird:
retest this please

ned21:
If @kwaegema can do another review and request changes, then @jrha will arrange for someone to make the changes.

ncm-ceph/src/main/pan/components/ceph/schema.pan
kwaegema: This comment is still valid :smile:

jrha:
@kwaegema can you take another look at this?

ncm-ceph/src/main/pan/components/ceph/schema.pan
jrha: What do you propose we do with these?

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
     'mon_cluster_log_to_syslog' : boolean = true
     'mon_initial_members' : type_network_name [1..]
+    'mon_osd_down_out_interval' ? long(0..)
+    'mon_osd_down_out_subtree_limit' ? string # takes values of bucket types in crushmap, some of which are user definable
     'mon_osd_min_down_reporters' ? long(0..)
     'mon_osd_min_down_reports' ? long(0..)
     'mon_osd_max_op_age' ? long = 32
     'ms_type' ? string with match(SELF, '^(simple|async|xio)$')
-    'op_queue' ? string with match(SELF, '^(prio|wpq)$')
+    'osd_backfill_scan_max' ? long(0..)
```
kwaegema: these are also in schema-osd. should they be here too?

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
+    'osd_backfill_scan_min' ? long(0..)
+    'osd_client_op_priority' ? long(1..63)
```
kwaegema: is also in schema-osd

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
+    'osd_crush_chooseleaf_type' ? boolean
     # the component sets this to false when generating crushmap itself, and true when crushmap is generated by ceph
     'osd_crush_update_on_start' ? boolean
+    'osd_disk_thread_ioprio_class' ? string with match(SELF, '^(idle|be|rt)$')
```
kwaegema: use choice('idle','be','rt') ?

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
+    'osd_disk_thread_ioprio_priority' ? long(0..)
+    'osd_disk_threads' ? long(0..)
+    'osd_heartbeat_grace' ? long(0..)
+    'osd_heartbeat_interval' ? long(0..)
     'osd_journal_size' : long(0..) = 10240
+    'osd_max_backfills' ? long(0..)
+    'osd_max_scrubs' ? long(0..)
+    'osd_mon_report_interval_min' ? long(0..)
     'osd_objectstore' ? string
+    'osd_op_queue' ? string with match(SELF, '^(prio|wpq)$')
```
kwaegema: use choice

ncm-ceph/src/main/pan/components/ceph/schema.pan
```diff
@@ -187,19 +187,39 @@ type ceph_cluster_config = {
...
     'osd_pool_default_min_size' : long(0..) = 2
     'osd_pool_default_pg_num' ? long(0..)
     'osd_pool_default_pgp_num' ? long(0..)
     'osd_pool_default_size' : long(0..) = 3
+    'osd_recovery_max_active' ? long(0..)
```
kwaegema: these options also global and osd ?

kwaegema:
Sorry I forgot about this PR..
I made a new v2 schema in https://github.com/quattor/configuration-modules-core/pull/1230
Deploying ceph using that schema, makes an [osd] section (instead of a section for each individual osd). So I guess some global options could move to the [osd] section for that schema.

jrha:
I'll take a look to see if any of this is still needed in the next release cycle.

jrha:
No longer needed, we can just drop this.

PR #1044, opened 2017-02-23
ncm-aiiserver: switch to 00-tqu and buildtools 1.52

stdweird:


ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan
```diff
@@ -33,8 +29,10 @@ type structure_aiidhcp = {
     "norestart" ? boolean
 };
 
-type structure_component_aiiserver = {
+type aiiserver_component = {
     include structure_component
+    @{Configures the aii-shellfe tool.}
     "aii-shellfe" : structure_aiishellfe
+    @{Configures the aii-dhcp legacy tool.}
```
stdweird: true, i just copied the text from the pod. i'll make it `AII::DHCP and legacy aii-dhcp tool`?

PR #1045, opened 2017-02-24
ncm-altlogrotate: cleanup and refactor

### LINKED ISSUE #1038: ncm-altlogrotate: fix noolddir issue
while doing spring cleaning for `ncm-altlogrotate`, i came across the following bug
```perl
if (defined($entry{noolddir})) {
        $contents .= "noolddir\n" unless (exists $entry{olddir} && $entry{olddir}->getValue() eq 'true');
}
```
what is certainly wrong is that the `noolddir` value is not used, but also the logic seems wrong. i guess what is intended is 
```perl
unless (exists $entry{olddir} || $entry{noolddir}->getValue() eq 'false');
```
or for humans,
```perl
if (! exists($entry{olddir}) ) && $entry{noolddir}->getValue() eq 'true'
```

### LINKED ISSUE #1039: ncm-altlogroatate: tabooext is wrong
* tabooext should be rendered as a comma-separated list
* the taboo_replace logic is missing/wrong: when taboo_replace is true, no `+` should be added (as `+` means append to existing/default list)

stdweird:
Fixes #1038 
Fixes #1039


stdweird:
fixed the panlint failure on the unittest `simple.pan`

PR #1056, opened 2017-02-27
ncm-opennebula: Add aii component rpm

alvarosimon:
This PR includes the missing rpm template for ONE AII


alvarosimon:
@stdweird I think that everything is in place now...

PR #1058, opened 2017-03-01
ncm-ssh: disable insecure ssh ciphersuites and macs

JensTimmerman:
* sha1, des, 3des and md5 should no longer be used!
* This might break for people who are still using them, they should remove old ciphers, MAC's and key exchange algorithms.
* also remove rc4, it should no longer be used since the  Bar Mitzvah attack was published.

Replaces #983.


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -10,18 +10,64 @@ include 'pan/types';
 
 type ssh_preferred_authentication = string with match(SELF, '^(gssapi-with-mic|hostbased|publickey|keyboard-interactive|password)$');
 
+
+type ssh_ciphers = string with is_valid_ssh_cipher(SELF);
```
stdweird: you introduce a new type that already exists in `metacofnig/ssh` and it is not used in this template. make an issue to resolve this, and add a comment with the issue number here, and comment this type for now.
you can already check that the current definition in metaconfig /ssh is the same regexp

jrha:
Pushing back to 17.4 as we are trying to avoid user-facing changes in 17.3.

jrha:
@JensTimmerman could you address @stdweird's comment please?

JensTimmerman:
@jrha addressed. The travis test seems to be failing because of very long prefixes in the tests.

stdweird:
@JensTimmerman can you fix the typo in the title and prefix the commit message with `ncm-ssh`

jrha:

JensTimmerman:
@jrha I think I still need to address @stdweird's requests.

jrha:
@JensTimmerman any news?

jrha:
@JensTimmerman ping?

ncm-ssh/src/main/pan/components/ssh/schema.pan
JensTimmerman: what exactly do I need to resolve?

This way of working was recommended to me by @wdpypere  To keep it backwards compatible.

JensTimmerman:
@jrha @stdweird's comment isn't completely clear to me, so I'll await his clarification on what exactly needs to be resolved.

ncm-ssh/src/main/pan/components/ssh/schema.pan
wdpypere: I did what now?

stdweird: the types are unique. haveing this type here implies that you cannot include the metaconfig service that has the same type defined. there can be only one...

jrha:
ok thanks

jrha:
@stdweird can you clarify

stdweird:
@jrha i think it's clear now for @JensTimmerman he renamed the duplicate type names, to be removed in later PR when the types are part of template libary core

PR #1059, opened 2017-03-01
ncm-metaconfig: don't allow specific turning on of SEED and CAMELLIA

JensTimmerman:
* SEED and CAMELLIA are also insecure ciphersuites, you shouldn't be allowed to explicitly turn them on.
* This may break for people who have these turned on, they probably shouldn't.

Replaces #983.

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/httpd/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/httpd/pan/schema.pan
@@ -7,11 +7,23 @@ include 'components/accounts/functions';
 type httpd_sslprotocol = string with match(SELF, '^\+?(TLSv1|TLSv1\.[012])$')
     || error("Use a modern cipher protocol, for Pete's sake!");
 
-type httpd_ciphersuite = string with match(SELF, '^(\+?TLSv1|!(RC4|LOW|[ae]NULL|MD5|EXP|3DES|IDEA))$')
+type httpd_ciphersuite = string with match(SELF, '^(\+?TLSv1|!(RC4|LOW|[ae]NULL|MD5|EXP|3DES|IDEA|SEED|CAMELLIA))$')
     || error("Use a modern cipher suite, for Pete's sake!");
 
-type httpd_nss_protocol = string with match(SELF, '^(TLSv1\.[012]|SSLv3|All)$');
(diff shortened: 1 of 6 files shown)


ned21:
Could you please add a bit more detail to the commit log? It doesn't even contain as much as the PR description, so for instance anyone browsing through the log cannot see that this applies to the ncm-metaconfig config, and the httpd schema at that and not something else that uses ciphers.

I'm not a proponent of getting people who have a legitimate bug fix to clean up all the lint errors in a file they touch but if you could also fix up some or all of the lint errors as part of this PR then that would be a great help!

jrha:
Pushing back to 17.4 as we are trying to avoid user-facing changes in 17.3.

ncm-metaconfig/src/main/metaconfig/httpd/pan/schema.pan
```diff
@@ -7,11 +7,22 @@ include 'components/accounts/functions';
...
     || error("Use a modern cipher protocol, for Pete's sake!");
 
-type httpd_ciphersuite = string with match(SELF, '^(\+?TLSv1|!(RC4|LOW|[ae]NULL|MD5|EXP|3DES|IDEA))$')
+type httpd_ciphersuite = string with match(SELF, '^(\+?TLSv1|!(RC4|LOW|[ae]NULL|MD5|EXP|3DES|IDEA|SEED|CAMELLIA))$')
     || error("Use a modern cipher suite, for Pete's sake!");
 
-type httpd_nss_protocol = string with match(SELF, '^(TLSv1\.[012]|SSLv3|All)$');
-type httpd_nss_cipherstring = string with match(SELF, '^(\+|-)(rsa_3des_sha|rsa_des_56_sha|rsa_des_sha|rsa_null_md5|rsa_null_sha|rsa_rc2_40_md5|rsa_rc4_128_md5|rsa_rc4_128_sha|rsa_rc4_40_md5|rsa_rc4_56_sha|fortezza|fortezza_rc4_128_sha|fortezza_null|fips_des_sha|fips_3des_sha|rsa_aes_128_sha|rsa_aes_256_sha)$');
+# These are the settings for old clients, see https://access.redhat.com/articles/1467293 for stricter values.
+type httpd_nss_protocol = string with match(SELF, '^(TLSv1\.[012]|SSLv3)$');
```
kwaegema: @JensTimmerman  please remove ';' as this is not the end of the line and won't compile

jrha:
@JensTimmerman could you address @kwaegema's comment?

JensTimmerman:
@jrha comment addressed. travis fails due to long lines in the tests.

JensTimmerman:
please do check if this is something you would want in quattor by default. I believe these are good strict ciphersuites and ciphers, but these will break support for IE6 and older on clients browsers.

jrha:
We generally run with [Mozilla's Intermediate compatibility TLS settings](https://wiki.mozilla.org/Security/Server_Side_TLS#Intermediate_compatibility_.28default.29) as a default, so I'm pretty sure we're ok with this.

PR #1061, opened 2017-03-04
ncm-aiiserver: add support for AII Grub-2 related options

### LINKED ISSUE #960: aii-server: add support for new options introduced for Grub-based PXE
See https://github.com/quattor/aii/pull/225 and https://github.com/quattor/aii/issues/216 for details.

jouvin:
Unit test added.

Fixes #960


ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan
```diff
@@ -9,10 +9,13 @@ type structure_aiishellfe = {
     'ca_file' ? string
     "cdburl" : type_absoluteURI
     'cert_file' ? string
+    'grub2_efi_kernel_root' ? string
```
stdweird: there's a type for absolute paths, i guess that is what is required.

ncm-aiiserver/src/test/resources/aii-options.pan
```diff
@@ -0,0 +1,30 @@
+@{
+Profile for testing main aii-server optioins
+@}
+object template aii-options;
+
+include 'components/aiiserver/schema';
```
stdweird: just include the config like http://www.quattor.org/documentation/2017/02/10/components-spring17-cleaning.html#add-a-unittest

ncm-aiiserver/src/test/perl/options.t
```diff
@@ -0,0 +1,44 @@
...
+my $comp = NCM::Component::aiiserver->new('ncm-aiiserver');
+my $cfg = get_config_for_profile('aii-options');
+$comp->Configure($cfg);
+
+for my $part (@CONFIG_PARTS) {
+    my $part_config = $cfg->getElement($CONFIG_ROOTS{$part})->getTree();
+    my $config_file = $CONFIG_FILES{$part};
+
+    # Be sure to do the same has the components in term of ordering
+    my $expected_contents = '^#.*\R#.*\R';
```
stdweird: why do you make a regexp? just make the text that is supposed to be generated?

jouvin: Because of the header... I think it makes things unnecessary rigid to try to match the exact text. The regexp reflects what is valid (spaces around `=` are not).

ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan
jouvin: Done. Not sure why GH didn't notice it...

ncm-aiiserver/src/test/perl/options.t
jouvin: Regexp has been improved to match the whole content of the file.

stdweird:
test this please

stdweird:
@jouvin can you rebase one master with #1044 merged?

ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan
stdweird: shouldn't this be an absolute path also?

ncm-aiiserver/src/main/pan/components/aiiserver/schema.pan
```diff
@@ -1,40 +1,41 @@
...
 type structure_aiishellfe = {
-    "cachedir" ? string
-    'ca_dir' ? string
+    "cachedir" ? absolute_file_path
+    'ca_dir' ? absolute_file_path
     'ca_file' ? string
     "cdburl" : type_absoluteURI
     'cert_file' ? string
+    'grub2_efi_kernel_root' ? string
+    'grub2_efi_linux_cmd' ? string with match(SELF, 'linux|linux16|linuxefi')
```
stdweird: make the regexp stricter, like `'^(linux|linux16|linuxefi)$

ncm-aiiserver/src/test/perl/options.t
stdweird: imho this regexp should not test what is valid, it should the result of the component, and e.g. there be no reason why the component changes the spacing around the `=` signs.

jouvin: I agree, I can change this but I thought it was one thing that could change in the future without affecting the test. I will make it stricter (require exactly one space around `='`).

jrha:
Pushing back to 17.4 as we are trying to avoid user-facing changes in 17.3 and this changes types in the schema.

jouvin:
As discussed in https://github.com/quattor/aii/pull/225, the schema cleanup/improvment has been factored out from this PR and is now in #1078.

PR #1062, opened 2017-03-05
ncm-freeipa: mock_rpc: handle JSON boolean string conversion with mixed JSON::PP and JSON::XS environment

stdweird:
Unittests fail on new jenkins environment because there is a mixed JSON::XS and JSON::PP environment (due to loading of `Devel::Cover`).

### DIFF
--- a/ncm-freeipa/src/test/perl/mock_rpc.pm
+++ b/ncm-freeipa/src/test/perl/mock_rpc.pm
@@ -191,7 +191,21 @@ The space after the method name is significant, it is used to determine the used
 
 =cut
 
-sub _format { my $value = shift; return ref($value) eq 'ARRAY' ? join(',', @$value) : $value; }
+# Force booleans to 1/0; prevent mixed string prettification between
+# JSON::PP (0/1) and JSON::XS (true/false)
+sub _format
+{
+    my $value = shift;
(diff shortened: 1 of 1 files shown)


ncm-freeipa/src/test/perl/mock_rpc.pm
```diff
@@ -191,7 +191,21 @@ The space after the method name is significant, it is used to determine the used
...
-sub _format { my $value = shift; return ref($value) eq 'ARRAY' ? join(',', @$value) : $value; }
+# Force booleans to 1/0; prevent mixed string prettification between
+# JSON::PP (0/1) and JSON::XS (true/false)
+sub _format
+{
+    my $value = shift;
+    my $txt;
+    if(ref($value) eq 'ARRAY') {
+        $txt = join(',', @$value);
+    } elsif(JSON::XS::is_bool($value)) {
```
ned21: I see there's no ```use JSON::XS``` at the top of this file. I know this is a unit test, but should there not be one to ensure this subroutine is always available?

stdweird: @ned21 `Test::MockModule->new('JSON::XS')` loads this automatically

PR #1065, opened 2017-03-08
ncm-icinga: check service names for invalid characters

wdpypere:
Some characters are not accepted as a service name ```,:;!?&``` etc. Letting these through to the config will cause icinga to refuse to restart.

PR #1068, opened 2017-03-09
ncm-icinga: start warning about deprecated config options.

wdpypere:
ido2db will not fail if the options are present but will ignore them. Start warning about it. Also make it optional.


wdpypere:
fixed the panlint errors for this PR. the others are fixed in https://github.com/quattor/configuration-modules-core/pull/1065

PR #1070, opened 2017-03-13
ncm-network: cleanup and switch to 00-tqu

stdweird:
Requires https://github.com/quattor/CAF/pull/230 and https://github.com/quattor/maven-tools/pull/160


stdweird:
Still todo: 
[ ] add copy and readdir to CAF::Path (and maven-tools)
[ ] switch to CAF::Path
[ ] add more unittests

stdweird:
Feel free to start reviewing. I'll probably squash all commit before merging, but having these commits might make it easier.

ncm-network/src/main/perl/network.pm
```diff
@@ -1,31 +1,92 @@
...
-$EC=LC::Exception::Context->new->will_store_all;
+=head1 DESCRIPTION
+
+The I<network> component sets the network settings through C<< /etc/sysconfig/network >>
+and the various files in C<< /etc/sysconfig/network-scripts >>.
+
+For restarting, a sleep value of 15 is used to make sure the restarted network
+is fully restarted (routing may need some time to come up completely).
+
+Because of this, adding/changing may cause some slowdown.
```
ned21: "slowdown" == of the ncm-ncd process? Doesn't this also cause a network interruption?

ncm-network/src/main/perl/network.pm
```diff
@@ -1,31 +1,92 @@
...
+
+New/changed settings are first tested by probing the CDB server on the port
+where the profile should be found. If this fails, the previous settings are reused.
```
ned21: "If this fails, the previous settings are reused." => "If this fails, the component reverts all settings to the previous values."

ncm-network/src/main/perl/network.pm
```diff
@@ -1,31 +1,92 @@
...
+
+The I<network> component sets the network settings through C<< /etc/sysconfig/network >>
+and the various files in C<< /etc/sysconfig/network-scripts >>.
+
+For restarting, a sleep value of 15 is used to make sure the restarted network
+is fully restarted (routing may need some time to come up completely).
+
+Because of this, adding/changing may cause some slowdown.
+
+New/changed settings are first tested by probing the CDB server on the port
```
ned21: Is it really a "probe" or is full profile download performed/required for success?

ncm-network/src/main/perl/network.pm
```diff
@@ -577,348 +582,276 @@ sub gather_existing
...
+    };
 
-    # current setup, will be printed in case of major failure
-    my $init_config = $self->get_current_config();
+}
 
-    my $mac2dev = $self->make_mac2dev();
+# return anonymous sub that calls
+# _make_ifcfg_line with first arg $href
+# and appends result it not empty string to arrayref
```
ned21: it => if ?

ncm-network/src/main/perl/network.pm
```diff
@@ -182,82 +183,95 @@ sub ethtool_get_current
...
+{
+    my ($self, $file) = @_;
+    return $self->backup_filename($file) . $FAILED_SUFFIX;
 }
 
-# writes some text to file, but with backup etc etc it also
-# checks between new and old and return if they are changed
-# or not
+# Make copy of original file (if exists) for testing (name from testcfg_filename).
+# (This is not the backup created by mk_bu; and this method makes not backup).
```
ned21: makes *no* backup?

ned21:
@stdweird - I took a first pass through and nothing bad jumped out. This is super-complex code with which I'm not familiar though. :(

@gombasg - do we have our patches to the 1.2.5 version easily accessible? Perhaps we can get the functionality addressed as part of this rewrite rather than trying to reimplement them ourselves later?

stdweird:
@ned21 @gombasg i'd rather not add any new features to this cleanup PR. and th ePR isn't finished yet. still need to switch to CAF::Path...
open an issue, maybe include the patches, i'll see what can be done.

ncm-network/src/main/perl/network.pm
stdweird: yes, of `ncm-ncd`. the network should already be up at this point. i'll rephrase to
```
Because of this additional wait time, adding/changing may cause some delay in the ncmncd run.
```

ncm-network/src/main/perl/network.pm
stdweird: it's a full download, not a test or a probe. i'll rephrase to
```
New/changed settings are first tested by retrieving the latest profile from the CDB server (using ccm-fetch).
```

ncm-network/src/main/perl/network.pm
ned21: I would suggest: "may cause the ncm-ncd run to take longer than usual". I think we should also highlight that any change to the network settings can lead to a brief network interruption.

stdweird:
we have started testing this.

the only thing planned and not done is the `implement copy in CAF::Path`. i've worked around it using source/reopen, but that a lot less elengant than originally anticipated. however, it's NoAction safe (just a bit verbose one the logging) and uses CAF::Path (and helped uncover some CAF::FileWriter issues 😄 ).

this PR is also not yet tagged as NoActionSupported. will add that in a follow up PR so it can be tested in its current starte.

ned21:
I've slightly lost track of where we are with the review but it looks like just the only outstanding change I've requested is:

- [ ] network.pm lines 15, 17 and 18 - reword doc

stdweird:
@ned21 if there are no further remarks, feel free to close this PR. i'll modify #1087 (something is wrong with the history)

PR #1075, opened 2017-03-15
ncm-spma (yum): Allow unmanaged repositories to be controlled seperately from packages

jrha:
This allows spma to cooperate with other tools (e.g. RedHat subscription manager) that configure repositories essential to the system. The implementation matches what I have done for APT in #821 based on an idea originally proposed by @ned21.

This is backwards incompatible as the the behaviour of yum based spma on systems that currently have `userpkgs` set will change unless `userrepos` is also set.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/yum/schema.pan
+++ b/ncm-spma/src/main/pan/components/spma/yum/schema.pan
@@ -65,6 +65,8 @@ type component_spma_yum = {
     "quattor_os_release" ? string # string to write to quattor_os_file
     "plugins" ? spma_yum_plugins
     "main_options" ? spma_yum_main_options
+    @{ Allow user defined (i.e. unmanaged) repositories to be used }
+    "userrepos" : boolean = false
 };
 
 bind "/software/components/spma" = component_spma_yum;
--- a/ncm-spma/src/main/perl/spma/yum.pm
+++ b/ncm-spma/src/main/perl/spma/yum.pm
@@ -109,7 +109,7 @@ sub _match_noaction_tempdir
 # in the system that are not listed in $allowed_repos.
 sub cleanup_old_repos
 {
-    my ($self, $repo_dir, $allowed_repos, $allow_user_pkgs) = @_;
+    my ($self, $repo_dir, $allowed_repos, $allow_user_repos) = @_;
 
     if ($NoAction) {
         if($self->_match_noaction_tempdir($repo_dir)) {
@@ -128,7 +128,7 @@ sub cleanup_old_repos
     }
 
     # Test this after the NoAction bit for unittesting
-    return 1 if $allow_user_pkgs;
+    return 1 if $allow_user_repos;
 
     my $dir;
     if (!opendir($dir, $repo_dir)) {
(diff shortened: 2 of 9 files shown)


jrha:
While we're breaking things we could also change `userpkgs` to a boolean.

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -1138,11 +1138,11 @@ sub Configure
     $t->{run} = $t->{run} eq 'yes';
     $t->{userpkgs} = defined($t->{userpkgs}) && $t->{userpkgs} eq 'yes';
 
-    # When userpkgs are allowed, there is no control over what is in the reposdir
-    # If they are not allowed, and retry is allowed, use a non-standard location
-    # to avoid any repositories getting added during the retries (e.g. from rpms).
-    my $main_repos_dir = ($NoAction || $t->{userpkgs} || (!$t->{userpkgs_retry})) ?
-                             REPOS_DIR_DEFAULT : REPOS_DIR_QUATTOR;
+    # When userrepos are allowed, there is no control over what is in the reposdir
```
stdweird: i wouldn't implement it like this.
yum.conf can have more than one dir in the reposdir entry, so you can have a fully managed one by quattor (eg /etc/yum.quattor.repos.d`) and a one by something external.

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -65,6 +65,8 @@ type component_spma_yum = {
     "quattor_os_release" ? string # string to write to quattor_os_file
     "plugins" ? spma_yum_plugins
     "main_options" ? spma_yum_main_options
+    @{ Allow user defined (i.e. unmanaged) repositories to be used }
+    "userrepos" : boolean = false
```
ned21: I think user_repos would be more in keeping with the other options above?

ncm-spma/src/main/perl/spma/yum.pm
ned21: But is that normal/expected? Although looking further down, isn't that what actually happens?

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -1138,11 +1138,11 @@ sub Configure
...
+    # If they are not allowed, use a non-standard location to avoid any repositories
+    # getting added during the retries (e.g. from rpms).
+    my $main_repos_dir = ($NoAction || $t->{userrepos}) ? REPOS_DIR_DEFAULT : REPOS_DIR_QUATTOR;
```
ned21: What happened to the userpkgs_retry check? i.e. Why was that previously needed and not now?

jnovy:
@jrha - the whitelist approach sounds definitely better than to allow unmanaged repo definitions. In yumng we already use a similar approach for packages - package whitelist to accommodate 3rd party installers, e.g. Satellite using Puppet to pull in and install its packages.

Whitelisting will also prevent people from installing unwanted/untracked packages.

stdweird:
main `yum.conf` support `reposdir`, which can handle multiple diretcories. one fully controlled by quattor/spma, and others managed by other tools. you can make one and symlink whatever you trust form other sources.

jrha:
@jrha please respond to comments

jrha:
For RHEL we have taken the simpler/cleaner option of directly including the RHSM generated repo file directly from `yum.conf`.

This only requires adding one option to the schema (we probably should add some validation):
```pan
type spma_yum_main_options = {
    "exclude" ? string[]
    "installonly_limit" ? long(0..) = 3
    "keepcache" ? boolean
    "retries" ? long(0..) = 10
    "timeout" ? long(0..) = 30
    "include" ? string
};
```

So that we can then set:
```pan
​'/software/components/spma/main_options/include' = '/etc/yum.repos.d/redhat.repo';
```

I still think it makes sense to revisit how `reposdir` gets set, but I don't have the bandwidth at the moment.

jrha:
Abandoning this for now.

Issue #648, opened 2015-12-09
ncm-dirperm modernisation


ned21:
ncm-dirperm suffers from some of the same problems as ncm-symlink documented in #288 #380 #502 #503 #593:

1. Paths stored in a list with no checking for duplicates/conflicts at the leaf, or any depth of the tree.
1. No dryrun (NoAction) mode.
1. No unit tests.
1. checkmount implementation is not aware of filesystems that may be mounted or managed outside of /etc/fstab and the Quattor profile.
1. No ability to delete files as well as create them.

One option that would be to write a new component that merges the functionality of dirperm and symlink and addresses all the open issues in one go. Opening this here to encourage discussion at hte next workshop. It would also be useful to record the state of the art on how to unit test low level file manipulation operations currently done by LC::* functions since overhauling that area may be a predecessor task to this one.

stdweird:
all the lowlevel stuff should be moved to CAF, and `dirperm` and `symlink` should call whatever that CAF module provides. the only reason why there should a new component is a new schema.
unittests in the components shouldn't need to worry about lowlevel stuff, everything should be mocked. in the CAF repo, there should be unittests for actual testing both NoAction and no NoAction.

jouvin:
With `CAF::Path` now covering all the major file/directory operations including symlinks and hardlinks, it may be a good time to start reimplementing these configuration modules.

jouvin:
Merging `ncm-symlink` and `ncm-dirperm`:
* Pro: main logic is the same, would make cross validation of configuration much easier (detecting conflicts between dirperm and symlink configurations). And will avoid the final configuration to depend on the order of execution of these two configuration modules.
* Cons: schema pretty different.

jouvin:
Move to `dict` rather than `list`: agreement this is desirable. May look at `HashOrdered`Perl package but not sure it is available on all the platforms we support.

jouvin:
Proposed name for the new component: `ncm-path`

jrha:
Discussed at March 2017 workshop, agreed to create new component named ncm-path.

PR #1076, opened 2017-03-21
ncm-path: add new component skeleton code

stdweird:
Partial fix for #648


stdweird:
@ned21 merge it, and make a PR with the schema you want

PR #1077, opened 2017-03-22
CCM cleanup fallout

stdweird:
After https://github.com/quattor/CCM/pull/162 usage of `CCM::Resource` and direct class testing with old name are not supported anymore


stdweird:
retest this please

PR #1085, opened 2017-04-07
ncm-network: fixes and new features

stdweird:
Requires #1070

PR #1087, opened 2017-04-10
ncm-network: cleanup, fixes and new features

### LINKED ISSUE #870: ncm-network: hostname is not set in el7
In el7, hostname should be set through either:
-  **/etc/hostname** [requires reboot] 
-  **hostnamectl set-hostname --static MYHOSTNAME** [requires restart of hostnamed].

Right now, it means the hostname applied is the one from **ks**. This poses problems when you install through private network, but want a public hostname.
One service particularly affected is cdp-listend, which listens on the hostname ip, which can be in some cases different from the one notified.

### LINKED ISSUE #1082: ncm-network: Allow extra routing options for the default route
For rto_min and similar can be handled via a custom /etc/sysconfig/network-scripts/route-eth0.  However you can only add new route entries using /etc/sysconfig/network-scripts/route-eth0, modifying existing entries (including the default route) is not possible. 

Setting `DEFROUTE=no` in `/etc/sysconfig/network-scripts/ifcfg-eth0` prevents the default route being added which allows you to set a custom default route in `/etc/sysconfig/network-scripts/route-eth0 `(including with rto_min)

- [ ] Modify `structure_interface` in `aquilon/components/network/core-schema.tpl`  to accept ```defroute=yes|no```
- [ ] update ncm-network to add DEFROUTE setting to ifcfg-ethX

### LINKED ISSUE #1109: ncm-network: systemd rpm scripts remove HOSTNAME from /etc/sysconfig/network
on systemd install/upgrade, `/etc/sysconfig/network` is checked for the HOSTNAME variable and if it exists and `/etc/hostname` doesn't, it is used to created /etc/hostname. and then the varaible is always removed.

but this garantees a network restart the next time `ncm-network` runs

i do not know since when the systemd rpm started to do this, but please be warned.

i'll try to come up with a fix of some sort, but very likely only in the rewrite.

stdweird:
- cleanup and switch to 00-tqu
- support route/route6/rule/rule6 in new format
- fixes #1082, fixes #870 and fixes #1109 
- improved ovs support
- support VXLAN via plugin script (eg http://193.169.5.193/centos/7/base/SRPMS/initscripts-vxlan-0.1-1.el7.centos.src.rpm)

- Replaces #1085 and replaces #1070
- Requires https://github.com/quattor/CAF/pull/230


stdweird:
@ned21 see https://github.com/quattor/configuration-modules-core/pull/1087/commits/fdf05a1d99f159be67735671548fbf386017f69c for rephrased DESCRIPTION.

jrha:
Does this introduce any potential for backwards incompatibility?

stdweird:
@jrha it should not. unless you rely on the existence of HOSTNAME in /etc/sysconfig/network, which is removed in el7 due to https://github.com/quattor/configuration-modules-core/issues/1109

stdweird:
retest this please

jrha:
Ok, I'll leave the flag for the release notes.

stdweird:
jenkins failed due to new `NetAddr::IP` dependency. is installed now

stdweird:
hmm, merging the CAF PRs doesn't trigger the main CAF jenkins project to run, so the new CAF wasn't yet installed.

stdweird:
retest this please

jrha:
I'm leaning towards moving this to 17.8 — I don't feel like I have time to review it properly right now... any objections?

stdweird:
@jrha i don't care that much, we run it already. #1109 is particularly nasty imho, but this is a big change overall. 
people who want to run this, can do so with https://jenkins0.ugent.be/job/NCM_Components-pull_requests/2077/org.quattor.cfg.module$network/ (has the rpm and `-cfg.tar.gz` has the templates)

ned21:
We are still stuck on an ancient fork of ncm-network so don't have any opinion on its inclusion.

stdweird:
@ned21 feel free to open an issue with the diff of your fork against some related release, interesting features are always welcome

ned21:
We have one internally, waiting for @gombasg  to remember what they are and figure out if they are even still needed. :-)

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -120,8 +120,8 @@ type structure_interface = {
     "netmask" ? type_ip
     "broadcast" ? type_ip
     "driver" ? string
-    "bootproto" ? string
-    "onboot" ? string
+    "bootproto" ? string # TODO: only allow match(SELF, '^(static|bootp|dhcp)')
+    "onboot" ? string # TODO: only allow match(SELF, '^(yes|no)$')
```
jrha: Or use a real boolean?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -120,8 +120,8 @@ type structure_interface = {
     "netmask" ? type_ip
     "broadcast" ? type_ip
     "driver" ? string
-    "bootproto" ? string
-    "onboot" ? string
+    "bootproto" ? string # TODO: only allow match(SELF, '^(static|bootp|dhcp)')
```
jrha: Why wait?

ncm-network/src/main/perl/network.pm
```diff
@@ -434,118 +433,115 @@ sub make_mac2dev
...
 
-        # collect /hardware info
-        my $hw_path = "/hardware/cards/nic/".$ifacename;
+        # Handle hardware address
+        my $nicname = $ifname;
 
-        # CERN nic = list patch
-        if (! $config->elementExists($hw_path)) {
-            $hw_path = "/hardware/cards/nic/".$1 if ($ifacename =~ m/^eth(\d+)/);
+        # TODO: can we get rid of this?
```
jrha: I don't see why not, the schema enforces a `dict` here.

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -97,7 +99,7 @@ type structure_ethtool_ring = {
             s  Enable SecureOn(tm) password for MagicPacket(tm)
             d  Disable (wake on nothing).  This option clears all previous option
 }
-type structure_ethtool_wol = string with match (SELF, '^p|u|m|b|a|g|s|d$');
+type structure_ethtool_wol = string with match (SELF, '^[pumbagsd]+$');
```
jrha: This changes what passes validation, for example `pppppppp` would now pass. Is this okay?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -145,6 +145,40 @@ type structure_ethtool = {
...
+    @{VXLAN Network Identifier (or VXLAN Segment ID); derived from devicename vxlan[0-9] if not defined}
+    'vni' ? long(0..) # up to 16M
+    @{multicast ip to join}
+    'group' ? type_ip
+    @{destination IP address to use in outgoing packets}
+    'remote' ? type_ip
+    @{source IP address to use in outgoing packets}
+    'local' ? type_ip
+    @{UDP destination port}
+    'dstport' ? long(0..)
```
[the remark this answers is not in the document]
jrha: Hmm... in the general case it should probably be `(2..65535)` to avoid use of magical wild-card ports.

Should this default to either the linux default of `8472` or the IANA assigned `4789`?

jrha:
Fixes for commit messages:

* add the ~used~ commands **used** for collecting the current state to the gathered info
* cleanup: reform DEVICE_REGEXP to extended regexp with some ~explanantion~**explanation**
* cleanup: factor out ~clenaup~**clean-up** of backup and test config and add stopstart sanity check

jrha:
@stdweird the test results don't seem to stay around on jenkins for very long, is this intentional?

ncm-network/src/main/perl/network.pm
```diff
@@ -1,1183 +1,1728 @@
...
+            if ($route->{prefix}) {
+                $ip = NetAddr::IP->new("$route->{address}/$route->{prefix}");
             } else {
-                $self->error("vlan device with vlan[0-9]{0-9} naming convention need physdev set.");
+                # in absence of netmask, NetAddr::IP uses 32 or 128
+                $ip = NetAddr::IP->new($route->{address}, $route->{netmask});
             }
+            # Generate it
+            $route->{command} = "$ip";
+            $route->{command} .= " via $route->{gateway}" if $route->{gateway};
```
jrha: If the route doesn't have a gateway it should interpreted as a direct (link scoped) route and be followed by the device name, e.g.
```
10.16.184.0/22 dev eth0
```
or optionally:
```
10.16.184.0/22 dev eth0 scope link
```

i.e. it still has to be valid against the `ip route add` command schema

This preserves the previous behaviour and we rely on this working!

jrha: See also: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/6/html/Deployment_Guide/sec-Configuring_Static_Routes_in_ifcfg_files.html#sec-networkscripts-static-routes-ip-command

jrha:
Just tested this on a SL7 test node
1. Static routes broke (see in-line comment)
2. `"/system/network/interfaces/eth0/set_hwaddr" = true;` but the `HWADDR` is removed from `sysconfig/network-scripts/ifcfg-eth0`.

ncm-network/src/main/perl/network.pm
```diff
@@ -1,1183 +1,1728 @@
...
+                $self->verbose("No nic found for $ifname, using nic-as-list name $nicname");
+            };
+        };
+
+
+        # Each iface has set_hwaddr
+        # If set_hwaddr is true, hwaddr must be set
+        my $nic = $nics->{$nicname} || {};
+        my $mac = $nic->{hwaddr};
+        $iface->{set_hwaddr} = $set_hwaddr_default;
```
jrha: Doesn't this override the profile regardless?
Commenting it out restores the `HWADDR` line on my test node.

jrha:
Here's my bodge:

```diff
711c711
<         $iface->{set_hwaddr} = $set_hwaddr_default;
---
>         #$iface->{set_hwaddr} = $set_hwaddr_default;
1034c1034
<     my ($self, $flavour, $routes) = @_;
---
>     my ($self, $flavour, $routes, $device) = @_;
1048c1048,1049
<             $route->{command} .= " via $route->{gateway}" if $route->{gateway};
---
>             #$route->{command} .= " via $route->{gateway}" if $route->{gateway};
>             $route->{command} .=  $route->{gateway} ? " via $route->{gateway}" : " dev $device";
1596c1597
<                 my $text = $self->$method($flavour, $iface->{$flavour});
---
>                 my $text = $self->$method($flavour, $iface->{$flavour}, $ifacename);

```

ncm-network/src/main/perl/network.pm
stdweird: the docs have `This is only required if the gateway is not set via DHCP and is not set globally in the /etc/sysconfig/network file` 
so you have no default gateway? or it is not via the device?

ncm-network/src/main/perl/network.pm
```diff
@@ -9,13 +9,16 @@ network: Configure Network Settings
 The I<network> component sets the network settings through C<< /etc/sysconfig/network >>
 and the various files in C<< /etc/sysconfig/network-scripts >>.
 
-For restarting, a sleep value of 15 is used to make sure the restarted network
+New/changed settings are first tested by retrieving the latest profile from the
+CDB server (using ccm-fetch).
+If this fails, the component reverts all settings to the previous values.
+
+During this test, a sleep value of 15 is used to make sure the restarted network
```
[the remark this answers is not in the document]
stdweird: you want me to add `seconds` or you think 15 seconds is excessive?

ncm-network/src/main/perl/network.pm
stdweird: what is your default set to in the profile? (i need to know if this is a bug or some backwards incompatible change).

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -6,12 +6,45 @@ include 'pan/types';
...
     "address" ? type_ip
-    "netmask" ? type_ip
+    @{The PREFIX in ADDRESS/PREFIX via GATEWAY}
+    "prefix"  ? long(0..)
+    @{The GATEWAY in ADDRESS/PREFIX via GATEWAY}
     "gateway" ? type_ip
+    @{alternative notation for prefix (cannot be combined with prefix)}
+    "netmask" ? type_ip
+    @{full command used to add route (cannot be combined with other options)}
+    "command" ? string
```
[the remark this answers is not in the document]
stdweird: hmmm, no afaik, there are possible options that are not part of (or shouldn't be) an absolute_file_path?

ncm-network/src/main/pan/components/network/core-schema.pan
stdweird: heh, dunno, but i'll change it back to old regex. should be replaced with `choice` type anyway at some point

ncm-network/src/main/pan/components/network/core-schema.pan
stdweird: ok, i'll chnage this in a separte commit (it just introduces more backwards incompatibility)

stdweird:
@jrha @ned21 @jouvin just to be clear: when running  the new component, any changes to network (full restart or stop/start of interface etc ) is either a bug in the old component or the new one.
in our case, no nodes had any network interruption when switching from old to new component; and there shouldn't be any such case (unless actual bug). the backwards incompatible flag only applies to the schema. testing is much appreciated.

jrha:
@stdweird here's the network configuration for that test vm (sanitised)...
`# ccm --format yaml /system/network/`
```yaml
default_gateway: 10.16.216.254
domainname: example.org
hostname: vm273
interfaces:
  eth0:
    bootproto: static
    broadcast: 10.16.223.255
    fqdn: vm273.example.org
    gateway: 10.16.216.254
    ip: 10.16.221.214
    netmask: 255.255.248.0
    network_environment: internal
    network_type: wan
    route:
    - address: 10.16.176.0
      netmask: 255.255.248.0
    - address: 10.16.216.0
      netmask: 255.255.248.0
    - address: 10.16.184.0
      netmask: 255.255.252.0
    - address: 10.16.208.0
      gateway: 10.16.187.30
      netmask: 255.255.248.0
    - address: 172.16.176.0
      netmask: 255.255.248.0
    - address: 169.254.0.0
      netmask: 255.255.0.0
    set_hwaddr: true
ipv6:
  enabled: false
nameserver:
- 10.16.72.21
- 10.16.8.13
- 10.16.56.240
primary_ip: 10.16.221.214
routers:
  eth0:
  - 10.16.216.254
```

jrha:
It also complained that there was no default gateway, but there clearly is.
```
2017/09/06-16:41:24 [INFO] No default gateway configured. Found first gateway 10.16.216.254 on interface eth0
```

ncm-network/src/main/perl/network.pm
jrha: The default is set to false, see below.

ncm-network/src/main/perl/network.pm
jrha: This is not for setting default routes or "normal" routes via a gateway, we use this to allow access to multiple directly connected overlaid subnets, many of which the host itself may not have an IP for. In this case the host is essentially it's own gateway (aka. the route is link scoped).
The current component handles this correctly in the legacy route file format.

ncm-network/src/main/perl/network.pm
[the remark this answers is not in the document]
jrha: Sorry, I meant you should specify the units :-)

ncm-network/src/main/pan/components/network/core-schema.pan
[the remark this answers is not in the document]
jrha: What does "full command" mean then?

ncm-network/src/main/pan/components/network/core-schema.pan
jrha: Ok

ncm-network/src/main/pan/components/network/core-schema.pan
jrha: The TODO could be to use a real boolean (or the transitional type) but I don't like leaving too many TODOs lying around

ncm-network/src/main/pan/components/network/core-schema.pan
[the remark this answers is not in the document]
stdweird: the options passed to `/sbin/ip route add`, in case you want to do some non-standard stuff
actually, we should probably sanitize the line; eg make usre there's no `;` or escaped newlnies or something bash magical...

ncm-network/src/main/perl/network.pm
stdweird: ok, this looks like a trivial bug. will fix in a few minutes

ncm-network/src/main/pan/components/network/core-schema.pan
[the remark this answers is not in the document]
jrha: Ah ok, yes... probably needs to be sanitised.

ncm-network/src/main/perl/network.pm
stdweird: nvm, this is a stupid bug

stdweird:
@jrha issues addressed. the `dev eth0` fix will trigger a restart of interfaces with this PR already deployed (but not with ncm-network from previous releases)

jrha:
Much better! :grinning: 

Still getting `No default gateway configured` despite the gateway being configured…
```
[INFO] No default gateway configured. Found first gateway 130.246.216.254 on interface eth0
```

stdweird:
@jrha makes sense, since i forgot to do that one 😄

ncm-network/src/main/perl/network.pm
```diff
@@ -1,1183 +1,1728 @@
...
-    #
-    my $missing_default_gateway_autoguess = 1;
-    if ($config->elementExists($path."/guess_default_gateway")) {
-        if ($config->getValue($path."/guess_default_gateway") eq "false") {
-            $missing_default_gateway_autoguess = 0;
-        }
+    my $guess_dgw = defined($nwtree->{guess_default_gateway}) ? $nwtree->{guess_default_gateway} : 1;
+
+    my $nodgw_msg = "No default gateway configured";
+    my $dgw = $nwtree->{default_gateway};
```
jrha: Is this ever used? It looks like it is simply overwritten if `$guess_dgw` is true.

ncm-network/src/main/perl/network.pm
```diff
@@ -1,1183 +1,1728 @@
...
+    if ($guess_dgw) {
```
jrha: Something like this?
```perl
if (!$dgw && $guess_dgw) {
```

stdweird:
@jrha fixed

jrha:
:tada:

ned21:
> @jrha @ned21 @jouvin just to be clear: when running the new component, any changes to network (full restart or stop/start of interface etc ) is either a bug in the old component or the new one.

@stdweird - are you saying that with the new component, a change to a file no longer triggers  a full network restart which might cause an interface to go bounce? Or just that upgrading the component in place does not cause any network restart provided there was also no configuration change?

jrha:
My understanding (which matches what I have seen in testing) is the latter.
Upgrading the component (with no other changes made) should not restart the network.

stdweird:
@ned21 the latter, because quite some files do get modified (eg route files move from old to new format)

but full network restarts should not occur that often anyway. the only time full network is restarted is when `/etc/sysconfig/network` is modified (maybe something else is modifiying it, eg systemd rpm or another component).
interfaces are mostly stopped/started on their own, i guess this is what you mean?

with the rewrite, we can probably insert more control so admins can decide what ncm-network should do: restart or not (if you feel lucky or something like that. hard to test new network settings without making modifications);
maybe we can add more logic so certain types of changes can be handled in a cleaner way (if people want to reverse engineer initscripts, be my guest. but if eg a route changes, we could call `ifroute` rather then `ifup`).

we are aware that things can go wrong. we have seen part of routing tables dissapear for no reason (routes defined on vlans; we think bug in initscripts), or we also know that we need to reload shorewall after any change to the network (some services do not like changes to the network).

if you can give more details what you see happening, it's always useful as input for more features.

ned21:
> interfaces are mostly stopped/started on their own, i guess this is what you mean?

tbh I no longer remember. We definitely have custom routes that disappeared (which we could probably move from our custom perl script to quattor control), but we also have things like cluster VIPs which are controlled by the HA daemon and not the OS. 

I think we have decided that we are content to simply not run the component between reboots.

jrha:
I've now tested the new component code on one of our pre-prod HA load-balancers and happily it does not remove the cluster IP if there are no other changes, the old code removes the VIP regardless.

stdweird:
and i found a bug with adding the `dev abc` to the routes.

stdweird:
@jrha issue fixed, part of https://github.com/quattor/configuration-modules-core/pull/1087/commits/cf5548669eb71083ea735f148644ea5d8e69558d

jrha:
@stdweird which issue?

stdweird:
the previous fix to add `dev abc` to each route added the interface name, not the actual device name

jrha:
:ok_hand:

jrha:
From my point of view I am happy with the behaviour now, I'll try to do another review this afternoon.

PR #1093, opened 2017-04-18
ncm-ofed: add support for opensm partitions

stdweird:


ncm-ofed/src/main/pan/components/ofed/schema.pan
```diff
@@ -127,7 +128,50 @@ type component_ofed_openib = {
...
+type component_ofed_opensm = {
+    @{daemons to restart on configuration changes}
+    "daemons" : string[] = list('opensmd')
+    @{SM partitions configuration. Dict key is the partition name}
+    "partitions" ? component_ofed_partition{}
+    @{Node name map configuration. Dict key is the GUID starting with 'x' (the 0 is prefixed automatically)}
+    "names" ? string{}
+} with {
+    if (exists(SELF['names'])) {
+        foreach (guid;descr;SELF['names']) {
```
jrha: panlint says:
> Semicolons should be followed exactly one space or end-of-line
>
> ```pan
>     foreach (guid;descr;SELF['names']) {
>                   ^     ^
> ```

ncm-ofed/src/main/pan/components/ofed/schema.pan
```diff
@@ -127,7 +128,50 @@ type component_ofed_openib = {
     true;
 };
 
+type component_ofed_partition_property = {
+    @{Port GUID}
+    'guid' : string with match(SELF, '^(ALL(_(SWITCHES|V?CAS|ROUTERS))?|SELF|0x[0-9a-fA-F]{1,10})$')
+    'membership' ? string with match(SELF, '^(limited|full|both)$')
```
jrha: Could `membership` benefit from an annotation explaining it's purpose?

ncm-ofed/src/main/perl/ofed.pm
```diff
@@ -13,31 +13,81 @@ use EDG::WP4::CCM::TextRender;
...
+                                   $PARTITIONS_CONF_TT,
+                                   $PARTITIONS_FILENAME)
+            || 0;
+    }
+
+    if ($tree->{names}) {
+        $changed += $self->_render($config->getElement($self->prefix().'/opensm/names'),
+                                   $NAMES_CONF_TT,
+                                   $NAMES_FILENAME)
+            || 0;
```
jrha: IMHO this is much more readable.

```perl
    $changed += $self->_render(
        $config->getElement($self->prefix().'/opensm/names'),
        $NAMES_CONF_TT,
        $NAMES_FILENAME,
    ) || 0;
```

ncm-ofed/src/main/pan/components/ofed/schema.pan
stdweird: probably not. i added it here because it can be configured, but i only sortof understand what it is supposed to do, and we typically don't set it 😄

jrha: fair enough!

ncm-ofed/src/main/perl/ofed.pm
stdweird: i'll change it, but i don't care. someone who does should come up with perl tidy rules for stuff like this

stdweird:
@jrha done

jrha:
I CAN SEE JENKINS! :eyes: :fireworks: :tada: :balloon:

PR #1101, opened 2017-04-25
ncm-authconfig : Update ldap.tt (Issue #1100)

### LINKED ISSUE #1100: ncm-authconfig : no newline after krb5_canonicalize option
When I do an ncm-ncd --configure authconfig, the /etc/sssd.conf file contains the following line:
Krb5_canonicalize = Falsekrb5_realm = LAPP.IN2P3.EN
While it should be:
Krb5_canonicalize = False
Krb5_realm = LAPP.IN2P3.EN

This is due to an error in the file /usr/share/templates/quattor/authconfig/domains/ldap.tt

[% - SWITCH name -%]
[% - CASE ['krb5_canonicalize'] -%]
[% - desc. $ Name? 'True': 'False'%]
**[% - CASE -%]**
[% - desc. $ Name%]
[% END -%]

  should be

[% - SWITCH name -%]
[% - CASE ['krb5_canonicalize'] -%]
[% - desc. $ Name? 'True': 'False'%]
**[% CASE -%]**
[% - desc. $ Name%]
[% END -%]

seraf1:
When I do an ncm-ncd --configure authconfig, the /etc/sssd.conf file contains the following line:
Krb5_canonicalize = Falsekrb5_realm = LAPP.IN2P3.EN
While it should be:
Krb5_canonicalize = False
Krb5_realm = LAPP.IN2P3.EN

This is due to an error in the file /usr/share/templates/quattor/authconfig/domains/ldap.tt

[% - SWITCH name -%]
[% - CASE ['krb5_canonicalize'] -%]
[% - desc. $ Name? 'True': 'False'%]
[% - CASE -%]
[% - desc. $ Name%]
[% END -%]

  should be

[% - SWITCH name -%]
[% - CASE ['krb5_canonicalize'] -%]
[% - desc. $ Name? 'True': 'False'%]
[% CASE -%]
[% - desc. $ Name%]
[% END -%]

Fixes #1100

### DIFF
--- a/ncm-authconfig/src/main/pan/components/authconfig/sssd/ldap.pan
+++ b/ncm-authconfig/src/main/pan/components/authconfig/sssd/ldap.pan
@@ -95,7 +95,7 @@ type authconfig_sssd_ldap = {
 
     "krb5_backup_server" ? string
     "krb5_canonicalize" ? boolean
-    "krb5_realm" ? string[]
+    "krb5_realm" ? string
     "krb5_server" ? string
     "access_filter" ? string
     "access_order" : ldap_order = "filter"
@@ -122,6 +122,5 @@ type authconfig_sssd_ldap = {
     "referrals" : boolean = true
     "rootdse_last_usn" ? string
     "search_timeout" : long = 6
-    "use_object_class" : string = "posixAccount"
     "account_expire_policy" ? string with match(SELF, "^(shadow|ad|rhds|ipa|389ds|nds)$")
 };
--- a/ncm-authconfig/src/main/pan/components/authconfig/sssd/user.pan
+++ b/ncm-authconfig/src/main/pan/components/authconfig/sssd/user.pan
@@ -12,6 +12,7 @@
 declaration template components/authconfig/sssd/user;
 
 type sssd_user = {
+    "object_class" : string = "posixAccount"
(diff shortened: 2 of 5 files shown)


ned21:
OK to test.

@seraf1 - thanks for the PR. Since GitHub may not live as long as the commit history, would you mind amending the commit to describe the fix in more detail? Just a simple 1-2 sentence overview is fine, something that tells an reader what has changed and why without needing to load up a browser and surf to GH. If you end with:

Fixes: #1100 

it will cause GH to auto-close that issue when this is merged.

Thanks again!

seraf1:
I do not know how to change a commit on GH, can you explain the manipulation to do?
Thanks

stdweird:
@seraf1 can you add a unittest for this? add entry in https://github.com/quattor/configuration-modules-core/blob/master/ncm-authconfig/src/main/resources/tests/profiles/ipa.pan and add additional regexp line in https://github.com/quattor/configuration-modules-core/blob/master/ncm-authconfig/src/main/resources/tests/regexps/ipa/value

stdweird:
@seraf1 https://help.github.com/articles/changing-a-commit-message/
you basically change the commit message locally and then you use `push --force` to update it on github

PR #1106, opened 2017-04-26
ncm-metaconfig: add openvpn as replacement for ncm-openvpn

wdpypere:
I needed a fix for https://github.com/quattor/configuration-modules-core/issues/1104

As i'm no perl expert and saw that ncm-openvpn is on the "replace by metaconfig" list, here is the metaconfig openvpn config.

PR #1108, opened 2017-04-27
ncm-openvpn: push server attribute requires double-quoted value in recent versions

stdweird:


stdweird:
can be closed if #1106 is merged first (but that one needs more testing first), but we need a fix asap
@wdpypere can you followup on this?

Issue #1029, opened 2017-02-09
ncm-nfs: why restart nfs when fstab is updated? 


stdweird:
currently, on modifcations of fstab, a restart of the `nfs` service is triggered. why is this the case?
similar, is no exports are configured by `ncm-nfs`, why restart nfs at all?

this is flagged as a bug because it reports an error, yet nothing is wrong

jrha:
Removing milestone until a fix is proposed.

PR #1110, opened 2017-04-28
ncm-nfs: stop nfs reload when adding comments to empty exports file

drossy:
On ncm-nfs's first run, it adds comment lines to the exports file. Even If there are no exports this will trigger a reload of the nfs service. On RHEL6 this appears harmless as the reload returns 0, but on RHEL 7, systemd returns an error:

 `# /bin/systemctl reload nfs; echo $?`
`Job for nfs-server.service invalid. `
`1`

This patch attempts to stop this. I think should partially address #1029 but any changes to fstab may also make the component reload nfs.

### DIFF
--- a/ncm-nfs/src/main/perl/nfs.pm
+++ b/ncm-nfs/src/main/perl/nfs.pm
@@ -181,6 +181,11 @@ sub exports
 {
     my ($self, $tree) = @_;
 
+    # Take the original size of the exports file
+    my $backupsize = -s $EXPORTS;
+    # Initialise the flag for whether only comments are being added
+    my $onlycommentsadded = 0;
+    
     my $fh = CAF::FileWriter->new($EXPORTS, backup => ".old", log => $self);
 
     # Do not include timestamp, otherwise file changes every run
@@ -216,11 +221,19 @@ sub exports
     } else {
         $self->verbose("No exports defined");
         print $fh "# No exports defined\n";
+        if (defined($backupsize) && $backupsize == 0) {
+            $onlycommentsadded = 1;
+        }
     }
 
     my $result = $fh->close();
     if ($result) {
-        $self->info("$EXPORTS created/updated");
+        if (! $onlycommentsadded) {
+            $self->info("$EXPORTS created/updated");
+        } else {
+            $self->verbose("$EXPORTS unmodified");
+            $result = 0;
+        }
     } else {
         $self->verbose("$EXPORTS unmodified");
     }


stdweird:
@drossy hmm, so you can't restart the nfs service when the exports file only has a single comment line?
that sounds like a bug to me. on a server i just tried i get, i don't see this behaviour
```
[root@test2802 rsyslog]# cp /etc/exports .
[root@test2802 rsyslog]# echo '' > /etc/exports
[root@test2802 rsyslog]# service nfs restart
Redirecting to /bin/systemctl restart  nfs.service
[root@test2802 rsyslog]# mv exports /etc/exports
mv: overwrite ‘/etc/exports’? y   
[root@test2802 rsyslog]# service nfs restart
Redirecting to /bin/systemctl restart  nfs.service
```

stdweird:
i also tried with reload, no such error was reported

stdweird:
@drossy ah, it seems you get that error when nfs service isn't started yet before a reload is tried

drossy:
Yes, I'm using ncm-nfs to manage some network mounts in /etc/fstab - this isn't an nfs server. But when the component first runs it inserts its 'File generated by ncm-nfs' comments into the previously zero-length exports file, causing a reload of the nfs service - which generates an error with RHEL7 as systemd complains nfs isn't started. Subsequent runs are okay, but some of the initial setup of this system needs the nfs mounts working, so the system reboots into a half-working state.

There might be better ways of fixing this - a smarter FileWriter than is comment-aware so doesn't report changed files when only comments have been added, or splitting this component in two ncm-networkmounts, and ncm-nfsserver?

ned21:
> Yes, I'm using ncm-nfs to manage some network mounts in /etc/fstab

Wouldn't ncm-fstab be the right tool for that? I just realised I have never tried it but it would seem sensible that it supports it?

jrha:
We've always recommended the nfs component for managing access to all network filesystems, it seems like it needs a bit more decoupling from NFS though.

drossy:
test this please

stdweird:
@jrha can i unapprove? this is a workaround around another problem
it would be far easier and better to provide a boolean in the component to skip all nfs actions, and only edit fstab

jrha:
@stdweird you can dismiss my review if you want

stdweird:
@drossy in #1115 i added a boolean to indicate a host is never a nfs server, ignoring exports and never triggering a reload
if you do not manage any nfs servers, you can add a global `components/nfs/site-config.pan` with
```pan
unique template components/nfs/site-config;
"/software/components/nfs/server" = false;
```
if you use the new config and schema templates

PR #1115, opened 2017-05-08
ncm-nfs: add server boolean and limit nfs service reload to export changes

### LINKED ISSUE #1029: ncm-nfs: why restart nfs when fstab is updated?

stdweird:
Fixes #1029 
Replaces #1110


ncm-nfs/src/main/pan/components/nfs/schema.pan
```diff
@@ -18,6 +18,10 @@ type structure_nfs_mounts = {
 
 type ${project.artifactId}_component = {
     include structure_component
+    @{Configure a NFS server. In particular relevant for missing exports attribute.
+      If true, missing exports forces an empty exports file and a NFs service reload.
```
ned21: a NFs => an NFS

jrha:
@stdweird could you update annotation as per @ned21's comment?

stdweird:
@jrha done. but the travis check is wrong. via `fstab_simple` and `base`

jrha:
It's not wrong, the template using it hasn't included it directly, consider it a warning and ignore if you want.

stdweird:
@jrha jenkins fails on another component unittest. i added the commit here anyway

PR #1118, opened 2017-05-19
ncm-metaconfig: udev : add blockdev udev rules

kwaegema:
configure udev to run scripts for certain blockdevices

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/udev/blockdev.tt
+++ b/ncm-metaconfig/src/main/metaconfig/udev/blockdev.tt
@@ -1,0 +1,8 @@
+[%- FOREACH dv IN scsi_run -%]
+SUBSYSTEM=="block", SUBSYSTEMS=="scsi", PROGRAM=="/lib/udev/scsi_id -g -u -d /dev/%k", RUN+="[% dv  %]"
+[% END -%]
+
+[%- FOREACH dv IN dm_run -%] 
+SUBSYSTEM=="block", KERNEL=="dm-*", PROGRAM=="/lib/udev/scsi_id -g -u -d /dev/%k", RUN+="[% dv  %]"
+[% END -%]
+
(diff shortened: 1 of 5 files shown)


ncm-metaconfig/src/main/metaconfig/udev/blockdev.tt
```diff
@@ -0,0 +1,8 @@
+[%- FOREACH dv IN scsi_run -%]
+SUBSYSTEM=="block", SUBSYSTEMS=="scsi", PROGRAM=="/lib/udev/scsi_id -g -u -d /dev/%k", RUN+="[% dv  %]"
```
stdweird: it would be really great if you could generate the `/sys/block/` echo commands like https://github.com/finley/GPFS-Goodies/blob/master/sbin/tune_block_device_settings#L431 (i'm assuming that's what your magic script does 😄 

also, this is only useful if after the changes https://github.com/finley/GPFS-Goodies/blob/master/sbin/tune_block_device_settings#L568 is applied. this should be mentioned somewhere

kwaegema: the magic script first loops over devices and checks if /sys/block/$dev/device/model is a MD34xx to not include eg boot disk.

stdweird:
@kwaegema can you changethe title and commit message (prefix with `ncm-metaconfig: udev:`

jrha:
I can see what this does, but could you explain /why/ it is useful?

kwaegema:
changed title and messages

jrha:
Code looks fine to me, but I'll leave final approval up to @stdweird on this one.

stdweird:
@kwaegema can you address the travis/panlint issues

kwaegema:
@stdweird fixed panlint, only nagging about no metaconfig config included

stdweird:
retest this please

PR #1119, opened 2017-05-26
ncm-systemd: cleanup, restarting modified unitfiles and initial handling of unconfigured units

### LINKED ISSUE #1095: ncm-systemd: Manual intervention required in order for drop-in unit files to have any affect
If you add a drop-in unit file using ncm-systemd, e.g.
```
include 'components/systemd/config';
'/software/components/systemd/unit/{nrpe.service}/file' = dict(
    'only', true,
    'config', dict(
        'service', dict(
            'User', 'nagios',
            'Group', 'nagios',
        ),
    ),
);
```
you need to manually restart the service in order for this to take affect. Is there any way this can be automated? (besides using ncm-filecopy with `restart` instead of ncm-systemd).

stdweird:
Support 00-tqu and CAF::Path
Fixes #1095 
Also adds initial support for handling unconfigured units (unverified and very advanced usage)


ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
```diff
@@ -462,6 +459,14 @@ sub configured_units
         $units{$detail->{name}} = $detail;
     }
 
+    if (@tryrestart) {
+        # It is ok to do the conditional restart here. This does not change state.
```
ned21: What does this comment mean? Why might it not be OK to do the conditional restart here? What is the "state" that is not changed -- akin to keeps_state=>1 or because the restart should keep the service in the same running/not-running state as before? (In which case what about the case where stop succeeds but start fails)?

stdweird: This part of the code runs before the enabling/disabling/start/stop code of the units.
Conditional restarts do not changes the states: previously running (should) still be running; stopped is still stopped. 
But indeed, modiifcations to the unitfile can cause a unit not to get started. But that can be said about anything that you can configure.

stdweird:
@ned21 @jrha: wrt the failing panlint tests, can i start using panc 10.4 and use the choice type to clean this up?

ned21:
https://github.com/quattor/pan/releases says 10.4 is still only a release candidate. "Documentation is not yet ready, RPMs are not available." We should fix that first and make 10.4 an official release.

ncm-systemd/src/main/perl/Systemd/UnitFile.pm
```diff
@@ -1,24 +1,14 @@
...
 
 use 5.10.1;
-use strict;
-use warnings;
-
-use parent qw(CAF::Object Exporter);
 
-use File::Path qw(rmtree);
-use File::Copy qw(move);
+use parent qw(CAF::Object Exporter CAF::Path);
```
ned21: Should we declare the minimum version of CAF::Path functionality required here?

stdweird: no, the used methods `directory` and `cleanup` are in original `CAF::Path`

stdweird:
@ned21 it's backwards incompatible if you used to have the chkconfig `off` setting for unconfigured services. because that will now actually turn off units, and turning off units is potentially way more intrusive. on th eother hand, you could consider this a bug fix (as previoulsy, the chkconfig off setting was not honoured).

ned21:
@stdweird Sounds like a bug fix to me. :) Do we have an issue to mark as fixed? Or do you just want to call it out in the commit log? Something like (to check my understanding of what you have said):

```
ncm-systemd: support more options for unconfigured services

Previously services not listed in the profile were not turned off if the chkconfig
component was configured with default=off. This ensures that all unknown services
are correctly defaulted to disabled.
```

stdweird:
@ned21 no. `off` means disable and stop; esp the stop part might cause serious problems with systemd (extreme example: stop the default multi-user target, which i think is equivalent to shutdown. this was never an issue in chkconfig, as the init runlevels weren't services)
there is now a 'disable' state, which is recommended and what you probably want, but not via the chkconfig configuration, only via the systemd config tree. it takes precedence over the (existing) chkconfig values.

and disable also doesn't mean not running, the service can be started via a dependency etc etc.
so i'm not sure this is simple bug fix are a new feature with severe consequences.

we don't use the off state, and i have not actually tried it, so it requires testing by sites that have it.

what i could do for now is to introduce the code and activate it only via the systemd config, and when tested/understood, enable the legacy chkconfig behaviour. it should allow jindrich to try it out 😄

ned21:
OK, I'm still confused. :( So the off state needs to be mapped to when chkconfig startstop=true?

> what i could do for now is to introduce the code and activate it only via the systemd config, and when tested/understood, enable the legacy chkconfig behaviour. it should allow jindrich to try it out

Sounds good. I'm not at work this week so will let you liaise with him directly.

stdweird:
@ned21 in chkconfig, the non-configured `off` state means disable and stop, it does not consider startstop (see https://github.com/quattor/configuration-modules-core/blob/master/ncm-chkconfig/src/main/perl/chkconfig.pm#L224). that could be considered a bug in chkconfig of some sort, but it's probably an old one.
i could map legacy `off+start_stop=false` to `disable`, but that would not match current chkconfig behaviour (and who runs chkconfig with `startstop=false` anyway...); so i'm not going to do that.

lets start with tests using the systemd config only, and see how the legacy behaviour should look like.

ned21:
OK, so my previous understanding of the behaviour was correct, but your concern is that this is potentially dangerous in systemd because it could shutdown something more important than a simple service.

stdweird:
@ned21 this one

PR #1123, opened 2017-06-06
aiiserver: options unittest: support older perl versions that do not have generic newline

stdweird:


stdweird:
@jrha an el5 issue, not sure if it is blocking for https://github.com/quattor/release/issues/297

jrha:
Only affects the unit tests on EL5? Meh.

PR #1124, opened 2017-06-14
ncm-cron: cleanup and switch to 00-tqu

stdweird:
Requires https://github.com/quattor/CAF/pull/230


ncm-cron/src/main/perl/cron.pm
```diff
@@ -1,66 +1,123 @@
...
+cause unpredictable behaviour. Possible behavours are duplicate entries or
+entries being removed altogether.
+
+Editing BETWEEN the tags will cause the edits to be overwritten the next time
+ncm-cron runs.
+
+=cut
+
+
+# TODO: switch to 17.3.1 for listdir
```
ned21: A reminder to fix this before merging

stdweird:
@ned21 i piushed the fix. now that 17.3 i out, i can make 17.3.1 rpms for myself

ncm-cron/src/main/perl/cron.pm
```diff
@@ -1,66 +1,121 @@
...
+use CAF::Object;
 
 use Encode qw(encode_utf8);
 use English;
 
-$NCM::Component::cron::NoActionSupported = 1;
+our $EC = LC::Exception::Context->new->will_store_all;
 
-use constant CRON_LOGFILE_SUBDIR => "/var/log/";
+# TODO: double check if this is true
```
ned21: What's the doubt here? Any chance of confirming this before merging?

stdweird: ah, yes. there is/was some LC code in here, i will have to doublecheck.

ncm-cron/src/main/perl/cron.pm
```diff
@@ -453,76 +510,79 @@ sub cleanCrontabs {
...
             $cronEntryNames{$fname} = ''; # Value is meaningless
         }
     }
 
-    # Solaris: Collect the crontabs in /var/spool/cron/crontabs then delete
-    # any ncm-cron entries. These are identified as lines between:
-    #     NCM-CRON START:
-    #     NCM-CRON END:
+    my $crondir = $osname eq "solaris" ? $CRONDIR_SOLARIS : $CRONDIR_LINUX;
+    my $all_files = $self->listdir($crondir,
```
stdweird: hmmm, note to self: this is ugly, probably deserves it's own option (listing a dir and returning all files sounds very common for cfgmgt toolkit)

ncm-cron/src/main/perl/cron.pm
```diff
@@ -453,76 +510,79 @@ sub cleanCrontabs {
...
+            next if $filename !~ /$cron_entry_regexp$/;
+            next if defined($cronEntryNames{$filename});
+
             # untainted to_unlink to work with tainted perl mode (-t option)
-            if ($to_unlink =~ m{^($linux_crondir/.*$cron_entry_regexp)$}) {
-                $to_unlink = $1;  # $to_unlink is now untainted
+            if ($path =~ m{^($CRONDIR_LINUX/.*$cron_entry_regexp)$}) {
+                my $to_unlink = $1;  # $to_unlink is now untainted
                 $self->info("Deleting $to_unlink");
                 LC::Check::absence($to_unlink, file => "True");
```
stdweird: note to self: replace with CAF::Path

ttyS4:
Can someone summarize what this change is about?
Maybe it is just me, but "00-tqu" does not mean anything to me :-)

stdweird:
@ttyS4 i'm doing cleanup of all components, see http://www.quattor.org/documentation/2017/02/10/components-spring17-cleaning.html 
00-tqu is a common unittest that tests module loading, documentation test, perlcrtitic, non-fatal perltidy and TT render tests (if any).
normally this is cleanup only, unless there are commits that suggest otherwise.

stdweird:
@ned21 remarks addressed

ned21:
The Jenkins machine's SSL cert expired so I can't view the results to see why it failed

ncm-cron/src/main/perl/cron.pm
```diff
@@ -453,76 +511,85 @@ sub cleanCrontabs {
...
         }
     }
 
-    # Solaris: Collect the crontabs in /var/spool/cron/crontabs then delete
-    # any ncm-cron entries. These are identified as lines between:
-    #     NCM-CRON START:
-    #     NCM-CRON END:
+    my $crondir = $osname eq "solaris" ? $CRONDIR_SOLARIS : $CRONDIR_LINUX;
+
+    # Files only. See remark about LC::Check::absence below
```
ned21: I think this is confusing. I have to search forward to find another comment (which may well get deleted in the future) to understand it. If I have understood the comment correctly then I think a better way to say it would be:
```
# Files only. Do not modify listdir() call to return of directories unless 
# CAF::Path::cleanup has also been modified to support directories.
```

ncm-cron/src/main/perl/cron.pm
```diff
@@ -453,76 +511,85 @@ sub cleanCrontabs {
...
+            next if $filename !~ /$cron_entry_regexp$/;
+            next if defined($cronEntryNames{$filename});
+
             # untainted to_unlink to work with tainted perl mode (-t option)
-            if ($to_unlink =~ m{^($linux_crondir/.*$cron_entry_regexp)$}) {
-                $to_unlink = $1;  # $to_unlink is now untainted
+            if ($path =~ m{^($CRONDIR_LINUX/.*$cron_entry_regexp)$}) {
+                my $to_unlink = $1;  # $to_unlink is now untainted
+                # This used to be done with LC::Check::absence(file => True)
+                # but this is a loop over files only, so it's safe to use CAF::Path::cleanup
```
ned21: Are the consequences of a directory ending up in here so bad that it should be explicitly checked whether it's a directory or a file before passing to cleanup?

ncm-cron/src/main/perl/cron.pm
stdweird: ok, but `cleanup` should be modified to support a mode where only files are deleted, and anything else is ignored/skipped (that's what the old `absence(Files => True)` is about

ncm-cron/src/main/perl/cron.pm
stdweird: it would be backwards incompatible, but i don't think this is bad.

stdweird:
@ned21 comment modified, i hope it is clearer now

jrha:

jrha:
@ned21 passes tests, are you happy to approve?

PR #1125, opened 2017-06-14
ncm-ntpd: cleanup and switch to 00-tqu

stdweird:
Backwards incompatible because the extensible types are not extensible anymore


ncm-ntpd/src/main/perl/ntpd.pm
```diff
@@ -85,271 +78,265 @@ sub Configure {
...
+        } sort keys %$opts;
+        my $optstring = join(" ", @o);
+        print $fh "statistics $optstring\n";
+    }
+
+    # filegen
+    if ($cfg->{filegen}) {
+        foreach my $opts (@{$cfg->{filegen}}) {
+            my @opts = grep { !/name|file/ } sort keys %{$opts};
+            my @o    = map  { "$_ $opts->{$_}" } @opts;
```
ned21: Looks like this line missed the whitespace clean up

ncm-ntpd/src/main/perl/ntpd.pm
```diff
@@ -85,271 +78,265 @@ sub Configure {
...
+{
+    my ($self, $cfg, $ntp_servers, $client_networks, $fh_opts) = @_;
+
+    my $fh = CAF::FileWriter->new($NTPDCONF, %$fh_opts);
+
+    print $fh "# This file is under $COMPNAME control.\n";
+
+    if ($cfg->{restrictdefault}) {
+        my $opts = $cfg->{restrictdefault};
+        my @o    = map {
```
ned21: Further opportunity for whitespace clean up here?

ncm-ntpd/src/main/pan/components/ntpd/schema.pan
```diff
@@ -112,43 +141,73 @@ type ntpd_statistics = {
...
-    "name" : ntpd_filegen_name
+    "name" : string with match(SELF, '^(clock|crypto|loop|peer|raw|sys)stats$')
     "file" : string
     "type" ? string with match(SELF, '^(none|pid|day|week|month|year|age)$')
     "linkornolink" ? string with match(SELF, '^(no)?link$')
     "enableordisable" ? string with match(SELF, '^(en|dis)able$')
 };
 
-type component_ntpd_type = extensible  {
+type ${project.artifactId}_component = extensible  {
```
ned21: This being extensible looks like it has no effect/value since I don't see anything in the code that suggests that arbitrary additional keys will be added to the config file?

Looking closer I see it is used fairly liberally for the types defined here - perhaps they could/should all be removed?

stdweird:
@ned21 i made a separate commit to remove the extensible type. i also flagged this PR as backwards incompatible for this reason

PR #1126, opened 2017-06-15
ncm-metaconfig: cleanup and switch to 00-tqu

stdweird:


jrha:

jrha:
I don't  think the panlint warnings were caused by this change, but it would be nice if they were fixed if you have a moment.

stdweird:
@jrha panlint is happy

PR #1129, opened 2017-06-21
ncm-mysql: set initOnce to true by default

wdpypere:
I ran into a (luckily test) server dropping and recreating the database everytime ncm-mysql ran because this is set to false.

I do believe the use case of recreating every time is less likely than setting it up once. So setting it to true is more "what somebody would expect" IMHO.

### DIFF
--- a/ncm-mysql/src/main/pan/components/mysql/schema.pan
+++ b/ncm-mysql/src/main/pan/components/mysql/schema.pan
@@ -72,7 +72,7 @@ type component_mysql_db_options = {
     'server' : string
     'users' ? component_mysql_db_user{}
     'initScript' ? component_mysql_db_script
-    'initOnce' : boolean = false
+    'initOnce' : boolean = true
     'createDb' : boolean = true
     # tableOptions is a dict of table where value is a dict of parameter/value pairs.
     # If the parameter contains spaces, it must be escaped.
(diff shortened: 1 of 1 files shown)


wdpypere:
there is one line which has a panlint error which I won't be able to fix.

ned21:
Code looks fine but can you make the commit messages more descriptive please? They should mention the component name in the title, so e.g.

```
ncm-mysql: set initOnce to true by default

In the default configuration the databases would be dropped and recreated each time the
component ran which could lead to dataloss. This use case seems less likely than the
use case where initOnce=true.
```
```
ncm-mysql: modify PAN to pass panlint
```

wdpypere:
@ned21 commit messages updated

ned21:
Thanks for the update. jenkins seems to be down so I can't see why this failed but otherwise this LGTM

ned21:
retest this please

jrha:
@wdpypere LGTM, jenkins0 isn't currently visible outside UGent, so could you have a look and see what is causing the tests to fail?

stdweird:
retest this please
@jrha test results are gone. i'll rerun them

wdpypere:
@jrha tests pass again.

jrha:
Cool. Could you squash c9f81b6e12d7c805c9b0677955367311fb50ed8a, aa81b1b93549df0de8a575ecc2af225cc80543c2 and a487daa82e0da1853c9190a561249655e4d5147a into one commit please?

wdpypere:
@jrha Done

PR #1136, opened 2017-06-27
ncm-gmond: cleanup and support new configuration options

### LINKED ISSUE #1116: ncm-gmond: configuration options missing from schema
The options max_udp_msg_len, allow_extra_data and host_tmax are missing from the schema for the global section, also bind_hostname is missing from the schema for the udp_send_channel section, when trying to convert our ganglia 3.7 configuration into aquilon.

stdweird:
Fixes #1116


stdweird:
@drossy this does not add support for `max_udp_msg_len` which i'm clueless about

ncm-gmond/src/main/pan/components/gmond/schema.pan
```diff
@@ -1,107 +1,123 @@
...
     "mute" ? boolean = false
     "deaf" ? boolean = false
     "host_dmax" ? long(0..) = 0
+    "host_tmax" ? long(0..) = 0
     "cleanup_threshold" ? long(0..) = 0
     "gexec" ? boolean = false
     "send_metadata_interval" ? long(0..) = 0
     "module_dir" ? string
+    "allow_extra_data" ? boolean
+    "max_udp_msg_len" ? long(0..)
```
jrha: Would it make sense to set an upper limit here? AFAIK 64k might be sensible.

stdweird: dunno, i guess so

PR #1142, opened 2017-07-16
ncm-altlogrotate: fix bugs introduced in refactor/cleanup PR #1045

### LINKED ISSUE #1141: ncm-altlogrotate 17.3.1: invalid "frequency" keyword added to files
A code refactoring of `ncm-altlogrotate` has been done since 17.2 that produces invalid logrotate configuration files. The problem is that it adds to the config files lines that correspond to "key value" from the schema. But in some cases, in particular `frequency`, this is unappropriate. Checking 16.12, there was lines like this to handle these special cases (not sure if this is the only one):
```
    if (defined($entry{frequency})) {
        $contents .=
            $entry{frequency}->getValue() . "\n";
    }

```

As it is currently, `ncm-altlogrotate` is broken and is probably a show stopper for 17.3.1 deployment if you rely on ncm-altlogrotate (at least frequency may not be what you want and I was not able to find what is the default frequency or if the absence of a defined frequency makes the entry ignored).

stdweird:
Fixes #1141


ncm-altlogrotate/src/main/perl/altlogrotate.pm
```diff
@@ -20,13 +20,19 @@ use Readonly;
 
 Readonly my $HEADER => "\n#\n# Generated by ncm-altlogrotate.\n#\n";
 
+# no<key>
```
jouvin: I'd say `<key> (no value)` if property is true

jouvin:
BTW, thanks for fixing this so quicly!

stdweird:
@jouvin i addressed your remarks

jouvin:
It seems that there is a  problem with Jenkins...

stdweird:
retest this please

PR #1148, opened 2017-07-25
ncm-spma: Remove support from groups from spma yumng backend

### LINKED ISSUE #952: ncm-spma (yumng): Remove support for package groups
Remove support for package groups from the newer yumng backend.

See #950 for details.

jrha:
Resolves #952.

Would be nice if @jnovy could look over this.

### DIFF
--- a/ncm-spma/src/main/pan/components/spma/yumng/config.pan
+++ b/ncm-spma/src/main/pan/components/spma/yumng/config.pan
@@ -4,7 +4,7 @@
 
 unique template components/spma/yumng/config;
 
-# Prefix for packages/groups
+# Prefix for packages
 prefix '/software';
 
 # Package to install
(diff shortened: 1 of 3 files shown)


ned21:
hmm... I have a feeling we had to abandon this change on our RH6 builds because we couldn't prove that it didn't introduce unexpected changes. So we might want to keep this around but block its use on RH7+. Will check and come back.

jnovy:
You are well ahead of time @jrha :-)

I like the idea of getting rid of groups as they are. The yumng related patch breaks stuff for us in RHEL5 and 6 where we still use groups. Since RHEL7 I made the following modification so that spma would be happy in case of either groups being present or not:

`    my $groups = defined($config->getElement(GROUPS_TREE)) ? $config->getElement(GROUPS_TREE)->getTree() : [];`

Still leaving the rest of the code there. In RHEL7 we just make `/software/groups/names = list();` i.e. empty and use pregenerated groups. We use the same code in RHEL5,6 which works with the actual groups being used.

jrha:
Ok, I'll leave things as they are and remove the milestone for now.
I get the feeling you are not using the mainline code base though?

PR #1150, opened 2017-08-01
ncm-syslog: Expand facility regex to allow comma seperated lists

jrha:
Fixes problem reported in 17.7.0-rc1 (quattor/release#300).

### DIFF
--- a/ncm-syslog/src/test/perl/configure.t
+++ b/ncm-syslog/src/test/perl/configure.t
@@ -43,6 +43,7 @@ random stuff
 *.* other action
 user.crit\tawesome
 mail.debug\tawesome
+uucp,news.crit\t/var/log/spooler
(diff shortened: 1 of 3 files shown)


stdweird:
add a unittest?

jrha:
Yes!

ncm-syslog/src/test/perl/configure.t
```diff
@@ -43,6 +43,7 @@ random stuff
 *.* other action
 user.crit\tawesome
 mail.debug\tawesome
+uucp,news.crit\t/var/log/spooler
```
stdweird: need to add this line either to orig in the next tests, or always to the expected value

jrha:
@stdweird - better?

PR #1155, opened 2017-08-10
ncm-chkconfig: add chkconfig function to enable/disable services

stdweird:


stdweird:
i'm really fed up with the ususal way of enabling services.
i was going to add the `on+startstop=true` as default via schema, but https://github.com/quattor/pan/issues/163 got in my way

ncm-chkconfig/src/main/pan/components/chkconfig/schema.pan
```diff
@@ -37,6 +33,23 @@ function chkconfig_allow_combinations = {
     true;
 };
 
+@{Enables (on + startstop&#61;true) all arguments as services.
+  If argument starts with a -, it is disabled (off + startstop&#61;true).
+  Use as e.g.
+    "/software/components" &#61; chkconfig("service1", "-service2");
```
ned21: Why is it necessary to escape the = ?

ncm-chkconfig/src/main/pan/components/chkconfig/schema.pan
```diff
@@ -37,6 +33,23 @@ function chkconfig_allow_combinations = {
...
+}
+function chkconfig = {
+    foreach (idx;arg;ARGV) {
+        m = matches(arg, '^(-)?(.*)$');
+        what = 'on';
+        if (m[1] == '-') {
+            what = 'off';
+        };
+        SELF['chkconfig']['service'][escape(m[2])] = dict(what, '', 'startstop', true);
```
ned21: Can you operate a level lower in the tree? This will cause an error if any part of /software/components is marked final since the compiler cannot tell that you are only modifying the chkconfig subtree.

stdweird: i don't understand what you mean here. but if you want
```pan
"/software/components/chkconfig" = chkconfig(...);
```
we need a new function name. i don't like redundancy in names

stdweird: i opened https://github.com/quattor/pan/issues/164

ncm-chkconfig/src/main/pan/components/chkconfig/schema.pan
ned21: 😢

ncm-chkconfig/src/main/pan/components/chkconfig/schema.pan
ned21: I doubt the PAN bug is going to get fixed in time for 17.8 so how do we proceed? I can't think of a better name for the function than chkconfig()

ned21:
As discussed at workshop, fixing https://github.com/quattor/pan/pull/166 solves this in a better way so closing won't fix.

PR #1169, opened 2017-09-06
ncm-accounts: is_user_or_group function can be used without component

### LINKED ISSUE #1166: ncm-systemd: schema depends on components/accounts/functions
The systemd schema depends on the defined_group and defined_user functions from ncm-accounts. However these functions are not valid if users and groups are defined elsewhere (e.g. LDAP service). The schema should be modified to allow a site to disable this validation if it does not use ncm-accounts.

Meta: I am sure we have discussed and potentially even implement a fix for this same problem before but cannot find it. Do we have a standardised fix that can be applied?

I see [ncm-ntpd](https://github.com/quattor/configuration-modules-core/commit/837c8f799ad8661383deecafa5971e4735801008) has the the same bug.

ned21:
The is_user_or_group() function needs to take into account that some
sites use mechanisms other than ncm-accounts to provision users and
groups. Following the pattern introduced into ncm-ccm, allow the
function to be called but always return true if the
/software/components/accounts tree is not present.

Fixes #1166

### DIFF
--- a/ncm-accounts/src/main/pan/components/accounts/functions.pan
+++ b/ncm-accounts/src/main/pan/components/accounts/functions.pan
@@ -42,6 +42,11 @@ function is_user_or_group = {
         );
     };
     pref = "/software/components/accounts";
+    # not all sites use ncm-accounts, e.g. users and groups may be provisioned
+    # via LDAP instead so return now we've confirmed it is a string
+    if (!path_exists(pref)) {
+        return(true);
+    };
 
     name_exists = true;
     while(idx < length(names)) {


ned21:
panlint failures are unrelated to my change.

jrha:
Lint issues fixed in #1170.

ncm-accounts/src/main/pan/components/accounts/functions.pan
```diff
@@ -42,6 +42,11 @@ function is_user_or_group = {
         );
     };
     pref = "/software/components/accounts";
+    # not all sites use ncm-accounts, e.g. users and groups may be provisioned
```
stdweird: if this is the case, this function also needs to be moved out this component (imho). how would this function be available otherwise.

ncm-accounts/src/main/pan/components/accounts/functions.pan
```diff
@@ -42,6 +42,11 @@ function is_user_or_group = {
...
+    # via LDAP instead so return now we've confirmed it is a string
+    if (!path_exists(pref)) {
```
stdweird: if the path exists but component is inactive, it should also return true. that way you can include the component, but put it inactive.

ncm-accounts/src/main/pan/components/accounts/functions.pan
ned21: functions.pan is standalone so I have no problem including it even without accounts/config.

ncm-accounts/src/main/pan/components/accounts/functions.pan
ned21: Strictly true but feels wrong. Is the config in the active component valid? Partial? Typically we disable components during different phases of build so the config is valid but not enforced. That would interact badly with this proposal.

stdweird: ok, after reading the explanation in  #1166 i understand it better now.

PR #1171, opened 2017-09-06
ncm-cups enhancements and cleanups

### LINKED ISSUE #1164: ncm-cups 17.7.0: configuration of ServerAlias + miscellanous enhancements
In 17.7.0, `ncm-cups` allows only to configure the `ServerName` attribute. It would be good to allow to configure the `ServerAlias` too. This configuration module also deserves some cleanups and enhancements. The main tasks required are:

- [x] ServerAlias configuration support
- [x] Removal of code to support legacy schema (not allowed by the current schema)
- [x] Promote warnings raised in case of printer configuration errors to errors: this will allow the printer configuration to be retried at the next profile deployment, in the case of a transient problem (as it happens sometimes at boot time when `cupsd` is started but no yet answering requests)
- [ ] Check existence of the printer PPD file if explicitly defined
- [ ] Additon of unit tests for the main features
- [ ] Use CAF Rule-Based Editor to replace the config parser embedded in the config module (originally derived from `ncm-dpmlfc`, the ancestor of CAF RBE)

jouvin:
Fixes #1164.


jouvin:
Code almost complete but not yet fully tested. Unit tests still missing but planned as part of this PR.

ncm-cups/src/main/pan/components/cups/config.pan
```diff
@@ -1,8 +1,20 @@
 # ${license-info}
```
stdweird: can you use templates like `${componentconfig}` (see http://www.quattor.org/documentation/2017/02/10/components-spring17-cleaning.html#pan-cleanup)

stdweird:
@jouvin would be nice if you could do the 2017 spring cleanup  (http://www.quattor.org/documentation/2017/02/10/components-spring17-cleaning.html), incl component

jouvin:
@stdweird yes, sure, it was my intent. I used `ncm-metaconfig` as a template, thinking it would certainly be a good reference and was surprised to see the legacy contents... I'll do it, thanks for raising my attention!

stdweird:
@jouvin wrt metaconfig, see #1126

jouvin:
@stdweird pan template cleanup done.

ncm-cups/src/main/pan/components/cups/schema.pan
```diff
@@ -1,8 +1,4 @@
-# ${license-info}
-# ${developer-info}
-# ${author-info}
-
-declaration template components/cups/schema;
+${componentschema}
```
stdweird: you have to remove the bind statement at the end (it's in config), and rename the main component type to `${project.artifactId}_component`
if no unittests fail without these modifications, you have to modify an existing one so the new component config is actually used (see http://www.quattor.org/documentation/2017/02/10/components-spring17-cleaning.html#add-a-unittest)

PR #1172, opened 2017-09-07
ncm-accounts: add colord to protected user/group list

jouvin:
`colord` (standard) package creates a `colord` user and group that are not protected in `ncm-accounts`. Hence, they are removed by the first run of `ncm-accounts` after the package installation.


jrha:
It might seem obvious, but please explain why in the commit message body.

Issue #1012, opened 2017-01-12
ncm-network: trigger udev before restarting network service


stdweird:
if mac-addresses are used in the network configuration, el7 udev rule `/usr/lib/udev/rules.d/60-net.rules` will assign the interface name to this devices. this way you can enforce your own/legacy naming on el7.
however when a device is added to the quattor configuration (or the mac address is changed of an existing network device), a restart of the network will not work because the rules need to be rerun.

manual fix is as simple as `udevadm trigger --attr-match=subsystem=net` before the restart.

still needs more testing

PR #1175, opened 2017-09-12
Network component updates - not for merging

gombasg:
This branch contains our changes to ncm-network, rebased on top of current master. I cleaned up the obvious conflicts during rebasing, but this is mostly untested. I'm looking into porting these changes on top of #1087, but that may take a while.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -28,11 +28,24 @@ type structure_interface_alias = {
     Describes the bonding options for configuring channel bonding on EL5 and similar.
 }
 type structure_bonding_options = {
-    "mode" : long(0..6)
+    "ad_select" ? string with match(SELF, '^(\d+|stable|bandwidth|count)$')
+    "all_slaves_active" ? long
+    "arp_interval" ? long
+    "arp_ip_target" ? type_ip[]
+    "arp_validate" ? string with match(SELF, '^(\d+|none|active|backup|all)$')
+    "downdelay" ? long
+    "fail_over_mac" ? string with match(SELF, '^(\d+|none|active|follow)$')
+    "min_links" ? long
+    "use_carrier" ? long
+    "num_grat_arp" ? long
+    "num_unsol_na" ? long
+    "resend_igmp" ? long
+    "mode" : long(0..6) # TODO: allow mode to be given as a string
     "miimon" : long
     "updelay" ? long
     "downdelay" ? long
     "primary" ? valid_interface
+    "primary_reselect" ? string with match(SELF, '^(\d+|always|better|failure)$')
     "lacp_rate" ? long(0..1)
     "xmit_hash_policy" ? string with match (SELF, '^(0|1|2|layer(2|2\+3|3\+4))$')
 } with {
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -6,12 +6,190 @@ include 'pan/types';
 include 'quattor/functions/network';
 
 @documentation{
+  IPv4/IPv6 prefix notation
+}
+function is_ip_prefix = {
+    if (ARGC != 1 || !is_string(ARGV[0])) {
+        error("Usage: is_interface_name(string)");
+    };
+    if (!is_ip(ARGV[0])) {
+        slash = index("/", ARGV[0]);
+        if (slash < 1) {
+            return(false);
+        };
+        pfx = to_long(substr(ARGV[0], slash + 1));
+        if (is_ipv4(substr(ARGV[0], 0, slash))) {
+            if (pfx < 1 || pfx > 32) {
+                return(false);
+            };
+            return(true);
+        };
+        if (is_ipv6(substr(ARGV[0], 0, slash))) {
+            if (pfx < 1 || pfx > 128) {
+                return(false);
+            };
+            return(true);
+        };
+        return(false);
+    };
+    return(true);
+};
+
+@documentation{
+  IPv4/IPv6 prefix notation
+}
+type type_ip_prefix = string with is_ip_prefix(SELF);
+
+@documentation{
     Route
 }
 type structure_route = {
     "address" ? type_ip
     "netmask" ? type_ip
     "gateway" ? type_ip
+};
+
+@documentation{
+    Single byte
+}
+# TODO: This should likely be defined elsewhere
+type type_byte = long with { SELF >= 0 && SELF <= 255; };
+
+@documentation{
+    Define the conversion from human-readable names to numeric values used by iproute and the kernel
+}
+# TODO: disallow overriding values which are hardcoded in the 'ip' tool
+type structure_rt_table = {
+    "dsfield" ? type_byte{}
+    "protos"  ? type_byte{}
+    "realms"  ? type_byte{}
+    "scopes"  ? type_byte{}
+    "tables"  ? type_byte{}
+};
+
+@documentation{
+    Type of the routing entry
+}
+type type_route_type = string with match(SELF, '^(local|broadcast|anycast|multicast|unicast|prohibit|unreachable|blackhole|throw|nop)$');
+
+@documentation{
+    Type Of Service (TOS), as defined in /etc/iproute2/rt_dsfield
+}
+type type_rt_dsfield = string with exists("/system/network/rt_tables/dsfield/" + SELF);
+
+@documentation{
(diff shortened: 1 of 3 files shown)


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -6,6 +6,41 @@ include 'pan/types';
 include 'quattor/functions/network';
 
 @documentation{
+  IPv4/IPv6 prefix notation
+}
+function is_ip_prefix = {
```
stdweird: there is a `is_ipv4_prefix_length` and similar one for ipv6 in `pan/types` in template library core. this function should be moved there

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -15,6 +50,149 @@ type structure_route = {
 };
 
 @documentation{
+    Single byte
+}
+# TODO: This should likely be defined elsewhere
+type type_byte = long with { SELF >= 0 && SELF <= 255; };
```
stdweird: and use `log(0..255)`?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -28,11 +206,24 @@ type structure_interface_alias = {
...
+    "arp_ip_target" ? type_ip[]
+    "arp_validate" ? string with match(SELF, '^(\d+|none|active|backup|all)$')
+    "downdelay" ? long
+    "fail_over_mac" ? string with match(SELF, '^(\d+|none|active|follow)$')
+    "min_links" ? long
+    "use_carrier" ? long
+    "num_grat_arp" ? long
+    "num_unsol_na" ? long
+    "resend_igmp" ? long
+    "mode" : long(0..6) # TODO: allow mode to be given as a string
```
stdweird: why?  dual typed values are not recommended. we have run into it on some other occasions, and i think they are used to migrate from legacy yes/no boolean strings to actual booleans.

ncm-network/src/main/pan/components/network/schema.pan
```diff
@@ -10,6 +10,12 @@ include 'quattor/schema';
 
 type component_network_type = {
     include structure_component
+    # Name of the auto-generated udev rule which needs to be removed to allow udev re-consider the interface names
+    "udev_rules_file" ? string
+    # E.g. "/usr/bin/udevadm trigger", although restricting the scope to just network devices should not hurt
+    "udev_trigger_command" ? string
```
stdweird: what possible values could there be?

ncm-network/src/main/pan/components/network/schema.pan
```diff
@@ -10,6 +10,12 @@ include 'quattor/schema';
 
 type component_network_type = {
     include structure_component
+    # Name of the auto-generated udev rule which needs to be removed to allow udev re-consider the interface names
+    "udev_rules_file" ? string
```
stdweird: move all the `udev_` attributes to a new `udev` related type

ncm-network/src/main/perl/network.pm
```diff
@@ -60,6 +60,52 @@ my %ethtool_option_map = (
...
+    },
+    "tables" => {
+        0 => "unspec",
+        255 => "local",
+        254 => "main",
+        253 => "default",
+    },
+);
+
+my %mask_to_prefix = (
```
stdweird: use `NetAddr::Ip` like https://github.com/quattor/configuration-modules-core/pull/1087/files#diff-90bc6b2e2a105ae5c62daf55ef95b2a7R1014

ncm-network/src/main/perl/network.pm
```diff
@@ -570,6 +623,56 @@ sub Configure
...
+        } elsif (exists($net{$iface}{policy_route})) {
+            $file_name = "$dir_pref/route-$iface";
+            $exifiles{$file_name} = 1;
+            $text="";
+
+            # We can't mix different syntaxes in the same file, so convert
+            # simple routes to the policy routing syntax
+            if (exists($net{$iface}{route})) {
+                foreach my $rt (sort keys %{$net{$iface}{route}}) {
+                    my $route = {"dev" => $iface};
```
stdweird: this is not correct, should be devicename like https://github.com/quattor/configuration-modules-core/pull/1087/files#diff-90bc6b2e2a105ae5c62daf55ef95b2a7R1573

but new code makes this much easier; and also support arbitrary commands via `cmd` attribute

ncm-network/src/main/perl/network.pm
```diff
@@ -809,20 +933,49 @@ sub Configure
...
                 $self->error("replace modified/new files: can't copy ",
                              $self->gen_backup_filename($file).FAILED_SUFFIX,
                              " to $file. ($!)");
+            $was_change = 1;
         } elsif ($exifiles{$file} == -1) {
             unlink($file) || $self->error("replace modified/new files: can't unlink $file. ($!)");
+            $was_change = 1;
+        }
+    }
+    ## kick udev to rename devices if needed; do nothing if no files were changed
```
stdweird: see also #1012 
but we had some udev related issues with storage devices too, so maybe a CAF::udev might make sense

ncm-network/src/main/perl/network.pm
```diff
@@ -809,20 +933,49 @@ sub Configure
...
+    if ($was_change) {
+        # Remove the auto-generated rule file if present
+        if ($config->elementExists("/software/components/network/udev_rules_file")) {
```
stdweird: metaconfig has udev related TT files, maybe we can generate the udev files with metaconfig, and call the trigger/settle here?

ncm-network/src/main/perl/network.pm
stdweird: see #1118 for storage udev PR

ncm-network/src/main/perl/network.pm
```diff
@@ -1145,6 +1312,9 @@ sub Configure
         my @op;
 
         while (my ($k, $v) = each(%$opts)) {
+            if ($v->isType(LIST)) {
```
stdweird: i'm surprised this works. i'm going to assume you use a recent CCM?

stdweird:
@gombasg this doesn't look too hard, iwas worried for something far worse 😄 
i guess this PR can be closed and is only here as a reference for future PRs?

ncm-network/src/main/pan/components/network/schema.pan
gombasg: Nowadays I think only what's written in the comments, but I'd need to check. RH4 had an old udev with a different interface, so the value had to be configurable. But the new udevadm CLI is stable for a while, so I'd not mind removing these commands from the templates and just hard-code them in the component.

ncm-network/src/main/perl/network.pm
gombasg: Yes, that should likely happen here. We didn't want to introduce new dependencies when this code was originally written.

ncm-network/src/main/perl/network.pm
gombasg: I don't want to _create_ any udev files - but I need to _remove_ the one udev generates automatically. If the auto-generated file exists, then udev will not look into the ifcfg-* files, and it will not notice that some interfaces which did not have a configuration before may need to be renamed.

ncm-network/src/main/perl/network.pm
gombasg: It may not work :-) The original code which we still use is much uglier, and I just found this construct in one of the other components and decided it's not _that_ horrible. I think you already have a much more generic solution in #1087, but I did not have time to try that so far.

ncm-network/src/main/perl/network.pm
stdweird: what `autogenerates` it? anaconda? i always tought that ifcfg file preceded udev rules if HWADDR was set.
the location of this autogenerated udev file is OS release specific i guess? it would be nice to add the value somewhere (for aii we have os flavours); or do you think we can hardcode the location?

ncm-network/src/main/perl/network.pm
stdweird: the `->getTree` call a few lines higher returns arrayrefs for lists, that's why i was asking.

gombasg:
@stdweird: I just wanted to push this out to show what we'd like to be included, to make sure the changes you're working on are not outright incompatible. We should have published these changes a long time ago...

ned21:
@stdweird confirmed at RAL workshop that none of these items are in the latest ncm-network rewrite so they will all need to be reimplemented.

PR #1176, opened 2017-09-13
ncm-metaconfig: ganesha: make exports usable for ces

kwaegema:
gpfs ces need CLIENT and FSAL blocks last in the export config


kwaegema:
commit msg changed

PR #1179, opened 2017-09-18
ncm-gpfs: fix keydata files

kwaegema:
genkeyData1 had suffix 1 because it was first generation of key. GPFS increases this when a new key is generated, and uses this when it is committed. This fixes it to use the right key and the right location


kwaegema:
changed commit msg

PR #1183, opened 2017-09-25
ncm-pam: new function pam_add_access_group

ned21:
New function which mirrors the ability to add netgroups to the components/pam/access with one that can handle unix groups.

Also some small clean ups to remove un-necessary returns().

### DIFF
--- a/ncm-pam/src/main/pan/components/pam/config.pan
+++ b/ncm-pam/src/main/pan/components/pam/config.pan
@@ -67,9 +67,12 @@ function pam_add = {
             options = ARGV[4];
         };
     };
-    ret[service][pamtype][tail] = dict("control", control, "module", module, "options", options, "options_list", options_list);
+    ret[service][pamtype][tail] = dict("control", control,
+                                    "module", module,
+                                    "options", options,
+                                    "options_list", options_list);
 
-    return (ret);
+    ret;
 };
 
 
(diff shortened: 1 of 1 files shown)


ned21:
I think those lint errors are fixable (also just spotted a typo)

ned21:
Sadly the panlint errors are not trivially fixable without code changes that would require more significant testing.

jrha:
@stdweird?

stdweird:
@ned21 current panlint errors are very easy to fix. it's misisng 1 include to fix 11 errors and 1 whitespace issue.

ned21:
@stdweird - I tried adding the standard boilerplate header and it only reduced it to [10](https://travis-ci.org/quattor/configuration-modules-core/builds/279611865) errors which didn't seem worth the improvement. Looking again it might have been that the patch I tried introduced new errors which I mis-read as "not having fixed the old ones" but might have just been the way emacs formatted my continuation lines.

I'll have another look to see if we can clean this up in a separate PR.

ned21:
@stdweird OK now I remember why I thought this was as good as we could get. The include issues can't be fixed because it's complaining that this file should include itself. The first 5 could be fixed with a prefix command but the if() statements won't be. I think the only way to fix this is to move the functions to a new functions file and hope panlint doesn't expect to see an include components/config in there too. 

The whitespace error I disagree with: 
```pam_add_access_acl(key, "+", "("+group+")", "ALL");```

is more readable than:

```pam_add_access_acl(key, "+", "(" + group + ")", "ALL");```

PR #1187, opened 2017-10-04
ncm-download: kinit: add a possibility to pass arguments

natari:
kinit -k, called by get_gss_token, sometimes may need exstra arguments
passed in. Introducing kinit_args in the $tree makes it possible.
The change does not introduce any backwards incompatibilities - if kinit_args is not set, the default "kinit -k" will be called.

### DIFF
--- a/ncm-download/src/main/pan/components/download/schema.pan
+++ b/ncm-download/src/main/pan/components/download/schema.pan
@@ -63,4 +63,6 @@ type download_component = extensible {
     "head_timeout" ? long
     @{seconds, total timeout for fetch of file, can be overridden per file}
     "timeout" ? long
+    @{argumensts to be passed in kinit -k called in ncm-download}
+    "kinit_args" ? string[]
 };
--- a/ncm-download/src/main/perl/download.pm
+++ b/ncm-download/src/main/perl/download.pm
@@ -64,7 +64,7 @@ my $_cached_gss;
 
 sub get_gss_token
 {
-    my $self = shift;
+    my ($self, $tree) = @_;
 
     # Return module wide cache value
     return $_cached_gss if $_cached_gss;
@@ -80,7 +80,9 @@ sub get_gss_token
 
     # Assume "kinit" is in the PATH.
     my $errs = "";
-    my $proc = CAF::Process->new(["kinit", "-k"],
+    my @kinit_cmd = ("kinit", "-k");
+    push(@kinit_cmd, @{$tree->{kinit_args}}) if ($tree->{kinit_args});
+    my $proc = CAF::Process->new(\@kinit_cmd,
                                  stderr => \$errs,
                                  log => $self,
                                  keeps_state => 1);
@@ -122,7 +124,7 @@ sub Configure
         $file->{head_timeout} = $tree->{head_timeout};
 
         if ($file->{gssapi}) {
-            $file->{gss_ccache} = $self->get_gss_token();
+            $file->{gss_ccache} = $self->get_gss_token($tree);
             # immediate failure
             return 0 if ! $file->{gss_ccache};
         }


ncm-download/src/main/perl/download.pm
```diff
@@ -80,7 +80,9 @@ sub get_gss_token
 
     # Assume "kinit" is in the PATH.
     my $errs = "";
-    my $proc = CAF::Process->new(["kinit", "-k"],
+    my @kinit_cmd = ("kinit", "-k");
+    push(@kinit_cmd, split(/ /, $tree->{kinit_args})) if ($tree->{kinit_args});
```
stdweird: the `kinit_args` should be a list of strings, and have to be added to the schema.
@natari do you expect different args per file or not? if you do, we cannot use the cache

ned21:
@hpcugentbot Test this please

ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -63,4 +63,5 @@ type download_component = extensible {
     "head_timeout" ? long
     @{seconds, total timeout for fetch of file, can be overridden per file}
     "timeout" ? long
+    "kinit_args" ? string[]
```
ned21: Do we want to have a validation check here to prohibit shell metacharacters?

natari:
Added kinit_args[] to the schema, added testing it against a regex to avoid metacharacters.
Thank you for your comments @ned21  @stdweird

ned21:
@hpcugentbot Test this please

@jrha I think we missed the cut off for the 17.8 release but if we can get a Code Review approval as a sign that this will be merged for 17.10, we'll feel happy deploying a pre-release build internally. Thanks!

ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -63,4 +63,5 @@ type download_component = extensible {
     "head_timeout" ? long
     @{seconds, total timeout for fetch of file, can be overridden per file}
     "timeout" ? long
+    "kinit_args" ? string[] with match(SELF, "[\w-.@$/]")
```
stdweird: what is the macth supposed to achieve? this only requires to have one vaild character.
if it were only these chars allowed, filenames with whitespace in the path won't be allowed etc etc

ncm-download/src/main/perl/download.pm
```diff
@@ -65,7 +65,7 @@ my $_cached_gss;
 sub get_gss_token
 {
     my $self = shift;
-
+    my $tree = shift;
```
stdweird: use `my ($self, $tree) = @_;`

ncm-download/src/main/pan/components/download/schema.pan
stdweird: also, can you add short description (in annotation form)

natari:
@stdweird @ned21 I added the reqested modifications, can you take a look please? Thanks

ned21:
@hpcugentbot Test this please

ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -43,6 +43,9 @@ type component_download_file = {
     "allow_older" ? boolean
 };
 
+@{safety check against shell metacharacters, for kinit_args}
```
stdweird: shell metacharcters? the kinit commands are not executed in a subshell, so you shouldn't worry about these

ned21: @stdweird - ah, so you are saying it's safe to be just a string type as @natari originally had it?

stdweird: @ned21 i think so. but @natari originally had a string with some odd regexp i tought were releated to eg kinit commands or principals, not to avoid shell escaping

```perl
#!/usr/bin/perl

BEGIN {push(@INC, '/usr/lib/perl')}

use CAF::Process;
print CAF::Process->new(['echo', '$PATH'])->output();
print CAF::Process->new(['echo', 'abc;echo hellobash'])->output();
```
gives
```
$PATH
abc;echo hellobash
```
so no shell involved (unless this is not what you mean)

natari: @stdweird yeah my original regexp was to avoid shell metacharacters. It does look like CAF::Process doesn't need it, so I will remove the regexp and leave just a string[] for the args. Thanks!

ned21:
@hpcugentbot Test this please

ned21:
@hpcugentbot Test this please

PR #1192, opened 2017-11-14
ncm-openstack: Openstack component

alvarosimon:
At least the `OpenRC`   and Keystone are configured and it includes the methods to configure the rest of OpenStack services.

Missing things:
 * Some remarks from @stdweird are not in place yet like the schema change
 * Keystone API to add OS service users and endpoints


ncm-openstack/src/main/perl/OpenStack/Glance.pm
```diff
@@ -0,0 +1,50 @@
+#${PMpre} NCM::Component::OpenStack::Glance${PMpost}
+
+use parent qw(NCM::Component::OpenStack::Service);
+
+use Readonly;
+
+Readonly our $GLANCE_API_CONF_FILE => "/etc/glance/glance-api.conf";
+Readonly our $GLANCE_REGISTRY_CONF_FILE => "/etc/glance/glance-registry.conf";
```
stdweird: not used?

ncm-openstack/src/main/perl/OpenStack/Glance.pm
```diff
@@ -0,0 +1,50 @@
...
+    $self->{filename} = $GLANCE_API_CONF_FILE;
+    $self->{daemons} = [
+        'openstack-glance-api',
+        'openstack-glance-registry',
+    ];
+}
+
+=item post_populate_service_database
+
+Glance post db_sync execution
```
stdweird: mention that it doesn't do anything

ncm-openstack/src/main/perl/OpenStack/Glance.pm
```diff
@@ -0,0 +1,50 @@
...
+
+=cut
+
+sub post_populate_service_database
+{
+    my ($self) = @_;
```
stdweird: not needed

ncm-openstack/src/main/perl/OpenStack/Horizon.pm
```diff
@@ -0,0 +1,50 @@
...
+
+sub _attrs
+{
+    my $self = shift;
+
+    $self->{tt} = 'horizon';
+    $self->{filename} = $HORIZON_CONF_FILE;
+    $self->{daemons} = ['httpd'];
+    $self->{user} = 'apache';
+    # Horizon has no database
```
stdweird: no need to change that here

ncm-openstack/src/main/perl/OpenStack/Nova.pm
```diff
@@ -0,0 +1,54 @@
...
+    $self->{db_sync} = "db sync";
+}
+
+=item post_populate_service_database
+
+Nova post db_sync execution
+
+=cut
+
+sub post_populate_service_database
```
stdweird: you mean `populate_service_database`? instead of `post_`?

ncm-openstack/src/main/perl/OpenStack/Neutron.pm
```diff
@@ -0,0 +1,62 @@
...
+    $self->{db_sync} = $NEUTRON_DB_BOOTSTRAP;
+}
+
+=item post_populate_service_database
+
+Neutron post db_sync execution
+
+=cut
+
+sub post_populate_service_database
```
stdweird: the post is already `return 1`, no need to setthis here, unless you meant `populate_service_databas`

alvarosimon:
@stdweird thanks a lot for the help and sorry for the late reply. I was fixing some stupid typos and removing the methods that are not doing anything (the code still has some copy/paste code that it should be changed). We need to include an extra method `pre_populate_service_database`   that should be used by some services before the database(s) bootstrap. But first of all I want to try the current component in grimer to see if something else is missing..

alvarosimon:
btw tests are failing because this component requires `CAF::Process version v17.8.1`

ncm-openstack/src/main/pan/components/openstack/common.pan
```diff
@@ -156,3 +156,17 @@ type openstack_DEFAULTS = {
...
+    Type to enable RabbitMQ and the message system for OpenStack.
+}
+type openstack_rabbitmq_config = {
+    @{RabbitMQ user to get access to the queue}
+    'user' : string = 'openstack'
+    'password' : string
+    @{Set config/write/read permissions for RabbitMQ service.
+    A regular expression matching resource names for
+    which the user is granted configure permissions}
+    'permissions' : string[] = list('".*"', '".*"', '".*"')
```
stdweird: force size 3; also no duoble quotes

ncm-openstack/src/main/pan/components/openstack/glance.pan
```diff
@@ -39,16 +39,24 @@ type openstack_glance_store = {
...
+    'glance_store' ? openstack_glance_store
 };
+
+@documentation {
+    list of Glance service configuration sections
+}
+type openstack_glance_config = {
+    'service' ? openstack_glance_service_config
+    'registry' ? openstack_glance_service_config
+};
```
stdweird: add newline

ncm-openstack/src/main/perl/OpenStack/Glance.pm
```diff
@@ -0,0 +1,52 @@
...
+
+Write the required config files for Glance
+
+=cut
+
+sub write_config_file
+{
+    my ($self) = @_;
+
+    my $nelement = $self->{element};
```
stdweird: remove if not used

ncm-openstack/src/main/perl/OpenStack/Rabbitmq.pm
```diff
@@ -0,0 +1,63 @@
+#${PMpre} NCM::Component::OpenStack::Rabbitmq${PMpost}
+
+use parent qw(NCM::Component::OpenStack::Service);
+
+use Readonly;
+use Data::Dumper;
```
stdweird: remove if not used

alvarosimon:
@stdweird remarks addressed but I get the error from travis `ncm-openstack/src/test/resources/os_resources.pan:95: Component openstack in use, but component config has not been included`

The component  is working and running  in our testbed

alvarosimon:
@stdweird @jrha the ceph support is already in place (at least for nova and glance), I have tested this in our cluster. Maybe this PR should be re-reviewed at this point, it is getting bigger and bigger. 
Some services are still missing (manila, cinder...) but maybe we should open a new PR for these missing services...

ncm-openstack/src/main/pan/components/openstack/common.pan
```diff
@@ -152,4 +151,22 @@ type openstack_DEFAULTS = {
...
+    Type to enable RabbitMQ and the message system for OpenStack.
+}
+type openstack_rabbitmq_config = {
+    @{RabbitMQ user to get access to the queue}
+    'user' : string = 'openstack'
+    'password' : string
+    @{Set config/write/read permissions for RabbitMQ service.
+    A regular expression matching resource names for
+    which the user is granted configure permissions}
+    'permissions' : string[3] = list('.*', '.*', '.*')
```
jrha: This would be much more user friendly as a type, rather than a strange tuple.
Perhaps something like this?
```pan
type openstack_rabbitmq_permissions = {
    'conf' : string = '.*'
    'write' : string = '.*'
    'read' : string = '.*'
}
```

jrha: Also... why should ncm-openstack be configuring RabbitMQ, rather than a ncm-rabbitmq or metaconfig service?

stdweird: it shouldn't, but the config files are horrible, and you need a component to execute some random commands. so this is way easier, but feel free to open a PR :grin:

jrha: Heh. :see_no_evil:

jrha:
I'm going to spend some more time looking at this and would like to merge it after releasing 18.3, any objections?

ncm-openstack/src/main/pan/components/openstack/horizon.pan
```diff
@@ -11,9 +11,9 @@ declaration template components/openstack/horizon;
 type openstack_horizon_caches = {
      @{We recommend you use memcached for development; otherwise after every reload
      of the django development server, you will have to login again}
-    'backend' : string = 'django.core.cache.backends.memcached.MemcachedCache'
+    'BACKEND' : string = 'django.core.cache.backends.memcached.MemcachedCache'
     @{location format <fqdn>:<port>}
-    'location' : type_hostport
+    'LOCATION' : type_hostport = '127.0.0.1:11211'
```
apdibbo: The value here can actually be a list of hostports for HA

alvarosimon: @apdibbo this was already fixed by @stdweird in #1261

PR #1195, opened 2017-11-27
ncm-metaconfig: elasticsearch: Refactor schema to allow testing

jrha:
The multiple schema versions contain duplicate types which break tests when they are copied to the template library, so refactor the schema to avoid this.

* Rename duplicate type definitions.
* Include common schema from version specific schema (rather than the other way round).


jrha:
Will look at unit tests tomorrow.

jrha:
Superseded by quattor/template-library-core/pull/170.

PR #1196, opened 2017-11-27
ncm-metaconfig: logstash: Refactor schema to allow testing

jrha:
The multiple schema versions contain duplicate types which break tests when they are copied to the template library, so refactor the schema to avoid this.

* Rename duplicate type definitions.
* Include and bind version specific schema from version specific configs.
* Replace `logstash/pan/schema` with an empty common schema. The original was non-functional, it tried to include a template which did not exist.

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/logstash/pan/config_1.2.pan
+++ b/ncm-metaconfig/src/main/metaconfig/logstash/pan/config_1.2.pan
@@ -2,7 +2,7 @@ unique template metaconfig/logstash/config_1.2;
 
 include 'metaconfig/logstash/schema_1.2';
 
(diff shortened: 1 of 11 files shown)


jrha:
Will look at unit tests tomorrow.

stdweird:
hmmm, i understand your point, but the schemas are not supposed to be mixed. if this is the way to handle multiversioned anything (ie differrent versions still require unique types), then we might have to refactor again somehow. btw, there's a PR coming up for logstash 6 that will complicate things further.

jrha:
There is another alternative — rework the tests on template-library-core to understand multi-versioned schema.

Does it really matter if the types have horrible names?

stdweird:
@jrha see also #1204

jrha:
Superseded by quattor/template-library-core/pull/170.

PR #1197, opened 2017-11-28
ncm-icinga: file permissions changed to 0440 for config files instead…

wdpypere:
… of 0444.

Monitoring configuration tells you a lot about an environment and can contain sensitive information, therefore should not be world readable.


jrha:
Well that broke it...

wdpypere:
because I now close the filehandles in the function, the test fail. I'll see if I can fix it.

wdpypere:
@jrha tests are fixed.

jrha:
:green_heart:

PR #1117, opened 2017-05-18
ncm-authconfig : Change auth_provider, reconnection_retries, user object_class and ldap_uri

seraf1:
- Make most inputs optional
- Add sssd_auth_provider_string that accepts krb5 but not simple
- Add "reconnection_retries" in sssd_pam and sssd_nss
- Make most inputs optional n sssd_global, sssd_pam and sssd_nss
- Add  "reconnection_retries" in sssd_domain
- Make some inputs optional
- Remove use_object_class in sssd_ldap because this option is not available in the SSSD configuration but there is a user_object_class option.
- Add object_class and name in sssd/user
- Make most inputs optional
- Make ldap_uri optional

PR #1199, opened 2017-11-28
Change ldap_use_object_class

seraf1:
Change ldap_use_object_class to ldap_user_object_class


jrha:
Please make this change in #1117 instead.

PR #1201, opened 2017-12-01
ncm-puppet: puppet v4 and v5 compatibility

sartiran:
the previous version of the ncm-puppet was not compliant with puppet v4 and v5. In particular, as things in such version have been heavily reorganized (executable, configuration files, modules, manifests locations have changed, for example) it was necessary to update the schema and the component introducing more flexibility.
 
There is no real backward incompatibility. We added new fields in the schema but the default values of such fields are those used for puppet v3 (so, those that were hardcoded in the old version of the component).

### DIFF
--- a/ncm-puppet/src/main/perl/puppet.pm
+++ b/ncm-puppet/src/main/perl/puppet.pm
@@ -11,20 +11,16 @@ use CAF::FileWriter;
 use Config::Tiny;
 use YAML::XS;
 
+#Needed to prevent YAML::XS to quote integers when dyumping JSON.
+$YAML::XS::QuoteNumericStrings = 0;
+
 our $NoActionSupported = 1;
 
 use Readonly;
 
-
-Readonly::Array my @APPLY => qw(puppet apply --detailed-exitcodes -v -l);
-Readonly::Array my @MODULE_UPGRADE => qw(puppet module upgrade);
-Readonly::Array my @MODULE_INSTALL => qw(puppet module install);
-Readonly::Scalar my $NODEFILES_PATH => '/etc/puppet/manifests';
-Readonly::Scalar my $PUPPET_CONFIG_FILE => '/etc/puppet/puppet.conf';
-Readonly::Scalar my $HIERA_CONFIG_FILE => '/etc/puppet/hiera.yaml';
-Readonly::Scalar my $HIERA_DATA_FILE => '/etc/puppet/hieradata/quattor.yaml';
-Readonly::Scalar my $PUPPET_LOGS => '/var/log/puppet/log';
-
+Readonly::Array my @APPLY => qw(apply --detailed-exitcodes -v -l);
+Readonly::Array my @MODULE_UPGRADE => qw(module upgrade);
+Readonly::Array my @MODULE_INSTALL => qw(module install);
 
 sub Configure
 {
@@ -32,17 +28,17 @@ sub Configure
 
     my $confighash = $config->getElement($self->prefix)->getTree();
 
-    $self->tiny($confighash->{puppetconf},$PUPPET_CONFIG_FILE);
+    $self->tiny($confighash->{puppetconf}, $confighash->{puppetconf_file});
 
-    $self->yaml($confighash->{hieraconf},$HIERA_CONFIG_FILE);
+    $self->yaml($confighash->{hieraconf}, $confighash->{hieraconf_file});
 
-    $self->install_modules($confighash->{modules}) if(defined($confighash->{modules}));
+    $self->install_modules($confighash->{modules}, $confighash->{puppet_cmd}, $confighash->{modulepath}) if(defined($confighash->{modules}));
 
-    $self->yaml($confighash->{hieradata},$HIERA_DATA_FILE) if(defined($confighash->{hieradata}));
+    $self->yaml($confighash->{hieradata}, $confighash->{hieradata_file}) if(defined($confighash->{hieradata}));
 
-    $self->nodefiles($confighash->{nodefiles});
+    $self->nodefiles($confighash->{nodefiles}, $confighash->{nodefiles_path});
 
-    $self->apply($confighash->{nodefiles});
+    $self->apply($confighash->{nodefiles}, $confighash->{nodefiles_path}, $confighash->{puppet_cmd}, $confighash->{modulepath}, $confighash->{logfile});
 
     return 0;
 }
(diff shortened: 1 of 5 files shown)


ncm-puppet/src/main/perl/puppet.pm
```diff
@@ -1,29 +1,48 @@
-#${PMcomponent}
```
stdweird: please restore this

ncm-puppet/src/main/perl/puppet.pm
```diff
@@ -1,29 +1,48 @@
...
+
+use EDG::WP4::CCM::Path qw(escape unescape);
 
 use CAF::Process;
 use CAF::FileWriter;
 
 use Config::Tiny;
 use YAML::XS;
 
+$YAML::XS::QuoteNumericStrings = 0;
```
stdweird: why is this needed?

ncm-puppet/src/main/perl/puppet.pm
```diff
@@ -32,17 +51,17 @@ sub Configure
 
     my $confighash = $config->getElement($self->prefix)->getTree();
 
-    $self->tiny($confighash->{puppetconf},$PUPPET_CONFIG_FILE);
+    $self->tiny($confighash->{puppetconf},$confighash->{puppetconf_file});
```
stdweird: space after `,`

ncm-puppet/src/main/perl/puppet.pm
```diff
@@ -53,11 +72,11 @@ sub Configure
 #
 sub nodefiles
 {
-    my ($self, $cfg) = @_;
+    my ($self, $cfg, $path) = @_;
 
     foreach my $file (sort keys %{$cfg}){
         if($cfg->{$file}->{contents}){
-            my $path=$NODEFILES_PATH."/".unescape($file);
+            my $path=$path."/".unescape($file);
```
stdweird: space around `=`

stdweird:
@sartiran if you can, try to add spaces around `=` and `,` (but it's not mandatory.
it would be nice if you could squash the commits in one

ncm-puppet/src/main/perl/puppet.pm
sartiran: Error in C/P, I'll fix

ncm-puppet/src/main/perl/puppet.pm
sartiran: hiera 5 requires a "version: 5" in the hiera.yaml file and with the default value of $YAML::XS::QuoteNumericStrings this would be rendered as "version: '5'" which causes a parsing error while running puppet apply. I've not find a way to toggle quoting/not-quoting item per item..

stdweird: does the version come from https://github.com/quattor/configuration-modules-core/pull/1201/files#diff-f4e590f8087d8658009187202140cc6aL11? a string is a string. perl does duck typing, getTree will do the right thing, but it's not garanteed within perl (eg as soon as you print a varaible holding an interger, the variabel becomes a string).
but you can get it right, or use TextRender. 

you don't have to chnage the schema, but at least add a comment here why this is required. also a unittest checing this would be good (i briefly looked at the unittests, i couldn't find one)

stdweird: ok, i had a quick look at the rest of the code. what you do with `unescapekeys` is not in textrender, so you can't use that. but changing the schema will probably do the right thing. (anyway, just my 2cents 😄

sartiran: No no, version is in hieraconf field which is a generic dict (type puppet_hieraconf) as it varies too much from version to version for having a fixed schema (as it was in the old version of the component). The "version" parameter seems to be correctly passed as a integer.  In the json I see
        "hieraconf": {
          "defaults": {
            "data_hash": "yaml_data",
            "datadir": "/etc/puppetlabs/code/environments/production/data"
          },
          "hierarchy": [
            {
              "name": "quattor",
              "path": "quattor.yaml"
            }
          ],
          "version": 5  <--- here
        },

But YAML:XS quoted such integers. I've not found better solutions that disabling integer quotes completely.

stdweird: ok, so the profile has correct type, but perl mangles it. my bet would be that this happens in unescape keys (it probably assigns the integer to a varaible that used to hold a string). i opened https://github.com/quattor/CCM/issues/178 so maybe in future you can use textrender to do this correctly.

sartiran: OK. Thanks. I'll add a comment and a unittest

stdweird:
@sartiran thx. only thing left to do is to squash the commits and use a commit message that starts with the componet ie `ncm-puppet: ...`

sartiran:
Hi, made the squash (hopefully without making big mistakes...).

stdweird:
@sartiran thx

jrha:
@hpcugentbot please test this

stdweird:
retest this please

jrha:
Tests are failing on ncm-gpfs for some reason.

sartiran:
Hi James, sorry, I'm a bit lost. How can this be caused by this PR?

jrha:
It can't... @stdweird might have some insight.

stdweird:
@jrha i'm afraid that this is related to issue in https://github.com/quattor/maven-tools/pull/167. we will have to release new maven-tools and bump all pom.xml.i will test this locally first. expect final fixes by monday.

jrha:
@sartiran if you can rebase on master the tests should pass

sartiran:
Hi James, I'm sorry, I admit I'm not sure I've understood what I'm expected to do.

jrha:
Ok no problem, I've done it for you.
What I did:
```
git clone git@github.com:sartiran/configuration-modules-core.git configuration-modules-core-sartiran
cd configuration-modules-core-sartiran/
git checkout ncm-puppet-new
git remote add upstream git@github.com:quattor/configuration-modules-core.git
git pull -r upstream master
git push -f
```

sartiran:
Hi James, thanks! OK, now I see. 
One last question. The schema template in 'components/puppet/' will it be automatically updated to this new version or should I create a PR also for 'template-library-core'?

jrha:
You don't need to create a PR, it will be automatically updated when the first RC is built.

sartiran:
ok. Thanks for all the help!

jrha:
No problem!

jrha:
@hpcugentbot test this please

PR #1203, opened 2017-12-04
ncm-ceph: add mds options for log trimming

kwaegema:
http://docs.ceph.com/docs/master/cephfs/mds-config-ref/
mds log max segments
mds log max expiring


ncm-ceph/src/main/pan/components/ceph/schema-mds.pan
```diff
@@ -7,6 +7,9 @@ type ceph_mds_config = {
     'mds_max_purge_files' ? long = 64
     'mds_max_purge_ops' ? long = 8192
     'mds_max_purge_ops_per_pg' ? double = 0.5
+    'mds_log_max_expiring' ? long = 20
+    'mds_log_max_segments' ? long = 20
+
```
stdweird: remove newline

PR #1204, opened 2017-12-05
ncm-metaconfig: service logstash: remove include of non-existing version template

stdweird:
Required to make the template-library-core unittests pass


stdweird:
@jrha can you take care of pushing this in template-library-core master?

jrha:
Sure.

PR #1217, opened 2017-12-19
ncm-openstack: Schema fixes

jrha:
The Glance, Neutron and Nova schema use the `openstack_keystone_authtoken` type
which is defined in the Keystone schema.

The missing dependency is causing tests on **template-library-core** to fail.

The root type is defined as the reverse of what the built-tools templating generates, so fix that too.

Hopefully shouldn't cause conflicts with #1192.


alvarosimon:
@jrha thanks! I will rebase #1192 with master

PR #1218, opened 2017-12-19
ncm-metaconfig: rsyslog: Move common parts of schema to a common template

jrha:
Cleaner and allows for simpler testing when copied to core template library.


jrha:
@stdweird any comments on wisdom/naming?

stdweird:
@jrha not sure what you are asking...

jrha:
Just whether you thought "common" is the right name.

stdweird:
@jrha sure, why not?

PR #1220, opened 2017-12-19
ncm-cups: Stop daemon if server not enabled

### LINKED ISSUE #1189: ncm-cups: Nodetype of client should stop running cups daemon
_From @quattor-releaser on October 1, 2009 13:59_

Dears,

When using the ncm-cups component to a quattor template, if we define `"/software/components/cups/nodetype" = "client"`
then ncm-cups component does not stop running cups daemon. It should be the case, isn't it ?
Here's a fix:

``` diff
419a420,423
>         else {
>             # Be sure that the server is stopped if it is not enabled
>             $self->serviceControl($services{server},"stop")
>         }
```

the patch is based on:
http://quattor.svn.sourceforge.net/viewvc/quattor/trunk/ncm-components/core/ncm-cups/cups.pm.cin?revision=540

Reported by: Pansanel

_Copied from original issue: quattor/release#197_

jrha:
I believe this finally resolves #1189 from @Pansanel.

### DIFF
--- a/ncm-cups/src/main/perl/cups.pm
+++ b/ncm-cups/src/main/perl/cups.pm
@@ -367,6 +367,9 @@ sub Configure
         } else {
             $self->error("Default printer $default_printer doesn't exist. Ignoring");
         }
+    } else {
+        # Be sure that the server is stopped if it is not enabled
+        $self->serviceControl($services{server},"stop")
     }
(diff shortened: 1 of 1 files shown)


jrha:
@stdweird happy to merge?

PR #1222, opened 2018-01-03
ncm-cron: Implement some validation of cron timing fields

jrha:
Add a function to validate the contents of cron timing fields.

I've got tired of finding cron jobs which are not running due to simple mistakes such as entering day names in the `day` field rather than the `weekday` field due to the almost complete lack of any validation on this component.

This could get pretty complex and most of the public domain regexes I found do a poor job of covering the accepted syntax, so for now we'll just perform some basic sanity checking on numbers and text found in the fields.

While we're touching this, also improve the annotations a bit.

### DIFF
--- a/ncm-cron/src/main/pan/components/cron/schema.pan
+++ b/ncm-cron/src/main/pan/components/cron/schema.pan
@@ -38,13 +38,101 @@ type structure_cron_log = {
     'mode' ? string
 } with structure_cron_log_valid(SELF);
 
+@documentation{
+    Validate contents of cron timing fields (see CRONTAB(5) for details)
+
+    Cron timing fields can contain complex expressions (e.g. "1,5,13-23/2"). Rather than validate these in
+    depth the aim here is to catch things that are obviously wrong, such as:
+        * characters which are not valid in cron fields
+        * out of range numbers (e.g. "35" in the hour field)
+        * names in the wrong field (e.g. "tue" in the day of month field)
+}
+function valid_cron_timing = {
+    if (ARGC != 4) {
+        error(format('%s: expected 4 parameters, received %d', FUNCTION, ARGC));
+    };
+
+    timing_value = to_lowercase(ARGV[0]);
+    lower_bound = ARGV[1];
+    upper_bound = ARGV[2];
+    text_regex = ARGV[3];
+
+    # Check that the field contains only valid characters
+    if (!match(timing_value, '^(?:[a-z0-9/*-,]+)$')) error(format('"%s" contains invalid characters', timing_value));
+
+    # Find runs of digits and validate them against provided bounds
+    foreach(k; v; matches(timing_value, '([0-9]+)')) {
+        i = to_long(v);
+        if (i < lower_bound) error(format('Value %d is below lower bound of %d', i, lower_bound));
+        if (i > upper_bound) error(format('Value %d is above upper bound of %d', i, upper_bound));
+    };
+
+    # Find runs of letters and validate them against provided regex
+    foreach(k; v; matches(timing_value, '([a-z]+)')) {
+        if (!match(v, text_regex)) error(format('"%s" is not a valid value for this field item', v));
+    };
+
+    # Ignore all other characters
+    true;
+};
+
+@documentation{ Convenience wrapper for validating cron minute field }
+function valid_cron_minute = valid_cron_timing(ARGV[0], 0, 59, '^(?![a-z]).+$');
+
+@documentation{ Convenience wrapper for validating cron hour field }
+function valid_cron_hour = valid_cron_timing(ARGV[0], 0, 23, '^(?![a-z]).+$');
+
+@documentation{ Convenience wrapper for validating cron day of month field }
+function valid_cron_day_of_month = valid_cron_timing(ARGV[0], 1, 31, '^(?![a-z]).+$');
+
+@documentation{ Convenience wrapper for validating cron month field }
+function valid_cron_month = valid_cron_timing(ARGV[0], 1, 12, '^(?:jan|feb|mar|apr|may|jun|jul|aug|sep|oct|nov|dec)$');
+
+@documentation{ Convenience wrapper for validating cron day of week field }
+function valid_cron_day_of_week = valid_cron_timing(ARGV[0], 0, 7, '^(?:mon|tue|wed|thu|fri|sat|sun)$');
+
 type structure_cron_timing = {
-    'minute' ? string
-    'hour' ? string
-    'day' ? string
-    'month' ? string
-    'weekday' ? string
+    @{ minute of hour (0-59) }
+    'minute' ? string with valid_cron_minute(SELF)
+    @{ hour of day (0-23) }
+    'hour' ? string with valid_cron_hour(SELF)
+    @{ day of month (1-31) }
+    'day' ? string with valid_cron_day_of_month(SELF)
+    @{ month of year (1-12 or three-letter abbreviated lowercase name) }
+    'month' ? string with valid_cron_month(SELF)
+    @{ day of week (0-7 or three-letter abbreviated lowercase name) }
+    'weekday' ? string with valid_cron_day_of_week(SELF)
+    @{ Interval (in minutes) over which to randomly smear the start time of the job }
     'smear' ? long(0..1440)
+};
+
+@documentation{
+    Validate contents of cron frequency field
+}
+function valid_cron_frequency = {
+    if (ARGC != 1) {
+        error(format('%s: expected 1 parameter, received %d', FUNCTION, ARGC));
+    };
+
+    frequency = to_lowercase(ARGV[0]);
+
+    if (match(frequency, '^@(?:reboot|yearly|annually|monthly|weekly|daily|midnight|hourly)$')) {
+        return(true);
+    };
+
+    fields = split(' ', frequency);
+
+    if (length(fields) != 5) {
+        error(format('cron frequency "%s" should have 5 fields, it only has %d', frequency, length(fields)));
+    };
+
+    valid_cron_minute(fields[0]);
+    valid_cron_hour(fields[1]);
+    valid_cron_day_of_month(fields[2]);
+    valid_cron_month(fields[3]);
+    valid_cron_day_of_week(fields[4]);
+
+    true;
 };
 
 type structure_cron = {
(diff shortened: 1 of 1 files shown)


ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,12 +38,63 @@ type structure_cron_log = {
...
+        # Find all strings of letters and validate them against provided regex
+        foreach(k; v; matches(test_value, '([a-z]+)')) {
+            result = result && match(v, text_regex);
+            if (!result) error(format('"%s" is not a valid value for this field item', v));
+        };
+
+        # Ignore all other characters
+        return(result);
+    } else {
+        error(format('valid_cron_field_items: expected 4 parameters, received %d', ARGC));
```
stdweird: i prefer a `if (ARGC != 4) error(...)` in the beginning to avoid an extra indentation
also, you can use builtin `FUNCTION` with format for the function name

jrha:
Validation of the `frequency` field coming shortly.

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,12 +38,63 @@ type structure_cron_log = {
     'mode' ? string
 } with structure_cron_log_valid(SELF);
 
+@documentation{
+    Validate contents of cron fields
```
stdweird: can you add a (not too trivial) example of what retuns true and false. i read the code twice, i'm serioulsy confused with the foreach loops

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,12 +38,63 @@ type structure_cron_log = {
...
+}
+function valid_cron_field = {
+    if (ARGC == 4) {
+        test_value = ARGV[0];
+        lower_bound = ARGV[1];
+        upper_bound = ARGV[2];
+        text_regex = ARGV[3];
+
+        # Check that the field contains only valid characters
+        result = match(test_value, '^(?:[a-z0-9/*-,]+)$');
```
stdweird: no `_` allowed?

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,12 +38,63 @@ type structure_cron_log = {
...
+        if (!result) error(format('"%s" contains invalid characters', test_value));
+
+        # Find all strings of digits and validate them against provided bounds
+        foreach(k; v; matches(test_value, '([0-9]+)')) {
+            i = to_long(v);
+            if (i < lower_bound) error(format('Value %d is below lower bound of %d', i, lower_bound));
+            if (i > upper_bound) error(format('Value %d is above upper bound of %d', i, upper_bound));
+            result = result && i >= lower_bound && i <= upper_bound;
```
stdweird: is the `i >= lower_bound && i <= upper_bound` really needed here? you can't get here otherwise (in which case `result` is not needed either)

ncm-cron/src/main/pan/components/cron/schema.pan
jrha: Not unless cron allows it, no.

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,12 +38,63 @@ type structure_cron_log = {
...
+
 type structure_cron_timing = {
-    'minute' ? string
-    'hour' ? string
-    'day' ? string
-    'month' ? string
-    'weekday' ? string
+    @{ minute of hour (0-59) }
+    'minute' ? string with {
+        valid_cron_field(to_lowercase(SELF), 0, 59, '^(?![a-z]).+$')
```
stdweird: i would move the `to_lowercase` to the function

jrha:
@stdweird thanks for the review, hopefully better now.

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
...
+        if (i > upper_bound) error(format('Value %d is above upper bound of %d', i, upper_bound));
+    };
+
+    # Find runs of letters and validate them against provided regex
+    foreach(k; v; matches(timing_value, '([a-z]+)')) {
+        if (!match(v, text_regex)) error(format('"%s" is not a valid value for this field item', v));
+    };
+
+    # Ignore all other characters
+    return(true);
```
ned21: Stylistically it's better to write:
```
# Ignore all other characters
true;
```

because it's better to avoid return() for performance reasons (apparently).

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
...
+    };
+
+    frequency = to_lowercase(ARGV[0]);
+
+    if (match(frequency, '^(?:@reboot|@yearly|@annually|@monthly|@weekly|@daily|@midnight|@hourly)$')) {
+        return(true);
+    };
+
+    fields = split(' ', frequency);
+    if (length(fields) != 5) error(format('cron frequency "%s" should have five fields', frequency));
```
ned21: Please reformat - it's easy to mis-read this as only applying the below lines if the condition is true, as opposed to this being a one-line conditional.

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
...
+    valid_cron_minute(fields[0]);
+    valid_cron_hour(fields[1]);
+    valid_cron_day_of_month(fields[2]);
+    valid_cron_month(fields[3]);
+    valid_cron_day_of_week(fields[4]);
+    return(true);
```
ned21: See comment above about return() being not recommended

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
     Define specific attributes for cron log file.
 }
 type structure_cron_log = {
-    @{A boolean disabling the redirection of script output/error to a log file}
+    @documentation{A boolean disabling the redirection of script output/error to a log file}
```
stdweird: please don't change this. we don't do this for any attribute, only for the actual types. i'm nit sure the docs are generated correctly when this is changed

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
...
     'owner' ? string
-    @{Permissions of log file specified as a string interpreted as an octal number.}
+    @documentation{Permissions of log file specified as a string interpreted as an octal number.}
     'mode' ? string
 } with structure_cron_log_valid(SELF);
 
+@documentation{
+    Validate contents of cron timing fields (see CRONTAB(5) for details)
+
+    Cron timing fields can contain complex expressions (e.g. "1,5,13-23/2"). Rather than validate these in
```
stdweird: ok, thanks, much clearer now!

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -27,40 +27,123 @@ type structure_cron_syslog = {
...
+    Validate contents of cron frequency field
+}
+function valid_cron_frequency = {
+    if (ARGC != 1) {
+        error(format('%s: expected 1 parameter, received %d', FUNCTION, ARGC));
+    };
+
+    frequency = to_lowercase(ARGV[0]);
+
+    if (match(frequency, '^(?:@reboot|@yearly|@annually|@monthly|@weekly|@daily|@midnight|@hourly)$')) {
```
stdweird: why the `(?:..)`? also, factor out the leading `@` unless you plan to switch to choice at some point

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -73,24 +156,26 @@ type structure_cron = {
...
       the variable name, value is variable value.}
     'env' ? string{}
     'log' ? structure_cron_log
     'syslog' ? structure_cron_syslog
 } with {
     if (exists(SELF['log']) && exists(SELF['syslog'])) {
-        error("At most one of log or syslog can be defined");
+        error("Only one of log or syslog may be defined, not both.");
+    };
+    if (exists(SELF['timing']) && exists(SELF['frequency'])) {
```
stdweird: i have seen constructs like this in other schemas as well, maybe we should add function to pan to test for this, eg
`oneof(SELF, 'timing', 'frequency')` (or above `maxone(SELF, 'log', 'syslog')`). obviously these shouldn't call `error`, but this is a lot of boilerplate imho.

jrha: I agree, but this probably indicates that the schema is badly designed.

ncm-cron/src/main/pan/components/cron/schema.pan
jrha: Wrapping things in non-capturing groups is just force of habit, but yet I had thought that this would be an ideal use for `choice`.

ncm-cron/src/main/pan/components/cron/schema.pan
jrha: Thanks, this managed to survive re-factoring somehow :stuck_out_tongue:

ncm-cron/src/main/pan/components/cron/schema.pan
jrha: Good point, hopefully clearer now.

ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -38,15 +38,103 @@ type structure_cron_log = {
...
+    foreach(k; v; matches(timing_value, '([a-z]+)')) {
+        if (!match(v, text_regex)) error(format('"%s" is not a valid value for this field item', v));
+    };
+
+    # Ignore all other characters
+    true;
+};
+
+@documentation{ Convenience wrapper for validating cron minute field }
+function valid_cron_minute = valid_cron_timing(ARGV[0], 0, 59, '^(?![a-z]).+$');
```
stdweird: ah, more test_is_type crap, we do need https://github.com/quattor/pan/pull/123 .....

stdweird:
@jrha milestone?

ncm-cron/src/main/pan/components/cron/schema.pan
jrha: Indeed, these would be types if they weren't re-used to check the `frequency` field.

jrha:
Sooner the better? Well tested here.

ncm-cron/src/main/pan/components/cron/schema.pan
stdweird: or if we had a panc release with the validate function.... grrrr

jrha:
I've flagged this as potentially **backwards incompatible** as it is a schema change and should catch things that would have previously compiled without errors.

PR #1226, opened 2018-01-23
ncm-cron: Support magic AUTO value in minutes field of frequency

jrha:
Resolves issue reported in quattor/release#314 caused by #1222.

Should really be replaced by use of `smear` in `timing` structure, but there is too much code in the wild using `AUTO` to remove support for it right now. #1227 opened to track this in the longer term.

### DIFF
--- a/ncm-cron/src/main/perl/cron.pm
+++ b/ncm-cron/src/main/perl/cron.pm
@@ -249,7 +249,7 @@ sub Configure
         # Frequency of the cron entry.  May contain AUTO for the
         # minutes field : in this case, substitute AUTO with a random
         # value. This only works in the minutes field of the
-        # frequence.  We support two formats here: the traditional
+        # frequency.  We support two formats here: the traditional
         # "frequency" field and also a more complex "timing" structure
         # (which allows more smear)
         my $frequency = "";
@@ -303,7 +303,7 @@ sub Configure
             $frequency = $entry->{frequency};
 
             # Substitute AUTO with a random value. This only works in
-            # the minutes field of the frequence.
+            # the minutes field of the frequency.
             $frequency =~ s/AUTO/int(rand(60))/eg;
         }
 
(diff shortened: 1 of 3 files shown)


Pansanel:
Looks fine for me. It should fix the issue reported yesterday.

jrha:
Will add a unit test in a second.

ncm-cron/src/test/resources/cron_syslog-common.pan
```diff
@@ -55,4 +55,11 @@ include 'components/cron/config';
     "command", "some command",
 ));
 
+"/software/components/cron/entries" = append(SELF, dict(
+    "name", "test_magic_auto_value",
+    "user", "magic",
+    "frequency", "AUTO 7 * * *",
```
stdweird: AUTO -> auto

jrha: No, this is correct, existing code in template libraries only uses uppercase.
I've just updated the component to a case-insensitve substitution to allow either.

stdweird: but the scheam does not allow this

ncm-cron/src/main/perl/cron.pm
```diff
@@ -304,7 +304,7 @@ sub Configure
 
             # Substitute AUTO with a random value. This only works in
             # the minutes field of the frequence.
```
stdweird: `frequence` -> `frequency`

ncm-cron/src/test/resources/cron_syslog-common.pan
jrha: I'll add test cases for both.

ncm-cron/src/main/perl/cron.pm
jrha: Yeah... this code is special.

jrha:
Ok, I've made a bit of a mess getting there, but this should be the correct validation now.

PR #1194, opened 2017-11-24
ncm-nfs: mount type for cephfs is ceph

kwaegema:
Tested @ugent, see also http://docs.ceph.com/docs/master/cephfs/fstab/

PR #1229, opened 2018-01-24
ncm-nfs: fix ceph fs type in nfs.pm

kwaegema:
fix a bug where every run of ncm-nfs would create an additional entry in fstab for ceph mounts because of wrong match regex


jrha:
Related to #1194 I presume?

kwaegema:
funny note: we only see this when /software/components/fstab/keep/fs_types contains 'ceph' (which it should when it is managed by ncm-nfs). If not, ncm-fstab has removed the line before ncm-nfs adds it again..

kwaegema:
@jrha indeed

PR #1231, opened 2018-01-24
ncm-cron: Escape hyphen as we don't want it to be treated as a range

jrha:
The `*-,` part was being treated as a range of characters between `*` and `,` which is obviously nonsense.

### DIFF
--- a/ncm-cron/src/main/pan/components/cron/schema.pan
+++ b/ncm-cron/src/main/pan/components/cron/schema.pan
@@ -58,7 +58,7 @@ function valid_cron_timing = {
     text_regex = ARGV[3];
 
     # Check that the field contains only valid characters
-    if (!match(timing_value, '^(?:[a-z0-9/*-,]+)$')) error(format('"%s" contains invalid characters', timing_value));
+    if (!match(timing_value, '^(?:[a-z0-9/*\-,]+)$')) error(format('"%s" contains invalid characters', timing_value));
 
     # Find runs of digits and validate them against provided bounds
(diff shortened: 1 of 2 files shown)


ncm-cron/src/main/pan/components/cron/schema.pan
```diff
@@ -58,7 +58,7 @@ function valid_cron_timing = {
     text_regex = ARGV[3];
 
     # Check that the field contains only valid characters
-    if (!match(timing_value, '^(?:[a-z0-9/*-,]+)$')) error(format('"%s" contains invalid characters', timing_value));
+    if (!match(timing_value, '^(?:[a-z0-9/*\-,]+)$')) error(format('"%s" contains invalid characters', timing_value));
```
stdweird: hah, nice catch. you can also put the `-` last

ncm-cron/src/test/resources/cron_syslog-common.pan
```diff
@@ -62,4 +62,11 @@ include 'components/cron/config';
     "command", "a scary command",
 ));
 
+"/software/components/cron/entries" = append(SELF, dict(
+    "name", "test_range_skip",
+    "user", "skippy",
+    "frequency", "15 0-23/3 * * *",
+    "command", "Fifteen minites past every third hour",
```
stdweird: minites -> minutes

jrha: Doh. Fixed.

jrha:
💚

PR #1236, opened 2018-02-07
ncm-spma: yum: Support multiple reposdirs

jrha:
Not for merging yet, only tests so far.


jrha:
Locally I'm seeing the `FileEditor` inserting extra blank lines again...

jrha:
Found and fixed the source of the extra blank lines, also the tests which were actually checking for this strange behaviour.

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -1015,7 +1015,7 @@ sub configure_yum
         $fh->add_or_replace_lines(
             $name,
             $name. q{\s*=\s*}.$valuereg.'$',
-            "\n$name=$value\n",
+            "$name=$value\n",
```
stdweird: what if the file doesn't end with a newline? i'm not sure FileEditor tests this...

stdweird: seems it does by default: https://github.com/quattor/CAF/blob/master/src/main/perl/FileEditor.pm#L299 and https://github.com/quattor/CAF/blob/master/src/main/perl/FileEditor.pm#L349

can you modify the unittest below to check this?

jrha: FileEditor should deal with that, not the components.

jrha: The match should still work though...

jrha:
@stdweird I've been trying to write an end-to-end test case that uses the new property in a profile and then checks that it exists in `yum.conf` but can't figure out how to get it to work. Are there any of the existing tests that I can steal from/extend?

stdweird:
no end to end tests. but `yum-configure.t` should show more dirs passed to `configure_yum`. and `configure_yum` is tested.

jrha:
Also: any opinions on the name of the schema option?

ncm-spma/src/main/perl/spma/yum.pm
jrha: Done, the noaction test now starts with a file that has a missing newline at EOF.

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -37,6 +37,7 @@ type component_spma_yum = {
     "run" ? legacy_binary_affirmation_string # Run the SPMA after configuring it
     "userpkgs_retry" : boolean = true
     "userpkgs" ? legacy_binary_affirmation_string # Allow user packages
+    "extra_reposdirs" ? absolute_file_path[] # List of additional repo dirs to be included
```
stdweird: add comment as annotation. also add that the priority: indentical repos in the extra dir precede (or not?) the quattor generated ones.

stdweird: wrt name: just `reposdirs`? or `userreposdirs`?

jrha: Just `reposdirs` seems fine, I'll make sure the annotation explains what happens with regards to the managed repodir. As far as precendence... I might have to experiment to see how yum behaves.

ned21: Given the yum.conf option is `reposdir` I think `reposdirs` makes sense. The only downside is this is additional config to be added to `reposdir`, there is no way to remove the default? OTOH that seems like such a niche use case that it's OK?

jrha:
I've been playing around on a host with multiple `reposdirs` configured to see if one takes priority over another and there doesn't appear to be any form of precedence. It behaves the same as it does with all definitions in the same directory (or even same file).

All the repos are read and in the case of duplicate definitions, yum issues a warning and one of them (I haven't been able to predict which one) wins.

ned21:
@jrha We should be able to detect duplicates in PAN and error out before it's deployed? For the priorities I think you need [yum-plugin-priorities](https://wiki.centos.org/PackageManagement/Yum/Priorities). So the schema should allow the setting of priorities - does that mean the list needs to become a dict?

jrha:
@ned21 priorities are already supported in the individual repositories and work fine as is (we make heavy use of them). Unfortunately duplicates between managed and unmanaged repositories will be impossible to catch at build time, but that's the price you pay if you are want this functionality.

jrha:
This has been well tested at RAL and seems to be working well for us. Could someone re-review it?

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -37,6 +37,8 @@ type component_spma_yum = {
     "run" ? legacy_binary_affirmation_string # Run the SPMA after configuring it
     "userpkgs_retry" : boolean = true
     "userpkgs" ? legacy_binary_affirmation_string # Allow user packages
+    @{ List of additional repo dirs to be included }
```
stdweird: make it more clear that this are additional repository directories that are not managed by ncm-spma. maybe replacing `additional` with `external`?

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -1168,6 +1168,9 @@ sub Configure
     # TODO: for userpkgs, insert quattor unmanaged dir as first (so it takes priority)
     # TODO: check that the first one wins
     my $reposdir = [$quattor_managed_reposdir];
+    if ($t->{reposdirs}) {
+        $reposdir = [@$reposdir, @{$t->{reposdirs}}];
```
stdweird: `push(@$reposdir, @{$t->{reposdirs}})`

stdweird:
@jrha tiny remarks

jrha:
@stdweird comments addressed.

PR #1240, opened 2018-02-14
ncm-gpfs: support curl with GSS

stdweird:
Support gpfs data retrieval using GSSAPI (eg kerberos in FreeIPA only environment); inspired by ncm-download


jrha:
Seems reasonable to me... Could you provide some context in the description please?

stdweird:
@jrha better now?

jrha:
😄

PR #1244, opened 2018-03-01
ncm-spma: yum: support package filter

stdweird:


jrha:
Why? What's the use case?

stdweird:
we want to run spmalight (which is spma with filter to install ncm-accounts), then ncm-accoutns (to fix the local caaounts instead of the guess work or hardcoded values in packages) and then ncm-spma (without filters).

still testing though :smile:

jrha:
I see! Interesting...

stdweird:
seems to be working with following code to run ncm-accounts before spma

```pan
# spmalight config
"/software/components/spmalight" = value("/software/components/spma");
bind "/software/components/spmalight" = component_spma_yum;
prefix "/software/components/spmalight";
"userpkgs" = null;
"filter" = "^ncm-(accounts|spma)$";
"ncm-module" = "spma";
"dependencies/post" = append('spma');

"/software/components/spma/dependencies/pre" = append('accounts');

"/software/components/accounts/dependencies/pre" = {
    deps = list();
    foreach(idx; dep; SELF) {
        if (dep == 'spma') {
            dep = 'spmalight';
        };
        deps = append(deps, dep);
    };
    deps;
};
```

ncm-spma/src/main/pan/components/spma/yum/schema.pan
```diff
@@ -37,6 +37,9 @@ type component_spma_yum = {
     "run" ? legacy_binary_affirmation_string # Run the SPMA after configuring it
     "userpkgs_retry" : boolean = true
     "userpkgs" ? legacy_binary_affirmation_string # Allow user packages
+    @{regexp pattern to install only matching (unescaped) package names.
+        When userpkgs is not defined, it runs as if userpkgs is true.}
```
ned21: Which makes me wonder, what happens if `userpkgs` is defined and equal to "no"? Does running with `filter` cause it to remove everything? (which would seem inefficient 😀)

ned21:
So nam-accounts ensures accounts are in place prior to the package being installed, so any package that does a `getent` finds the account already there and carries on without creating a local account? That makes sense. So `filter` is primarily intended to be use in a sub-classed component rather than spma itself? It would be good to see this written up as a tutorial/article on quattor.org.

One other suggestion, does the dependency code work more efficiently if you do this? (To avoid the resource copy):
```
"/software/components/accounts/dependencies/pre" = {
    foreach(idx; dep; SELF) {
        if (dep == 'spma') {
            SELF[idx'] = 'spmalight';
        };
    };
    SELF;
};
```

ncm-spma/src/main/pan/components/spma/yum/schema.pan
stdweird: yes. i can also add explicitly add it to the annotation, but the component does as instructed.

stdweird:
i'll include a `components/spma/light` template and add annotation in it. not going to write an article for 10 lines of pan code :wink:

ncm-spma/src/main/pan/components/spma/yum/schema.pan
ned21: Oh, I missed and important bit - this introduces a new component via a template only change:
```
"/software/components/spmalight" = value(...
```

Yeah, a light.pan along with a few lines of explanation at the top would be the best way to handle describing it.

stdweird:
ok some bugs:
 - the current filter also removes the versionlocklist, and then yum distro-sync really doens't do what you want it to do: i'm not sure what to do here. probably disabling the distro-sync of something like that (or generate the verisonlock list without filter)

 - the current code does nto garantee that spma runs right after accounts, others might jump in between. this might be due to some other code we have in place, remains to be seen

stdweird:
@ned21 @jrha please rereview
i hope the templates make it clear enough

stdweird:
@ned21 i pushed another unittest and squashed the commit. https://github.com/quattor/configuration-modules-core/pull/1244/files#diff-eea37931bd644c06c5f5922ee7def4bbR237

stdweird:
@jrha rebased and resolved

ncm-spma/src/main/pan/components/spma/light.pan
```diff
@@ -0,0 +1,47 @@
...
+    2. accounts component
+    3. (full/regular) spma component run
+
+  The use case is to run the accounts component before (full) spma,
+  to make sure certain local users are created by
+  ncm-accounts and not by the packages.
+}
+unique template components/spma/light;
+
+include 'components/spma/config';
```
jrha: Also…
```pan
include 'components/accounts/config';
```

stdweird:
@jrha added

jrha:
@stdweird I'm pretty sure the unit tests are failing because the accounts config needs to be included after spma so that `pkg_repl` is defined before first use.

stdweird:
@jrha travis fails due to panlint bug https://github.com/quattor/pan/pull/196

jrha:
FIx merged, tests re-run, now :green_apple:

stdweird:
test this please

stdweird:
@jrha test fails due to https://github.com/quattor/maven-tools/pull/172/files, so we need new build tools

jrha:
@stdweird new build tools are out, feel free to start bumping versions!

jrha:
#1257 is merged, time to rebase?

stdweird:
retest this please

stdweird:
@jrha no, afaik jenkins tests a PR merged in master, not the branch on its own

jrha:
😞

PR #1246, opened 2018-03-06
ncm-metaconfig: Add support for OpenStack Nova

jeremyspencer39171:
Allows for multiple lines in the config to have the key `pci_alias`.
This is required to set multiple values for `pci_alias` in `nova.conf` on an openstack controller.

For example:
```pan
"pci_alias", list(
    '{ "vendor_id":"10de", "product_id":"13bb", "device_type":"type-PCI", "name":"nvidia-quadro-k620-vga"}',
    '{ "vendor_id":"10de", "product_id":"15f0", "device_type":"type-PCI", "name":"quadro_gp100"}',
)
```

Will give the following output:
```ini
pci_alias = { "vendor_id":"10de", "product_id":"13bb", "device_type":"type-PCI", "name":"nvidia-quadro-k620-vga"}
pci_alias = { "vendor_id":"10de", "product_id":"15f0", "device_type":"type-PCI", "name":"quadro_gp100"}
```


jrha:
@hpcugentbot please test this

jrha:
test this please

ncm-metaconfig/src/main/metaconfig/nova/main.tt
```diff
@@ -0,0 +1,23 @@
+[%- FOR section IN CCM.contents.pairs -%]
+[[%-    section.key -%]]
+[%      FOREACH item IN section.value.pairs -%]
+[%          IF (item.value.0 == 'DO-MULTI-LINE-CONFIG') -%]
```
jrha: Well ok, but magic values are evil... instead how about doing something like:
```
 +[%          IF (section.key == 'pci_alias') -%] 
```

stdweird:
please have a look at https://github.com/quattor/configuration-modules-core/pull/1192. includes schema and TT files

jrha:
@jeremyspencer39171 started this before the other PR, but yes it should ideally be merged into that work. 
Nova is a bit "special", it has it's own rather broken config parser, but again the TT work could be combined...

jeremyspencer39171:
I'll have a go at adding this to #1192.
Also realised that I left everything mandatory in the schema.

jeremyspencer39171:
I read through #1192.
I think this functionality could be added to there with this TT file and something like this line in here next to the daemon selection:
ncm-openstack/src/main/perl/OpenStack/Nova.pm
`$self->{tt} = $self->{hypervisor} ? 'common' : 'nova_controller';`
[nova_controller.tt.txt](https://github.com/quattor/configuration-modules-core/files/1811197/nova_controller.tt.txt)
(Apologies for the txt file, github doesn't accept .tt files. And adding it as a code block makes it lose all the whitespace)

However, #1192 seems to be for the more current versions of OpenStack, we are running Mitaka at the moment.
So I'm not sure if this will still be needed or if OpenStack have fixed the configuration.
I know that the "pci_alias" config moves from [DEFAULT] to [pci] and becomes "alias" (Ocata onwards), but I don't know if that changes the way we can define multiple aliases (on the controller) to something more sensible.

jrha:
@apdibbo is this still needed in light of ncm-openstack?

jeremyspencer39171:
I'm going to close this. I'm fairly certain it's not needed any more :)

PR #1247, opened 2018-03-09
ncm-sudo: spring cleanup

stdweird:
Backwards incompatible due to
 * renamed schema related functions, types and templates
 * both spma and accounts are now predependencies


stdweird:
@ned21 is the accounts predependency an issue for MS?
you can easily override it with site templates, once https://github.com/quattor/maven-tools/pull/170 is released and the new release makes use of it

stdweird:
also, we need the accounts and spma dependency for spmalight

ned21:
Yes the dependency on accounts is a problem for us (and anyone else not using something other than ncm-accounts). I don't think renaming schema functions or types is particularly backwards incompatible (unless you are reusing them outside the component). Changing the schema itself would be a real problem for us because the number of Quattor versions we have in use is high, and in order to keep up with the code versions we are no longer version locking the component rpms in templates. So we have one set of templates and then control the rpm versions installed by repointing the yum repos on the host.

I haven't read the second commit in detail yet.

stdweird:
@ned21 wet dependencies on spma and accounts, is the current code with possibility of site-config ok? (once the fixes are released)

ned21:
The second code looks fine.  I think being able to remove the account dependency via site-defaults will be fine since we already remove spma that way.

PR #1248, opened 2018-03-09
ncm-useraccess: spring cleanup

stdweird:
Backwards incompatible: sets both spma and accounts as predependency


stdweird:
@ned21 @jrha same remarks as with #1247

ned21:
Not familiar with this component but code looks OK. Would be better for someone who actually uses it to check it more closely!

jrha:
Hmm, not an issue with this change per-se, but do all of the test profiles actually get compiled? One of the files that you renamed that has been flagged by Panlint looks suspiciously like it should not compile, see [lines 20-22 of testrolesimple.pan](https://github.com/stdweird/configuration-modules-core/blob/cb92254c2b1531882d4b3a7d80469ece4c8482b5/ncm-useraccess/src/test/resources/testrolesimple.pan#L20-L22).

stdweird:
@jrha only files used in perl unittests get used. i'll fix this

jrha:
If they're not used and don't work just delete them?

PR #1256, opened 2018-04-09
ncm-metaconfig: add mailrc service

JensTimmerman:
This service allows you to configure mail.rc or .mailrc files

for an example see https://coderwall.com/p/ez1x2w/send-mail-like-a-boss

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/mailrc/tests/regexps/account_config
+++ b/ncm-metaconfig/src/main/metaconfig/mailrc/tests/regexps/account_config
@@ -1,0 +1,9 @@
+account config test for mailrc
+---
+/etc/mail.rc
+multiline
+---
+^account gmail \{$
+^set from=root@gmail.com$
+^set smtp=mailserver.gmail.com$
+^\}$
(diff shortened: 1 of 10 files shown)


ncm-metaconfig/src/main/metaconfig/mailrc/tests/regexps/account_config
```diff
@@ -0,0 +1,9 @@
+account config test for mailrc
+---
+/etc/mail.rc
+multiline
+---
+^account gmail {$
```
stdweird: you cannot have unescaped leftbrace here

ncm-metaconfig/src/main/metaconfig/mailrc/tests/regexps/account_config
```diff
@@ -0,0 +1,9 @@
...
+^set from=root@gmail.com$
+^set smtp=mailserver.gmail.com$
+^}$
```
stdweird: this very likely also requires unescaping

JensTimmerman: escaping?

jrha:
Looks good. What is the Dockerfile for?

jrha:
Cool! Any chance you could factor it out into a separate PR? Having it mixed in with the mailrc change is a bit confusing.

JensTimmerman:
Yeah, sorry for not getting back to you, but I'm no longer employed at UGent as a sysadmin, so I'm afraid I won't be following up on quattor development any more in the future.

PR #1264, opened 2018-05-15
ncm-sudo: Test names of alias list entries, rather than list itself

jrha:
Aliases are dictionaries of lists with the alias name as the key, but the validation code was treating them as a lists of alias names, so bad times ensued.

Fixes compilation problems with example template library.


stdweird:
@jrha thx. i'm not sure how i missed that.

jrha:
Feels like it should have been caught by a test earlier than in the template library.

PR #1268, opened 2018-05-21
ncm-grub: add pxeboot option to set pxe bootorder (using efibootmgr)

stdweird:


ncm-grub/src/main/perl/grub.pm
```diff
@@ -18,6 +19,7 @@ Readonly my $GRUB2_DIR           => '/boot/grub2';
 Readonly my $GRUB2_USER_CFG      => "$GRUB2_DIR/user.cfg";
 Readonly my $GRUBBY              => '/sbin/grubby';
 Readonly my $PREFIX              => '/boot';
+Readonly my $EFIBOOTMGR          => '/sbin/efibootmgr';
```
jrha: It's `/bin/efibootmgr` on debian if you care.

PR #1271, opened 2018-05-28
ncm-systemd: Do not start/stop services during boot and shutdown

gombasg:
If ncm-systemd is called during boot or shut-down, then it should avoid
starting/stopping services itself, as that could interfere with the
standard ordering of events. E.g., on shut-down, it could start services
again which were just stopped.

### DIFF
--- a/ncm-systemd/src/main/perl/Systemd/Service.pm
+++ b/ncm-systemd/src/main/perl/Systemd/Service.pm
@@ -5,8 +5,8 @@ use 5.10.1;
 use parent qw(CAF::Object Exporter);
 
 use NCM::Component::Systemd::Service::Chkconfig;
-use NCM::Component::Systemd::Service::Unit qw(:states :types);
-use NCM::Component::Systemd::Systemctl qw(systemctl_command_units);
+use NCM::Component::Systemd::Service::Unit qw(:states :types :targets);
+use NCM::Component::Systemd::Systemctl qw(systemctl_command_units systemctl_is_active);
 
 use CAF::Object qw (SUCCESS);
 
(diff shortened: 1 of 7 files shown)


ncm-systemd/src/main/perl/Systemd/Systemctl.pm
```diff
@@ -365,14 +365,53 @@ sub systemctl_is_enabled
...
     # Do not test on $ec; if unit is disabled, it is-enabled also returns ec > 0
     # If there's a real issue, like unknown unit, there is no stdout.
     return $stdout || undef;
 }
 
 
 
+=pod
+
+=item systemctl_is_active
```
stdweird: this is copy/paste code from systemctl_is_enabled. can you factor out the common code in a new function and pass the `is-active` and `is-enabled` to this new private function?

ncm-systemd/src/main/perl/Systemd/Service.pm
```diff
@@ -619,16 +619,26 @@ sub change
...
-    # TODO: same TODOs as with states
-    foreach my $act (sort keys %$acts) {
-        my @units = @{$acts->{$act}};
-
-        # TODO: process units wrt dependencies?
-        if (@units) {
-            # TODO: trap exitcode and stop any further processing in case of error?
-            # CAF::Process logger is sufficient
-            systemctl_command_units($self, $change_activation->{$act}, @units);
+    my $multi_user_active = systemctl_is_active($self, "multi-user.target");
```
stdweird: can you make a readonly/constant for the 'multi-user.target' and use that instead of the magic string

my only remark is if there would be a reason not to do this, and thus to make this configurable to skip. i can't think of any, so i'm just asking.

ncm-systemd/src/main/perl/Systemd/Service.pm
```diff
@@ -619,16 +619,26 @@ sub change
...
+
+            # TODO: process units wrt dependencies?
+            if (@units) {
+                # TODO: trap exitcode and stop any further processing in case of error?
+                # CAF::Process logger is sufficient
+                systemctl_command_units($self, $change_activation->{$act}, @units);
+            }
         }
+    } else {
+        $self->info("multi-user.target is not active, not starting or stopping any services.");
```
stdweird: would it make sense to eg wait and recheck a few times, instead of doing nothing? in case a boot, it could run later whne multi-user target is active.

i guess this should also return an error (since nothing was done). and ncm-ncd should be rerun

gombasg: During shutdown, waiting and retrying would definitely not do any good. I'm less sure about boot time, but I would still be hesitant to interfere with the boot sequence, because experience tells us dependencies are not always correct and/or complete.

If you really want to make sure services which were changed to be enabled while the host was down are started after reboot, it would be possible to schedule `ncm-ncd --configure systemd` to run _after_ `multi-user.target` has been reached.

PR #1272, opened 2018-05-29
ncm-spma: yum: disable distro-sync when the package filter is in use

stdweird:


stdweird:
This is not the way to do it. New PR upcoming

PR #1274, opened 2018-05-29
ncm-spma: yum: distro-sync only filtered packages (when filter is defined)

stdweird:
This prevents running full distro-sync when filter is in use; which might trigger updates of any number of packages and possible installations of dependencies.

### DIFF
--- a/ncm-spma/src/test/perl/yum-wanted.t
+++ b/ncm-spma/src/test/perl/yum-wanted.t
@@ -72,6 +72,28 @@ my $wanted = {
        },
 };
 
+=head1 valid_packages
+
+=cut
+
+my $vpkgs = $cmp->valid_packages($wanted);
+#diag explain $vpkgs;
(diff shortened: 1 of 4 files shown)


ncm-spma/src/test/perl/yum-wanted.t
```diff
@@ -72,6 +72,28 @@ my $wanted = {
        },
 };
 
+=head1 valid_packages
+
+=cut
+
+my $vpkgs = $cmp->valid_packages($wanted);
+#diag explain $vpkgs;
```
jrha: Is this supposed to be commented out?

jrha:
LGTM other than question above

ncm-spma/src/test/perl/yum-wanted.t
stdweird: yes, it's too verbose by default

PR #1282, opened 2018-06-04
ncm-cron: fix test of /system/archetype/os

### LINKED ISSUE #1281: ncm-cron 18.3.0: invalid /system/archetype/os test
As mentioned in https://github.com/quattor/template-library-core/pull/177, `config.pan` in `ncm-cron` is testing `/system/archetype/os` as a string when it was changed to a `dict` a while ago. This need to be fixed for `ncm-cron` be usable with Aquilon.

jouvin:
- /system/archetype/os is now a dict instead of a string

**Required for Aquilon support** (see https://github.com/quattor/template-library-core/pull/177).

Fixes #1281

### DIFF
--- a/ncm-cron/src/main/pan/components/cron/config.pan
+++ b/ncm-cron/src/main/pan/components/cron/config.pan
@@ -1,8 +1,8 @@
 ${componentconfig}
 
 'securitypath' ?= {
-    if (exists('/system/archetype/os') &&
-            value('/system/archetype/os') == 'solaris') {
+    if (exists('/system/archetype/os/name') &&
+        value('/system/archetype/os/name') == 'solaris') {
         '/etc/cron.d';
     } else {
         '/etc';


jouvin:
@jrha I agree that it would be good to define the allowed values for the OS name at least. Could you open an issue about this so that we don't forget it?

ned21:
Please can we revert this? It's going to break us. I didn't object to https://github.com/quattor/template-library-core/pull/142 at the time because I thought it was unused anywhere but it now seems that was an incorrect assumption. I thought our coding standard was to use $^O rather than /system/archetype/os but that seems not to be the case here.

stdweird:
@ned21 i'm not sure what you object against: the value (or rather the path where the value is taken from) or the fact that the behaviour of an os is set via the configuration. the code uses `$^O` in several places, so the selection of this path can (or should) be moved there (or maybe the default value, we can allow to override it using this path).

ned21:
Sorry I mis-read the code while trying to catch up on 50 GitHub mails and now realise this is template code where we must use the PAN path not the $^O perl-ism!

jouvin:
@ned21 I don't think that there is a major backward compatibility problem but I agree that until you move to the new schema, you may have to hack `ncm-cron` `config.pan` which is not ideal... I don't think it makes sense to reverse this one alone as it is just fixing the consistency with https://github.com/quattor/template-library-core/pull/142. If you really disagree with https://github.com/quattor/template-library-core/pull/142, I think you should open an issue in the `template-library-core` repo so that we rediscuss the change and agree on what we do. If we decide to revert https://github.com/quattor/template-library-core/pull/142, then we'll revert this one...

stdweird:
@ned21 @jouvin you can use the `components/cron/site-config` templates to tune component config to your site preference. 
you can also get rid of this code in the template and move the default to the perl code.

jrha:
It's a default. If the value doesn't make sense at a particular site, the value should be set explicitly.

Issue #1289, opened 2018-06-12
ncm-useraccess: provide a way to thoroughly disable users


wdpypere:
ncm-useraccess does not provide a way to disable/remove users. for example making sure the authorized keys file is empty, ...

also the documentation states:
```
LOCKING USER ACCOUNTS

When you lock user accounts, it may not be enough to just lock them with passwd -l. Depending on how you configured SSH, a locked user may still be able to log-in with his public key. You can run <ncm-ncd --unconfigure useraccess> to temporarily lock all accounts, before removing the user's entry from CDB. Or remove manually .k*login and .ssh/authorized_keys.
```

I do believe this is very bad practice and should be removed from the documentation.

ned21:
Which bit specifically do want to remove? The warning that `passwd -l` is insufficient seems useful?

jrha:
ncm-accounts does though, by setting the shell to nologin, but that only works for local users.

wdpypere:
@ned21 I was specifically looking at the ```ncm-ncd --unconfigure useraccess``` part to be removed. If you want I can leave the ```passwd -l``` part.

wdpypere:
@jrha true, but keys remain in the authorized_keys file for instance.

ned21:
> I was specifically looking at the ncm-ncd --unconfigure useraccess part to be removed. 

I agree that makes sense.

PR #1290, opened 2018-06-12
ncm-useraccess: remove bad practice example from documentation

wdpypere:
I believe that what the documentation states about disabling/locking users is bad practice. see #1289 .


wdpypere:
@ned21 restored a part of the doc as requested.

PR #1291, opened 2018-06-19
ncm-grub: remove password from grub.conf if disabled

patrickbms:

Removing a password previously set in grub.conf won't work.

CAF::FileEditor->remove_lines removes lines matching the first
parameter, but not the second parameter. The empty string ''
matches all lines, so it will never remove lines.

Replacing it with 'q(no good line)' as in the ncm-named/*/named.pm
will have the desired effect.

If someone was accidentally or deliberately relying on setting the password to 'disabled' in a template, NOT removing the password line from grub.conf, they may be affected by making it work as documented.

If they were using the "enabled=false" line as designed, it wouldn't have worked.

### DIFF
--- a/ncm-grub/src/main/perl/grub.pm
+++ b/ncm-grub/src/main/perl/grub.pm
@@ -165,7 +165,7 @@ sub password
         return SUCCESS;
     } elsif (!$tree->{enabled}) {
         $self->info("removing grub password");
-        $grub_fh->remove_lines(qr/^password\s+/, q(no good line));
+        $grub_fh->remove_lines(qr/^password\s+/, qr/$./);
         return SUCCESS;
     }
 
--- a/ncm-grub/src/main/perl/grub.pm
+++ b/ncm-grub/src/main/perl/grub.pm
@@ -165,7 +165,7 @@ sub password
         return SUCCESS;
     } elsif (!$tree->{enabled}) {
         $self->info("removing grub password");
-        $grub_fh->remove_lines(qr/^password\s+/, '');
+        $grub_fh->remove_lines(qr/^password\s+/, q(no good line));
         return SUCCESS;
     }
 
--- a/ncm-grub/src/test/perl/methods.t
+++ b/ncm-grub/src/test/perl/methods.t
@@ -3,6 +3,7 @@ use warnings;
 
 use Test::More;
 use Test::Quattor qw(password);
+use Test::Quattor qw(nopassword);
 use Test::Quattor::Object;
 use CAF::Object;
 use Test::MockModule;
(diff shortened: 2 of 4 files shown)


ncm-grub/src/test/resources/nopassword.pan
```diff
@@ -0,0 +1,13 @@
+object template nopassword;
+
+include 'base';
+
+
+prefix "/software/components/grub/password";
+"enabled" = false;
+"password" = "";
```
stdweird: is this intentionally empty? i wonder why we allow an empty string as a password

ncm-grub/src/main/perl/grub.pm
```diff
@@ -165,7 +165,7 @@ sub password
         return SUCCESS;
     } elsif (!$tree->{enabled}) {
         $self->info("removing grub password");
-        $grub_fh->remove_lines(qr/^password\s+/, '');
+        $grub_fh->remove_lines(qr/^password\s+/, q(no good line));
```
stdweird: something that will never match is better instead of the `no good liine`: use `qr/$./` (match a character after end of string)

ncm-grub/src/test/perl/methods.t
```diff
@@ -3,6 +3,7 @@ use warnings;
 
 use Test::More;
 use Test::Quattor qw(password);
+use Test::Quattor qw(nopassword);
```
stdweird: add the `nopassword ` after the `password in previous line

ncm-grub/src/test/resources/nopassword.pan
patrickbms: Mostly accidental? Just an artifact of testing. I was trying with `enabled=true && password=""` which I believe was accepted and gave an empty `password=` on the command line.

stdweird: but i guess empty `password=` makes no sense?
can you tighten the schema https://github.com/quattor/configuration-modules-core/blob/master/ncm-grub/src/main/pan/components/grub/schema.pan#L25, use `string(1..)` as type

jrha:
@hpcugentbot please test this

stdweird:
test this please

stdweird:
@ned21 @gombasg @urbonegi i guess @patrickbms is from MS? if so, i can add him to the quattor group so his tests are triggered automatically

ned21:
@stdweird yes please

patrickbms:
I'm not sure about what travis is complaining about. The nopassword.pan is basically a copy of password.pan, and the travis log isn't very helpful to narrow it down.

stdweird:
@patrickbms don't worry about the travis error. it's panlint that is complaining about code in the test
(it detects that the template is configuring something in a component, but that the components config.pan isn't included in the same template.)

patrickbms:
@stdweird Thanks, it was a bit confusing. :)

PR #1297, opened 2018-06-27
ncm-authconfig: active directory support

ajf8:

To support using sssd with our multiple active directory realms, various new configuration attributes needed to be supported.

Nothing backwards incompatible, just adding new options.


jrha:
Thanks!

PR #1299, opened 2018-07-03
ncm-sysconfig: Automatically quote values containing whitespace

### LINKED ISSUE #1191: ncm-sysconfig: Values should be quoted
Sysconfig files are normally intended to be sourced from shell scripts, so values on the right hand side should be quoted to prevent globbing and word splitting.

This is particually important when values contain spaces, at present we have to have code like the below with quotes in the string which is rather fragile.
```pan
prefix '/software/components/sysconfig/files/xrootd';
'XROOTD_DEFAULT_OPTIONS' = '"-l /var/log/xrootd/xrootd.log -c /etc/xrd.cf -k fifo"';
```

jrha:
Fixes #1191.

If the user has already included quotes in the value, this is detected and worked around.

### DIFF
--- a/ncm-sysconfig/src/main/perl/sysconfig.pm
+++ b/ncm-sysconfig/src/main/perl/sysconfig.pm
@@ -20,7 +20,6 @@ our $NoActionSupported = 1;
 
 use Readonly;
 
-Readonly my $QUOTE => "\"";
 Readonly my $SYSCONFIGDIR => "/etc/sysconfig"; # The base directory for sysconfig files.
 
 sub filelist_read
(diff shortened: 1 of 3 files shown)


stdweird:
@jrha perhaps rename to automatic quoting of value containing whitespace

jrha:
I've just discovered a rare case where a bash array is being passed as a value, quoting would break this so I should probably update this PR to handle that.

jrha:
Re-based and added exception for values that look like arrays

ncm-sysconfig/src/main/perl/sysconfig.pm
```diff
@@ -87,7 +86,21 @@ sub Configure
...
                 if ($key ne 'prologue' && $key ne 'epilogue') {
-                    $contents .= "$key=$pairs->{$key}\n";
+                    my $value = $pairs->{$key};
+                    # Remove any leading or trailing whitespace from the value
+                    $value =~ s/^\s+|\s+$//g;
+                    # Quote values containing whitespace.
+                    # Only if they have not already been quoted and do not look like an array.
+                    # If a value contains quotes, use a different quotation mark.
+                    # Values without whitespace should not be modified.
+                    if ($value !~ m/^".*"$/ && $value !~ m/^'.*'$/ && $value !~ m/^\(.*\)$/ && $value =~ m/\s/) {
```
stdweird: i'm not sure what tis does "for array", but you should only do this on scalars (and probabnly report an error for anything else:
```perl
my $ref = ref($value);
if ($ref) {
    error("something something should not pass non-scalar to sysconfig: got type $ref")
} else {
   # do magic wrt quotes and whitespace
}
```

stdweird: nvm, i was totally confused

PR #1300, opened 2018-07-04
ncm-metaconfig: nginx: support return

stdweird:
To be used for eg redirects (http to https; root to default path)


jrha:
@hpcugentbot please test this please

PR #1301, opened 2018-07-05
ncm-metaconfig: httpd: add missing wsgi and some other configuration options

stdweird:
Adding OIDC to openstack keystone implies generating the whole config.

Requires panc >= 10.7 (uses equals in annotation).


ncm-metaconfig/src/main/metaconfig/httpd/pan/schema.pan
```diff
@@ -386,15 +387,33 @@ type httpd_perl_handler = {
...
-type httpd_wsgi_iportscript = {
+type httpd_wsgi_importscript = {
     "path" : string
     "process" ? string
     "application" ? string
 };
 
+type httpd_wsgi_daemonprocess = {
+    "name" : string
+    @{converted in list of key=value}
```
jrha: panc on Jenkins is falling over the equals in this annotation, time to update to 10.7?

stdweird: sure, but then there's no way back.you can move the milestone to a release where we can expect that 10.7 is ok to run. this is a parse error, not a compilation error, so afaik, as soon as this temnplate in your repo, you must be running 10.7

jrha: Hmm, could you remove the `=` then?

stdweird:
retest this please

PR #1304, opened 2018-07-13
ncm-grub: escape interpretation of value of md5 password as a regex and add sensitive flag to file editor.

patrickbms:
ncm-grub: escape interpretation of value of md5 password as a regex and add sensitive flag to file editor.

A regular expresion 'qr/^password $val$/' results in the interpretation of the contents of $val as a regular expression if it contains regex characters, which md5 passwords do. This properly escapes it. There should be no backwards incompatibility unless someone was relying on this bug, but I can't think of how that could be used.

We should not show diffs of grub.conf changes since it will expose encrypted password+salt in the logs, so I've added the sensitive flag. If someone was relying on the diff of the grub.conf being available in the logs, or via --noaction, there may be a regression, but this change is necessary so there's no unintended leaks of salt+hash.

### DIFF
--- a/ncm-grub/src/main/perl/grub.pm
+++ b/ncm-grub/src/main/perl/grub.pm
@@ -200,7 +200,8 @@ sub password
             owner => "root",
             group => "root",
             mode  => oct(400),
-            log   => $self );
+            log   => $self,
+            sensitive => 1 );
 
         $usercfg_fh->add_or_replace_lines(qr/^GRUB2_PASSWORD=/, qr/^GRUB2_PASSWORD=\Q$password\E$/, "GRUB2_PASSWORD=$password\n", SEEK_END);
         $usercfg_fh->close();
--- a/ncm-grub/src/main/perl/grub.pm
+++ b/ncm-grub/src/main/perl/grub.pm
@@ -202,14 +202,14 @@ sub password
             mode  => oct(400),
             log   => $self );
 
-        $usercfg_fh->add_or_replace_lines(qr/^GRUB2_PASSWORD=/, qr/^GRUB2_PASSWORD=$password$/, "GRUB2_PASSWORD=$password\n", SEEK_END);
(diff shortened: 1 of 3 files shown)


ncm-grub/src/test/resources/md5password.pan
```diff
@@ -0,0 +1,15 @@
+object template md5password;
+
+include 'base';
+
+
```
ned21: include "components/grub/config";
here would make the travis panlint check pass.

although it's not actually required by pan due to its inclusion via base. But I guess we live by the rules of panlint @stdweird @jrha ?

stdweird: i will happily merge this without extra include. panlint on tests is interpreted a bit less strict also by @jrha.

ncm-grub/src/main/perl/grub.pm
```diff
@@ -200,16 +200,17 @@ sub password
             owner => "root",
             group => "root",
             mode  => oct(400),
-            log   => $self );
+            log   => $self,
+            sensitive => 1 );
```
stdweird: remove the whitespce.
even better, add a `,` and move the `);` to the next line (so next option will have clean diff)

stdweird:
@patrickbms just a tiny cosmetic remark
@ned21 feel free to merge if you don't want to wait for it

stdweird:
also @patrickbms thanks for spotting this. (but i have the impression only MS uses grub passwords :wink: )

patrickbms:
@stdweird No problem. My boss found the bug, I've just got the practice building/testing/pushing quattor components. :)

I also made the travis/cosmetic changes. Thanks.

stdweird:
retest this please

stdweird:
retest this please

patrickbms:
I've re-tested this locally without any issues; is there a button I'm missing to force a jenkins rebuild?

jrha:
@patrickbms you're not missing anything, the magic phrase above triggers them!

stdweird:
retest this please

stdweird:
retest this please

PR #1306, opened 2018-07-24
ncm-network: allow default broadcast

### LINKED ISSUE #1305: ncm-network 18.6.0: why requiring a brodcast address when using a static bootproto
With the new version of `ncm-network` (was previously running 17.7), `ncm-network` raises an error if there is no broadcast address defined for an interface using a static bootproto. If I'm right it used to be a warning and was working perfectly as the broadcast was computing by ifcfg from the network mask. Any good reason for this change?

stdweird:
Improves upon fix to #1305 
Also provides a path without network restart to remove any configured broadcast that are equal to the computed default (via `ipcalc --broadcast ipaddr netmask`)

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -1544,6 +1544,47 @@ sub legacy_keeps_state
     };
 }
 
+
+# check if file change was due to removal of explicit broadcast in favour of computed default
+#   (broadcast configuration has a default via "ipcalc --broadcast ipaddr mask")
+sub default_broadcast_keeps_state
+{
+    my ($self, $file_name, $name, $data, $exifiles, $alias) = @_;
+
+    # compute broadcast via ipcalc, remove BROADCAST=
+    my $proc = CAF::Process->new(
+        ["ipcalc", "--broadcast", $data->{ip}, $data->{netmask}],
+        log => $self,
+        keeps_state => 1
+        );
+    my $output = $proc->output();
+    my $exitcode = $?;
+    my $broadcast;
+    if ($exitcode == 0) {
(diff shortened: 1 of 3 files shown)


stdweird:
@jouvin should do the trick

jouvin:
@stdweird thanks, works like a charm!

jrha:
How serious is this? Because it's quite a large block of code to throw in this late in the release cycle, without it I would happily release final this afternoon, with it we'll need another RC and time for testing.

jrha:
@stdweird thanks, that seems more managable (if @jouvin is happy!)

ncm-network/src/main/perl/network.pm
```diff
@@ -1544,6 +1544,47 @@ sub legacy_keeps_state
...
+
+# check if file change was due to removal of explicit broadcast in favour of computed default
+#   (broadcast configuration has a default via "ipcalc --broadcast ipaddr mask")
+sub default_broadcast_keeps_state
+{
+    my ($self, $file_name, $name, $data, $exifiles, $alias) = @_;
+
+    # compute broadcast via ipcalc, remove BROADCAST=
+    my $proc = CAF::Process->new(
+        ["ipcalc", "--broadcast", $data->{ip}, $data->{netmask}],
```
jrha: I presume it is reasonable to assume that `ipcalc` is installed?

stdweird: it is the code run by initscripts, so i'd say yes

jrha: Ok, it's provided by the initscripts package in RH-land.
It is not installed by default on debian (if we care i) - it is however available in the ipcalc package.

PR #1307, opened 2018-07-25
ncm-network: broadcast is optional in schema, so do not report an error

### LINKED ISSUE #1305: ncm-network 18.6.0: why requiring a brodcast address when using a static bootproto

stdweird:
Fixes #1305


jouvin:
Sorry I was busy this morning but the code seems ok to restore a behaviour similar to the 17.7 one.

PR #1309, opened 2018-07-27
ncm-spma: Add new SPMA module using DNF (Fedora)/YUM4 (RHEL7/CentOS)…

jnovy:
This change adds a new SPMA module allowing to use DNF package manager which uses SAT solver for dependency resolution. This package manager is many folds faster than YUM and finds solution every time once it exists. The code is based on initial YUM-ng SPMA module I implemented earlier.

This module is not supposed to be 100% compatible with YUM based package manager as DNF itself is different than YUM. This module is not supposed to be only working with native DNF based OS-es (Fedora) but also on RHEL7 given the YUM4 repository from CentOS is provided.


jrha:
@hpcugentbot please test this

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -0,0 +1,22 @@
...
+
+type component_spma_dnf = {
+    include structure_component
+    include component_spma_common
+    "excludes"	    ? string[] # packages to be excluded from metadata
+    "dnfconf"	    ? string # /etc/dnf.conf DNF configuration
+    "whitelist"	    ? string[] # packages not shipped by repositories but generated by 3rd party installer
+    "quattor_os_file" ? string # file to write quattor_os_release as confirmation of successful YUM spma pass
+    "quattor_os_release" ? string # string to write to quattor_os_file
+    "proxy"         ? legacy_binary_affirmation_string # Enable proxy
```
jrha: As this is a new back-end, why not use boolean from the get-go?

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -0,0 +1,22 @@
...
+    "proxyhost"     ? string # comma-separated list of proxy hosts
+    "proxyport"     ? string # proxy port number
+    "proxyrandom"   ? legacy_binary_affirmation_string # randomize proxyhost
```
jrha: Ditto.

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -0,0 +1,22 @@
...
+    "proxytype"     ? string with match (SELF, '^(forward|reverse)$') # select proxy type, forward or reverse
+    "run"           ? legacy_binary_affirmation_string # Run the SPMA after configuring it
```
jrha: Ditto.

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -0,0 +1,22 @@
...
+    "verbose"       ? string with match (SELF, '^(0|1)$') # verbose (0,1)
```
jrha: This could probably be a boolean too.

ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -0,0 +1,598 @@
...
+        }
+    }
+
+    # Continue only if package content is supposed to be changed
+    return 1 unless $t->{run};
+
+    # Run test transaction to get complete list of packages to be present on the system
+    $self->execute_command( [ "rm -rf " . DNF_TEST_CHROOT ], "cleaning DNF test chroot", 1 );
+    $self->execute_command( [ "mkdir -p " . DNF_TEST_CHROOT . "/var/cache" ],                 "setting up DNF test chroot",    1 );
+    $self->execute_command( [ "ln -s /var/cache/dnf " . DNF_TEST_CHROOT . "/var/cache/dnf" ], "setting DNF test chroot cache", 1 );
```
jrha: These commands feel like they should be defined `Readonly` at the top of the module.

jrha:
@hpcugentbot please test this please

stdweird:
@jnovy when do plan to start using this in production? i'd like to cleanup some of the code so i can add some features from the regular yum plugin so we can also use it :wink:

ned21:
Superseded by #1404

PR #1312, opened 2018-08-07
ncm-accounts: Add a valid_user type to validate user names

XaF:
In some cases, we need to be able to check a username received as
parameter, this allows it by providing a new valid_user type and a
valid_user_name function that checks that the username received
as parameter is a valid one, i.e. comprised only of alphanumeric
characters, plus a few special ones: .+:-

This is required by some changes that we will introduce internally at Morgan Stanley, but as this code could be reused for other templates, we thought it best to submit it to the community.


XaF:
Will edit as part of the patch is now submitted to https://github.com/quattor/template-library-core/pull/184

XaF:
Not needed anymore here as now entirely included in #184

PR #1316, opened 2018-08-15
ncm-nscd: Add support for caching /etc/services and /etc/netgroup

jrha:
Support for these were added to nscd some time ago:
* services by the version shipped with EL6.
* netgroup by the version shipped with EL7.

The version of nscd is checked to determine support for each.

### DIFF
--- a/ncm-nscd/src/main/perl/nscd.pm
+++ b/ncm-nscd/src/main/perl/nscd.pm
@@ -5,12 +5,14 @@
 package NCM::Component::nscd;
 
 use strict;
+use version;
 use NCM::Component;
 use vars qw(@ISA $EC);
 use NCM::Check;
 use LC::Check;
 use File::Temp;
 use File::Copy qw(copy);
+use CAF::Process;
 
 @ISA = qw(NCM::Component);
 $EC = LC::Exception::Context->new->will_store_all;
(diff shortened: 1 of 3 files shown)


stdweird:
@jrha milestone?

ned21:
So it's incumbent on the user to not add the "netgroup" attribute on hosts that run <EL7 to avoid creating a config that is invalid on RH6? It would be nice if we could validate that in the schema but I don't think we have a consistent way of representing the OS version.

jrha:
Strictly it is dependent on the version of nscd, which is harder to check, but could be done at run-time.
I'm going to test an EL7 config with EL5 and EL6 versions to see how unhappy it gets.

jrha:
I've just checked and the version shipped with 6.9 includes `netgroup` in the config.

jrha:
Updated to check the version of nscd

ncm-nscd/src/main/perl/nscd.pm
```diff
@@ -91,6 +109,18 @@ sub Configure {
 
     # check per-service options we know about
     my @services = ('passwd', 'group', 'hosts');
+
+    my $nscd_version = $self->_get_nscd_version();
+
+    if ($nscd_version >= 2.12) {
```
ned21: Perl is magic 🤩

The method returns a string. >= is a numeric operator. I can believe Perl is clever enough to cast $nscd_version automatically, and this will work nscd gets to version 2.100 which is < 2.12. Seems safe enough, I think!

jrha:
@ned21 - Perl is indeed magic, but I've switched this to use `version` for more robust version checking.

ncm-nscd/src/main/perl/nscd.pm
```diff
@@ -19,6 +21,24 @@ $NCM::Component::NoActionSupported = 1;
 my $cfgfile='/etc/nscd.conf';
 my $p ='/software/components/nscd';
 
+
+sub _get_nscd_version {
+    my ($self) = @_;
+
+    my $proc = CAF::Process->new ([qw (nscd -V)], log => $self, keep_state => 1);
```
ned21: keep_state is not valid, it's `keeps_state` according to https://quattor-documentation.readthedocs.io/en/17.8.0/CAF/Process/

jrha: Thanks, stupid typo.

PR #1321, opened 2018-09-12
ncm-filecopy: Make restart failures non-fatal

gombasg:
Currently, if restarting a service fails, that marks the whole component
as failed, preventing other components which depend on filecopy from
running. In practice, that usually causes more damage than the restart
failure, so lower the severity of restart failures to warnings.

I considered making the behavior configurable, but I'm not sure if that
would worth the effort.

Comments are welcome.

### DIFF
--- a/ncm-filecopy/src/main/perl/filecopy.pm
+++ b/ncm-filecopy/src/main/perl/filecopy.pm
@@ -149,7 +149,8 @@ sub Configure
                                     stderr => "stdout");
         $cmd->execute();
         if ( $? ) {
-            $self->error("Command failed. Command output: $cmd_output\n");
+            my $method = $confighash->{ignore_restart_failure} ? 'warn' : 'error';
+            $self->{$method}("Command failed. Command output: $cmd_output\n");
         } else {
             $self->debug(1,"Command output: $cmd_output\n");
         }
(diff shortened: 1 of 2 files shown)


stdweird:
@gombasg i'm not in favour, i'd propose to make it configurable (a component setting is fine for me); if you need a name, i propose `    'dontcare' ? boolean` :wink: 
if you need it on a file basis, one can always suffix the command with `;/bin/true`

gombasg:
I've made the behaviour configurable.

ncm-filecopy/src/main/perl/filecopy.pm
```diff
@@ -149,7 +149,12 @@ sub Configure
                                     stderr => "stdout");
         $cmd->execute();
         if ( $? ) {
-            $self->error("Command failed. Command output: $cmd_output\n");
+            my $message = "Command failed. Command output: $cmd_output\n";
```
stdweird: more perlish (and it's `warn`, not `warning`; if this annoys you, open an issue in CAF)
```perl
my $method = $confighash->{ignore_restart_failure} ? 'warn' ? 'error'
$self->$method("Command failed. Command output: $cmd_output\n");
```

PR #1324, opened 2018-09-27
ncm-postgresql: Relax regex for 10.x versions of PostgreSQL

jrha:
Versios of postmaster before the 10 series returned a 3 digit version number, the 10 series only returns two digits. Adapt the regex to handle this and add a test case to verify that it works correctly.


jrha:
FYI we are running the Jenkins built snapshot RPM of this PR happily with PostgreSQL 10.5.

PR #1329, opened 2018-10-05
metaconfig: Add script for running tests

jrha:
Split out from #1256.


stdweird:
these are bash functions i use to run the tests, but hey are a bit outdated and you should use https://github.com/quattor/maven-tools/pull/175
you should be able to use mavenprove for these as well, but i haven't tried it to make it work (yet)

jrha:
I didn't think that #1328 used this script, which is why I split it out into a separate PR.

jrha:
I don't think this is relevant any more, please re-open with a justification if you disagree.

PR #1330, opened 2018-10-09
ncm-gpfs: rpms nfs-ganesha renamed

kwaegema:
for gpfs 5, rpms of nfs-ganesha from CES are renamed, so should be recognized by component


stdweird:
@kwaegema maybe already add the boolean to skip the whole rpm handling?

PR #1331, opened 2018-10-24
ncm-cron: Throw an error if multiple cron jobs are defined with the same name

### LINKED ISSUE #1228: ncm-cron: Catch cron entries with duplicate names at build time
Currently, multiple entries can be added to the cron entries list with the same name, in this case both entries are written to disk and the most recent (highest numbered) entry in the list wins.

A very long term solution might be to transition from a `list` of cron entries to a `dict` with `name` as the key, a more pragmatic solution would be to validate the names for uniqueness in the schema.

jrha:
Otherwise this causes the last processed to take precedence.

Resolves #1228.

### DIFF
--- a/ncm-cron/src/main/pan/components/cron/schema.pan
+++ b/ncm-cron/src/main/pan/components/cron/schema.pan
@@ -184,9 +184,31 @@ type structure_cron = {
     true;
 };
 
+function no_duplicate_cron_entries = {
+    if (ARGC != 1) {
+        error(format('%s: expected 1 parameter, received %d', FUNCTION, ARGC));
+    };
+
+    names = dict();
+    foreach(k; v; ARGV[0]) {
+        name = v['name'];
+        if (is_defined(names[name])) {
+            error(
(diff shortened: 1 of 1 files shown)


ned21:
Adding backwards incompatible label as this will break any existing code base where this feature is used which might hamper adoption of the new version. I know at previous workshops that @jouvin described how the use of list and "last entry wins" was a deliberate design decision so I think a bit more discussion is warranted.

jrha:
My recollection was that it was a design mistake, but discussion is welcome.

jrha:
Also: could easily allow this check to be disabled if needed.

jrha:
Discussed at workshop, agreed to define the validation function but leave unbound and leave the decision to bind in a `site_local` template to be site policy.

jrha:
Re-based and updated to simply provide a validation function as discussed.
Removed "backwards incompatible" label as the function is not applied by default.

jrha:
Fair question, I'd prefer a hard error, can easily make it so.

PR #1332, opened 2018-10-24
ncm-systemd: handle systemd escaping bug

stdweird:
This is needed if you need to configure units with systemd-escape'd values, which seems to be buggy on certain EL7 version (e.g. on centos 7.5)

### DIFF
--- a/ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
+++ b/ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
@@ -676,6 +676,35 @@ sub get_type_shortname
 }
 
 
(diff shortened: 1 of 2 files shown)


gombasg:
Has this been officially acknowledged as a bug in systemd somewhere, which could be referenced here?

stdweird:
@gombasg i opened a case with redhat (nr 02238014; but i don't think the cases are public), but this was not present in earlier centos7 versions, and is also not present in fedora. i guess redhat backported something and forgot a few commits

gombasg:
@stdweird I think it would be good to mention the case number in the comments - even if it is not public, others (us included) can still reference it when talking with RH, should that be needed.

stdweird:
i added the BZ1643172 that was created after our support case

PR #1334, opened 2018-10-25
ncm-opennebula: fix cluster host add

alvarosimon:
This is an opennebula issue. 
Hyps should not be removed from any cluster (if not they are included into the default cluster again), just add the required one.


alvarosimon:
@jrha @stdweird I have included just one more small fix in this PR, `ip6tables`   is required by the new ONE versions

Issue #1071, opened 2017-03-13
ncm-network: support daemons to restart after network changes


stdweird:
eg. shorewall doesn't like to restart network interfaces. 
propose to add a `daemons` dict like eg ncm-metaconfig

jrha:
Couldn't this be solved by expressing dependencies with systemd?

stdweird:
if we assume all systems have systemd, sure. and then we can also do this for eg metaconfig etc ...

PR #1335, opened 2018-10-25
ncm-metaconfig: Remove caf_service_action type to allow reuse

jrha:
Requires quattor/template-library-core#191 which moves the type into the core template library.

Enabling work for #1071.

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -84,8 +84,6 @@ type ${project.artifactId}_textrender_convert = {
     true;
 };
 
-type caf_service_action = string with match(SELF, '^(restart|condrestart|reload|stop_sleep_start)$');
-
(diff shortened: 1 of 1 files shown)


stdweird:
@jrha @ned21 bit of a mess right now

- https://github.com/quattor/configuration-modules-core/pull/1341 was merged, but no changes to schema

- https://github.com/quattor/template-library-core/pull/201 was moved, but https://github.com/quattor/template-library-core/pull/191 is still open

i propose to keep it simple: close https://github.com/quattor/template-library-core/pull/191, remove any usage of `caf_service_action` (i.e. keep the current commit) and also switch to `caf_serviceaction` here.

ned21:
> i propose to keep it simple: close [quattor/template-library-core#191](https://github.com/quattor/template-library-core/pull/191), remove any usage of `caf_service_action` (i.e. keep the current commit) and also switch to `caf_serviceaction` here.

Is that fully backwards compatible?

PR #1336, opened 2018-10-25
ncm-network: Add support for CAF::Service actions after changes

### LINKED ISSUE #1071: ncm-network: support daemons to restart after network changes

jrha:
Largely transcribed from ncm-metaconfig.

Requires #1335.
Resolves #1071.

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -194,6 +194,7 @@ Readonly my $NEW => 2;
 # changes to file, but same config (eg for new file formats)
 Readonly my $KEEPS_STATE => 3;
 
+Readonly::Hash my %ALLOWED_ACTIONS => { restart => 1, reload => 1, stop_sleep_start => 1 };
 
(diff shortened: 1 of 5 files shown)


ncm-network/src/main/perl/network.pm
```diff
@@ -1694,6 +1695,36 @@ sub down_rename_devices
     return $action;
 }
 
+# Prepare and take the action for all daemons as defined in hash-reference C<$daemons>.
```
stdweird: i would like to see this moved (and perhaps generalised a bit) to `CAF::Service`. then we can use it in even more components. replace `$self` with `$logger` and export the function there

jrha: Me too, are you volunteering? :smile_cat:

stdweird:
Try to use https://github.com/quattor/CAF/pull/267

PR #1348, opened 2018-11-06
ncm-metaconfig: add mysql conf

alvarosimon:
I have realised that we are using mysql/mariadb metaconfig setups but our current schema is not available from upstream. We need to change the mysql conf to time to time for our cloud clusters for instance. 

Maybe this could be useful for someone else.


jrha (dismissed):

jrha:
rebased

jrha (approved):

PR #1351, opened 2018-11-13
ncm-gpfs: needs boolean value instead of string

kwaegema:


ncm-gpfs/src/main/perl/gpfs.pm
```diff
@@ -66,7 +66,9 @@ sub Configure
     # how to retrigger spma afterwards?
     my $baseinstalled = GPFSCONFIGDIR . "/.quattorbaseinstalled";
     my $basiccfg = GPFSCONFIGDIR . "/.quattorbasiccfg";
-    my $skiprpm = $config->getValue($self->prefix."/skiprpm");
+    my $tree = $config->getTree($self->prefix);
```
stdweird: tree is not used anywhere, maybe do
```
my $skiprpm = $config->getTree($self->prefix)->{skiprpm};
```

kwaegema:
@jrha done

PR #1355, opened 2018-12-03
ncm-ceph: test blkid of device before trying to deploy

kwaegema:
adds an extra check before trying to run ceph-volume create

### DIFF
--- a/ncm-ceph/src/main/perl/Ceph/OSDserver.pm
+++ b/ncm-ceph/src/main/perl/Ceph/OSDserver.pm
@@ -128,8 +128,15 @@ sub deploy_osd
 
     # ceph-volume lvm create --data /dev/sdn --bluestore (--dmcrypt --crush-device-class baa)
     my $devpath = "/dev/" . unescape($name);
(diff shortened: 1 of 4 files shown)


ncm-ceph/src/main/perl/Ceph/OSDserver.pm
```diff
@@ -128,8 +128,15 @@ sub deploy_osd
...
     my $devpath = "/dev/" . unescape($name);
-    my $success = $self->run_command([qw(ceph-volume lvm create), "--data", $devpath, @options],
+    my $devexist = $self->run_command([qw(blkid -p), $devpath], "test $devpath blkid", test => 1);
+    my $success = 0;
+
+    if ($devexist) {
+        $self->warn("Device $devpath is not empty, not trying to deploy osd");
+    } else {
+        $success = $self->run_command([qw(ceph-volume lvm create), "--data", $devpath, @options],
         "deploy osd $devpath");
```
stdweird: indent

PR #1357, opened 2018-12-05
ncm-openstack: documentation fix code block not starting with newline

wdpypere:
and also fix a small typo.


ncm-openstack/src/main/perl/OpenStack/Service.pm
```diff
@@ -266,8 +266,9 @@ sub _set_elpath
 =item _attrs
 
 Add/set/modify more attributes
-Conviennce method for inheritance
+Convience method for inheritance
```
jrha: Nice try, but it should probably be "Convenience"

wdpypere:
@jrha fixed. Typing is difficult!

PR #1322, opened 2018-09-14
ncm-network: support rules and routes using routing tables

stdweird:

PR #1372, opened 2019-02-20
ncm-network: support ethtool channels

stdweird:
Requires #1322


jrha:
Looks interesting. Can you explain why someone would want to use this?

stdweird:
we have some broadcom adapters that have very low number of channels enabled in firmware or driver, and the only way to get it higher is via ethtool  (but it is documneted, so i guess it's a feature)

jrha:
Heh, eww.

Issue #1371, opened 2019-02-18
ncm-opennebula: Add VMs virtio-net queues support


alvarosimon:
This is an useful option to increase virtio performance. VM templates should include an optional value:

`VIRTIO_QUEUES`

More info:
https://docs.opennebula.org/5.6/deployment/open_cloud_host_setup/kvm_driver.html#usage

PR #1373, opened 2019-02-21
ncm-opennebula: add virtio queues support

alvarosimon:
Include virtio network queues support in libvirt

Fix: #1371 
See also: https://docs.opennebula.org/5.6/deployment/open_cloud_host_setup/kvm_driver.html


ncm-opennebula/src/main/resources/network_level1.tt
```diff
@@ -19,3 +19,6 @@ IP = "[% interfaces.${interface}.ip %]",
 [%     IF !ignoremac  -%]
 MAC = "[% hardware.cards.nic.${data.key}.hwaddr %]",
 [%     END -%]
+[%     IF system.opennebula.virtio_queues -%]
+VIRTIO_QUEUES = "[% system.opennebula.virtio_queues %]",
```
jrha: Just checking that this really should be quoted? It looks a little odd as it is a number.

PR #1376, opened 2019-03-21
ncm-systemd: Perform actions one unit at a time

gombasg:
If performing an action fails for one unit (e.g., because one tries to
enable a service which is not installed), then systemctl usually returns
a non-descriptive error like "No such file or directory", without any
indications for which unit caused the failure. What's worse, the rest of
the units will not be acted on either, often resulting in much more
breakage than the problematic unit would warrant.

To avoid this, perform actions one unit at a time.

### DIFF
--- a/ncm-systemd/src/main/perl/Systemd/Systemctl.pm
+++ b/ncm-systemd/src/main/perl/Systemd/Systemctl.pm
@@ -311,27 +311,32 @@ Returns exitcode and output.
 sub systemctl_command_units
 {
     my ($logger, $command, @units) = @_;
+    my ($success, $failure);
+
+    $success = $failure = 0;
 
     # TODO: any relevant options?
-    my $proc = CAF::Process->new(
-        [$SYSTEMCTL, $command],
-        log => $logger,
-        );
+    foreach my $unit (@units) {
+        my $proc = CAF::Process->new(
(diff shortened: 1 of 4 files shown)


ncm-systemd/src/main/perl/Systemd/Systemctl.pm
```diff
@@ -311,27 +311,32 @@ Returns exitcode and output.
...
-    # TODO: any relevant options?
-    my $proc = CAF::Process->new(
-        [$SYSTEMCTL, $command],
-        log => $logger,
-        );
+    $success = $failure = 0;
 
-    if (@units) {
-        $proc->pushargs('--', @units);
+    # TODO: any relevant options?
```
stdweird: is there any reference where this behaviour is defined? also, CAF::Service works like this, so might need changes as well

gombasg: No, `man systemctl` is quiet about what happens if you specify multiple units and one or more causes a failure. The biggest problem is the lack of indication _which_ unit caused the failure. So yes, I think CAF::Service should also handle one unit at a time.

jrha:
@stdweird?

stdweird:
test this please

PR #1377, opened 2019-04-11
ncm-metaconfig: ssh daemon configuration support

gombasg:
This allows deprecating ncm-ssh. I'd appreciate if someone cross-checked the schema with the sshd_options man page.


ned21:
@hpcugentbot test this please

jrha:
Looks nice, ncm-ssh does run-time validation of the generated configuration, is this now considered unnecessary?

gombasg:
@jrha: It would still be nice to have the run-time check, because the templates cannot verify if the generated configuration is compatible with the version of sshd which happens to be installed. I just don't know how that could be implemented with metaconfig. Maybe @stdweird has ideas :-)

ncm-metaconfig/src/main/metaconfig/ssh/tests/profiles/server_allopts.pan
```diff
@@ -0,0 +1,96 @@
+object template server_allopts;
+
+include 'metaconfig/ssh/server_config';
+
+prefix "/software/components/metaconfig/services/{/etc/ssh/sshd_config}/contents";
+
+"main/AcceptEnv" = list("LC_CTYPE", "LANG", "TERM");
```
stdweird: use relative `main` prefix (or add main to the absolute prefix path)

ncm-metaconfig/src/main/metaconfig/ssh/tests/profiles/server_allopts.pan
```diff
@@ -0,0 +1,96 @@
+object template server_allopts;
+
+include 'metaconfig/ssh/server_config';
```
stdweird: can you add a comment on what ssh release this is based? or which OS?

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
```diff
@@ -0,0 +1,44 @@
+[% spacelist = ['AcceptEnv', 'AllowGroups', 'AllowUsers', 'AuthenticationMethods', 'AuthorizedKeysFile', 'AuthorizedPrincipalsFile',
+    'DenyGroups', 'DenyUsers',
+    'IPQoS',
+    'PermitListen', 'PermitOpen',
+    'RekeyLimit' ] -%]
+[% commalist = ['Ciphers', 'HostKeyAlgorithms', 'HostbasedAcceptedKeyTypes', 'KexAlgorithms', 'MACs', 'PubkeyAcceptedKeyTypes' ] -%]
+[% multilinelist = ['HostKey', 'ListenAddress', 'Port' ] -%]
+[% booleans = ['AllowAgentForwarding',
```
stdweird: these are not needed if they are true pan booleans. you can test those explicitly in TT using `CCM.is_boolean()` (but not in a case block

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
```diff
@@ -0,0 +1,44 @@
...
+    'IgnoreRhosts', 'IgnoreUserKnownHosts',
+    'KbdInteractiveAuthentication', 'KerberosAuthentication', 'KerberosGetAFSToken', 'KerberosOrLocalPasswd', 'KerberosTicketCleanup',
+    'PasswordAuthentication', 'PermitEmptyPasswords', 'PermitRootLogin', 'PermitTTY', 'PermitTunnel', 'PermitUserEnvironment',
+    'PermitUserRC', 'PrintLastLog', 'PrintMotd', 'PubkeyAuthentication',
+    'StreamLocalBindUnlink', 'StrictModes',
+    'TCPKeepAlive', 'UseDNS', 'UsePAM', 'X11Forwarding', 'X11UseLocalhost'] -%]
+
+[%- FOREACH pair IN data.pairs -%]
+[%     NEXT IF pair.key == 'matches' -%]
+[%     SWITCH pair.key -%]
```
stdweird: for booleans, you could do (i think ;)
```
[-% IF CCM.is_boolean(pair.value) -%]
[%             pair.key %] [% pair.value ? 'Yes' : 'No' %]
[% NEXT -%]
[-% END -%]
[% SWITCH %] ...
```
(to be tested)

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
```diff
@@ -0,0 +1,44 @@
...
+[%         CASE 'Subsystem' -%]
+[%             FOREACH item IN pair.value.pairs -%]
+Subsystem [% item.key %] [% item.value %]
+[%             END -%]
+[%         CASE 'SetEnv' -%]
+[%             FOREACH item IN pair.value.pairs -%]
+SetEnv [% item.key %]="[% item.value %]"
+[%             END -%]
+[%         CASE -%]
+[%             pair.key %] [% pair.value %]
```
stdweird: or do the boolean check here, and maybe add a the most common type pf list and have that here as default list format (saves you another static list)

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
```diff
@@ -0,0 +1,44 @@
+[% spacelist = ['AcceptEnv', 'AllowGroups', 'AllowUsers', 'AuthenticationMethods', 'AuthorizedKeysFile', 'AuthorizedPrincipalsFile',
```
stdweird: this seems the longest list; use this as the default list fallback

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
stdweird: it's `value.is_boolean`; see default case below

ncm-metaconfig/src/main/metaconfig/ssh/server_attrs.tt
stdweird:
```
[%             pair.key %] [% -%]
[-% IF pair.value.is_boolean -%]
[%- pair.value ? 'Yes' : 'No' -%]
[%- ELSIF CCM.is_list(pair.value) -%]
[%- pair.value.join(' ') -%]
[%- ELSE -%]
[%- pair.value -%]
[-% END -%]
```

ncm-metaconfig/src/main/metaconfig/ssh/tests/profiles/server_allopts.pan
stdweird: nvm, didn't see that this was a test.

ncm-metaconfig/src/main/metaconfig/ssh/tests/profiles/server_allopts.pan
stdweird: nvm, this is a test

stdweird:
@gombasg i have ideas :smile: 
i have an objection against implementing another "run arbitrary something as root" from the profiles, but hardcoding a whole bunch of test commands and being able to call them is fine for me.
and here it would be easy because the test does not require a file (or "the" final file).
anything that can be verified from stdin is easy to do.

verifying from file might be easy to do, if we patch/hack the atomic filewriter so we can test on the temp file. but it needs a patch to the atomicwriter.

jrha:
@gombasg if it was desirable then the cleanest (IMHO) way to implement it would be to make the existing component use textrender with these TT files.

jrha:
I'm bumping this - I want to look at migrating ncm-ssh to use these TT files with `textrender` so we can keep the validation logic.

gombasg:
@jrha The question is, where to put the schema? I don't want to duplicate the definitions which are common between the client and the server.

gombasg:
There is another issue with `ncm-ssh`. It is currently using `CAF::FileEditor``, and not `CAF::FileWriter`, which is a pain if one has to switch between different OpenSSH version supporting a different set of parameters, with the ability to back the change out if needed. However, making the switch in the `ncm-ssh` component would break deployments which do not define all SSH daemon options in profiles, but depend on system default remaining undisturbed. Using a different component means the template changes need updating anyway, so the missing options could be added.

ned21:
> There is another issue with ncm-ssh. It is currently using CAF::FileEditor, and not CAF::FileWriter, which is a pain if one has to switch between different OpenSSH version supporting a different set of parameters, with the ability to back the change out if needed. However, making the switch in the ncm-ssh component would break deployments which do not define all SSH daemon options in profiles, but depend on system default remaining undisturbed. Using a different component means the template changes need updating anyway, so the missing options could be added.

How about adding a "full control" option to ncm-ssh? That would allow us to keep the config validation check prior to restarting sshd? Or have I not understood the constraint with Atomic FileWriter?

gombasg:
A "full control" option need a new implementation of the Configure method. The schema used by ncm-ssh is incompatible with the metaconfig-based schema, so either the TT files will have to be re-written from scratch, or the entire schema need to be switched based on the value of "full control". So essentially, we'd end up with two independent, incompatible components disguised as one. I'm not sure keeping the validation would worth it. If validation is really needed, then we should convince @stdweird to add a generic validation step to metaconfig. Yes, that would in fact mean running arbitrary commands as root, because we do not always run the OS-supplied sshd, so the full path of the daemon would need to be specified in the templates, it cannot be hard-coded in the component.

stdweird:
@gombasg @ned21 @jrha @jouvin lets settle this: feel free to vote and/or campaign for

a: run arbitrary commands via ncm-metaconfig
b: make static list of allowed commands (or list of regexes or somesuch) for things that make sense (like the ssh config self test or others)
c: do nothing, if you want to run arbitrary code, make a service/unit and call it as part of the daemons (should fix the dependecy order to run ncm-systemd before ncm-metaconfig)

whatever is decided, i can then also implement it to run validation

gombasg:
Ask the question differently: do we want ncm-metaconfig to ever became a suitable replacement of ncm-filecopy or not? If you want to make ncm-metaconfig to replace ncm-filecopy, then you must allow running arbitrary commands, because not everything is a systemd service. E.g., if you want to disable the user list on the GDM login screen, then the command which needs to be run is “dconf update”.

Now listing all such commands in a distinct dictionary, outside of the configuration file contents, would make sense, as it would make extracting/auditing all such potential commands easier. Individual config files would then just reference a key inside that dictionary instead of spelling out the whole command. So if one uses structure templates to define the contents of a config file, then such a structure template would still be unable to cause arbitrary commands to run – it could only trigger commands which are defined elsewhere.

stdweird:
@gombasg thanks for voting! ;) i get that, i also struggle with that, as it's hard to try to limit the power of running arbitrary commands in a practical way.

i can add a pre, verifiy and post command option. verify will receive the file content on stdin (like ncm-ssh checks the validity).

in this case, i would then set a default value of the `verify` command to `list("/usr/sbin/sshd", '-t', '-f', '/dev/stdin')`. 
i would stick to the list format (no subshell), but not sure this is enough? should it be a list of lists to support running multiple commands? or just a string?

gombasg:
I’d not complicate it too much. A string should be fine. I did not write any substantial Perl code in a long time, but AFAIK some of the usual helper modules is smart enough to avoid going through the shell if the command line does not contain any shell metacharacters.

ned21:
Superseded by #1452

PR #1378, opened 2019-04-16
ncm-opennebula: fix add several ip/ars to same vnet

alvarosimon:
This is a bit silly issue (not sure why we didn't find this issue before...) 
It was not possible to add several ARs or IPs to the same virtual network from the AII when the VM is instantiated. This should fix that problem.


alvarosimon:
tested in altaria

jrha:
LGTM, lets see if we can poke Jenkins into life.
@hpcugentbot please test this.

PR #1379, opened 2019-05-06
ncm-ceph: add bitmap allocator to osd config

kwaegema:


jrha:
Is there any upstream documentation for these options? I couldn't find any :disappointed:

kwaegema:
I got this from a mailinglist thread, but indeed I don't find anything in the 'docs'..
This was probably still some 'experimental' feature in Luminous/Mimic that is now the default in Nautilus and shouldnt be set to something else:

[root@osd002 ~]# ceph config help bluestore_allocator
bluestore_allocator - Allocator policy
  (str, advanced)
  Default: bitmap
  Possible values:  bitmap stupid
  Can update at runtime: false

Allocator to use for bluestore.  Stupid should only be used for testing.
[root@osd002 ~]# ceph config help bluefs_allocator
bluefs_allocator - 
  (str, dev)
  Default: bitmap
  Can update at runtime: false

So we can probably get rid of configuring this completely on Nautilus

kwaegema:
This parameter never really made it in mainstream ceph code, closing it

PR #1380, opened 2019-05-09
ncm-freeipa: only request new certificate when local one is expired/will expire soon

stdweird:
Fixes a bug where each component run requests a new certificate.
Adds local certificate expiration validation check to determine if a new certificate needs to be retrieved


jrha:
@hpcugentbot test this please

PR #1382, opened 2019-05-29
ncm-metaconfig: add schema for elasticsearch >= 6.3.0

itkovian:
- Starting with 6.3.0, a change was made in the configuration settings for bulk requests
- This schema is not compatible with ES < 6.3.0


ncm-metaconfig/src/main/metaconfig/elasticsearch/pan/schema_6.0.pan
```diff
@@ -0,0 +1,58 @@
+declaration template metaconfig/elasticsearch/schema_6.3;
```
jrha: Should this end with `6.0` rather than `6.3`?

PR #1354, opened 2018-11-30
ncm-openstack: set nova cpu/mem overcommit and add new hyps

alvarosimon:
**NOTE:** this PR requires https://github.com/quattor/CCM/pull/193

* Fix: #1353

PR #1383, opened 2019-06-03
ncm-openstack:  add ceilometer service

### LINKED ISSUE #1381: ncm-openstack: add ceilometer service support
This will be useful to gather users metrics and provide cloud accounting.

alvarosimon:
**NOTE:** This PR requires #1354  to work, #1354 should be reviewed/merged first.

* Fix #1381


ncm-openstack/src/main/perl/OpenStack/Service.pm
```diff
@@ -534,18 +540,20 @@ Must return 1 on success.
 sub populate_service_database
 {
     my ($self) = @_;
+
+    # execute pre_populate commands before everything else
+    $self->pre_populate_service_database();
     # db_version is slow when not initialised
     # (lots of retries before it gives up; can take up to 90s)
-    if ($self->_do([$self->{manage}, @{$self->{db_version}}], 'determine database version', test => 1)) {
+    if ($self->_do(['/bin/bash', '-c', join(' ', $self->{manage}, @{$self->{db_version}})], 'determine database version', test => 1)) {
```
alvarosimon: @stdweird this is just a workaround to avoid the execution issue with `ceilometer-upgrade`. 
I have changed the unittests as well but now it is a bit messy, some commands are using `/bin/bash` workaround some of them  are not.
We should provide a better solution (maybe just adding a new flag into `CAF::Process` to use the "right" environment)

ncm-openstack/src/main/perl/OpenStack/Service.pm
```diff
@@ -534,18 +540,20 @@ Must return 1 on success.
...
         $self->verbose("Found existing db_version");
         return 1 if ($self->{flavour} eq "rabbitmq");
     } else {
         $self->verbose("Database not yet available for service $self->{flavour}");
     };
 
     # Always populate/sync the databases
-    $self->pre_populate_service_database();
-    if ($self->_do([$self->{manage}, @{$self->{db_sync}}], 'populate database')) {
+    if ($self->_do(['/bin/bash', '-c', join(' ', $self->{manage}, @{$self->{db_sync}})], 'populate database')) {
```
alvarosimon: same env issue with `ceilometer-upgrade` execution

apdibbo:
This looks good to me on the proviso that I havent really checked the perl

jrha:
Rebased on master to avoid conflicts now that #1354 has been merged.

ncm-openstack/src/main/perl/OpenStack/Service.pm
jrha: I've created a CAF issue for this, could you fill it in with more details of what would be required to avoid this workaround?
https://github.com/quattor/CAF/issues/272

PR #1387, opened 2019-06-13
ncm-metaconfig: the content should not only be a dict

itkovian:
- When producing JSON, a list should also be accepted as
  valid. Before the change, only JSON objects could be
  used as content.
- This should remain fully backwards compatible


ned21:
Change looks good but the explanation here in the body of the PR should really be in the commit log so it lives in git rather than some proprietary website DB owned by MSFT... :-) Can you amend the commit to include the full context/explanation please?

PR #1396, opened 2019-08-27
ncm-systemd: add extra unit options for sandboxing

wpoely86:
This add a couple of sandboxing options that can be used in systemd unit from https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Sandboxing

I've only added the ones I need, but the list of potential options is long. Should I add all of them?

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -145,6 +145,10 @@ http://www.freedesktop.org/software/systemd/man/systemd.exec.html
 valid for [Service], [Socket], [Mount], or [Swap] sections
 }
 type ${project.artifactId}_unitfile_config_systemd_exec = {
+    'CacheDirectoryMode' ? type_octal_mode
+    'CacheDirectory' ? string[]
+    'ConfigurationDirectoryMode' ? type_octal_mode
+    'ConfigurationDirectory' ? string[]
     'CPUAffinity' ? long[][] # start with empty list to reset
     'CPUSchedulingPolicy' ? string with match(SELF, '^(other|batch|idle|fifo|rr)$')
     'CPUSchedulingPriority' ? long(1..99) # 99 = highest
@@ -170,13 +174,20 @@ type ${project.artifactId}_unitfile_config_systemd_exec = {
     'LimitRTTIME' ? long(-1..) # Specifies a limit (in microseconds) on the amount of CPU time that a process scheduled under a real-time scheduling policy may consume without making a blocking system call.
     'LimitSIGPENDING' ? long(-1..) # Specifies the limit on the number of signals that may be queued for the real user ID of the calling process.
     'LimitSTACK' ? long(-1..) # The maximum size of the process stack, in bytes.
+    'LogsDirectoryMode' ? type_octal_mode
+    'LogsDirectory' ? string[]
     'Nice' ? long(-20..19)
     'OOMScoreAdjust' ? long(-1000..1000)
     'PrivateTmp' ? boolean
     'RootDirectory' ? string
+    'RuntimeDirectoryMode' ? type_octal_mode
+    'RuntimeDirectoryPreserve' ? choice('yes', 'no', 'restart')
+    'RuntimeDirectory' ? string[]
     'StandardError' ? ${project.artifactId}_unitfile_config_systemd_exec_stdouterr
     'StandardInput' ? string with match(SELF, '^(null|tty(-(force|fail))?|socket)$')
     'StandardOutput' ? ${project.artifactId}_unitfile_config_systemd_exec_stdouterr
+    'StateDirectoryMode' ? type_octal_mode
+    'StateDirectory' ? string[]
     'SupplementaryGroups' ? defined_group[]
     'SyslogFacility' ? syslog_facility
     'SyslogIdentifier' ? string
@@ -186,7 +197,7 @@ type ${project.artifactId}_unitfile_config_systemd_exec = {
     'TTYReset' ? boolean
     'TTYVHangup' ? boolean
     'TTYVTDisallocate' ? boolean
-    'UMask' ? string # octal notation, e.g. 0022
+    'UMask' ? type_octal_mode
     'User' ? defined_user
     'WorkingDirectory' ? string
 };
--- a/ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
+++ b/ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
@@ -53,6 +53,9 @@ bind "/unitfile" = systemd_unitfile_config[];
         'TTYVHangup', false,
         'LimitSTACK', -1,
         'LimitNPROC', 100,
+        'RuntimeDirectory', list('foo/bar', 'tmp'),
+        'RuntimeDirectoryMode', '0777',
+        'RuntimeDirectoryPreserve', 'restart',
         ),
     'install', dict(
         'WantedBy', list('1.service', '2.service'),
--- a/ncm-systemd/src/main/resources/regular/tests/regexps/unitfile0/value
+++ b/ncm-systemd/src/main/resources/regular/tests/regexps/unitfile0/value
@@ -19,6 +19,9 @@ contentspath=/unitfile/0
 ^ExecStart=/usr/bin/special$
 ^LimitNPROC=100$
 ^LimitSTACK=infinity$
+^RuntimeDirectory=foo/bar tmp$
+^RuntimeDirectoryMode=0777$
+^RuntimeDirectoryPreserve=restart$
 ^TTYReset=yes$
 ^TTYVHangup=no$
 ^\[Unit\]$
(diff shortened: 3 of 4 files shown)


ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -145,6 +145,10 @@ http://www.freedesktop.org/software/systemd/man/systemd.exec.html
 valid for [Service], [Socket], [Mount], or [Swap] sections
 }
 type ${project.artifactId}_unitfile_config_systemd_exec = {
+    'CacheDirectoryMode' ? string # octal notation
```
stdweird: either use a int and print it in octal, or make a new string type (if none exists already) with a regexp that locks this down

stdweird:
@wpoely86 no need to add them all, this is fine

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -145,6 +145,10 @@ http://www.freedesktop.org/software/systemd/man/systemd.exec.html
...
+    'CacheDirectory' ? string
```
ned21: These options take a whitespace-separated list of directory names. The specified directory names must be relative, and may not include "..".

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -170,13 +174,20 @@ type ${project.artifactId}_unitfile_config_systemd_exec = {
...
     'LimitSIGPENDING' ? long(-1..) # Specifies the limit on the number of signals that may be queued for the real user ID of the calling process.
     'LimitSTACK' ? long(-1..) # The maximum size of the process stack, in bytes.
+    'LogsDirectoryMode' ? string # octal notation
+    'LogsDirectory' ? string
     'Nice' ? long(-20..19)
     'OOMScoreAdjust' ? long(-1000..1000)
     'PrivateTmp' ? boolean
     'RootDirectory' ? string
+    'RuntimeDirectoryMode' ? string # octal notation
+    'RuntimeDirectoryPreserve' ? boolean
```
ned21: > Takes a boolean argument or restart. If set to no (the default),

So perhaps better as a string with validation? Otherwise you can't set it to the value of "restart" :(

jrha:
```suggestion
    'RuntimeDirectoryPreserve' ? choice('yes', 'no', 'restart')
```

stdweird:
@wpoely86 for the long, still need to adapt the TT to print is as octal

wpoely86:
On it. I'm trying to figure out how to run and adapted the tests. 

I should change the files under `src/main/resources/regular/tests/`?

Also still trying to figure out the TT magic.

ncm-systemd/src/main/resources/unitfile/section.tt
```diff
@@ -11,6 +11,6 @@
 [%-          CASE long_infinity -%]
 [%-              pair.key %]=[% value == -1 ? 'infinity' : value %]
 [%           CASE -%]
-[%-              pair.key %]=[% INCLUDE 'systemd/unitfile/attribute_value.tt' data=value %]
+[%-              pair.key %]=[% INCLUDE 'systemd/unitfile/attribute_value.tt' data=value octals=['CacheDirectoryMode', 'ConfigurationDirectoryMode', 'LogsDirectoryMode', 'RuntimeDirectoryMode', 'StateDirectoryMode'] %]
```
stdweird, line 14: split the line; if all keys ending with `Mode` are octals, best to match this in the CASE statements instead of maintaining the list

wpoely86: As far as I currently understand, you have to write it in base 10 and this will convert it to octal right?
So we have to write `mask = 511` to get `777` ? Isn't writing it as a string a bit more readable?

Dockerfile
```diff
@@ -31,12 +31,12 @@ RUN cpanm install Git::Repository Data::Structure::Util
 ENV QUATTOR_TEST_TEMPLATE_LIBRARY_CORE /quattor/template-library-core-master
 
 # set workdir to where we'll run the tests
-COPY --chown=99 . /quattor_test
```
stdweird, line 34: why did you change this?

ncm-systemd/src/main/resources/unitfile/section.tt
stdweird: what do you mean "write in base10"? in pan, you can also write `mask = 0777`

Dockerfile
wpoely86: That was not supposed to be committed. 

I need it to run the tests locally.

stdweird:
@wpoely86 ouch, forgot about on thing. please cleanup th ecommit history, and best to prefix the commit messages with `ncm-systemd:`

wpoely86:
@stdweird done

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -145,6 +145,10 @@ http://www.freedesktop.org/software/systemd/man/systemd.exec.html
 valid for [Service], [Socket], [Mount], or [Swap] sections
 }
 type ${project.artifactId}_unitfile_config_systemd_exec = {
+    'CacheDirectoryMode' ? long(1..1023) # octal notation
```
ned21, line 148: How do I specify `0755` ? Is the intention it should be written `755` ? Is that consistent with other components?

ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
```diff
@@ -52,6 +52,9 @@ bind "/unitfile" = systemd_unitfile_config[];
         'TTYVHangup', false,
         'LimitSTACK', -1,
         'LimitNPROC', 100,
+        'RuntimeDirectory', list('foo/bar', 'tmp'),
+        'RuntimeDirectoryMode', 0777,
```
ned21, line 56: Interesting - so 0777 is a valid long in PAN?!  Not sure that's obvious from someone looking at the schema.

ncm-systemd/src/main/pan/components/systemd/schema.pan
wpoely86: Why wouldn't you be able to write `0755`? 
I'm not that deep into PAN but 755 is decimal, 0755 is octal?

ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
wpoely86: I'm not sure I understand your remark about the schema?

I've tested it empirically and 0777 produces 777 in the output. If I remove the filter part, it becomes 511. So it seems like pan knows and reads 0777 as octal.

stdweird: @ned21 pan supports octal (any long that starts with an explicit `0`) and hex (same, but start with `0x`)
it's not obvious from the schema, since there's nothing to it. perhaps the docstring might use an example.

stdweird: even in docs https://quattor-pan.readthedocs.io/en/stable/pan-book/pan-book.html#properties-and-primitive-types

ned21: Yep, sorry I had a nagging doubt about this earlier. I remember there was a nasty regression in an one version of filecopy due to this not being handled well and I wanted a bit more time to think about it. 

I think there are two issues:
1. As a user should I specify 0777 or 777? Do they do the same thing? The answer is that they do mean different things in PAN (which can lead to subtle bugs) so it's down the implementor to make sure that the user intent is converted correctly.
2. 777 and 1777 are valid decimal numbers which are also valid modes when interpreted as a string. What happens to more esoteric modes like 011 and 0001? Again if the implementation is sufficiently robust then the right thing will happen.

So based on the unit test, it looks like the code is good. How do we communicate to a user that either form is acceptable? I don't think we need to spell it out in every line of the schema, but it might be useful to mention it somewhere, perhaps by way of example as was suggested.

wpoely86: I can add a docstring? The output is in octal. So you can write, 0777, 511 or 0x1FF to get 777. 
However, write 777 and you will get 1411. 

If you want both 777 and 0777 to produce the same thing, I need to use a string. 

What's the preferred way?

stdweird: there has to be a limit on the size to start `0..511` or `0..1023` if you want sticky bits.
but this indeed can be confusing for endusers. `0511` and `511` are valid and very different.
if you want to protect against it, you need to add compiler support for it. otherwise a string that parses as an octal is the only option.

ned21: > However, write 777 and you will get 1411.

yeah, so IMHO that's not user friendly. And if the number needs to always be written in octal then does 1777 actually need to be written as 01777? 

In which case I think I agree that using a string makes more sense here.

wpoely86: @ned21 I've pushed a commit that uses a string (with regex) for the modes/masks.

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -42,6 +42,9 @@ type ${project.artifactId}_unit_virtualization = string with match(SELF,
 #    make this more finegrained, e.g. has to be existing unit; or check types
 type ${project.artifactId}_valid_unit = string;
 
+# modes are in octal
+type ${project.artifactId}_systemd_mode = string with match(SELF, '^0?1?[0-7][0-7][0-7]');
```
ned21, line 46: Should this be made globally available via [pan/types.pan](https://github.com/quattor/template-library-core/blob/master/pan/types.pan) ?

wpoely86: I've opened a PR at quattor/template-library-core#200.

Once that one is merged, I can directly start using the type here or do I need to do anything else?

gombasg, line 46: Why not just `... = string with to_long(SELF, 8) >= 0 && to_long(SELF, 8) <= 07777`? That's what `systemd`'s parser is doing - there are no requirements on the number of digits.

wpoely86: even better, thanks for the hint.

I've pushed this to template-library-core#200 and the same change to this PR. As soon as the former is merged, I can switch the type here.

jrha:
rebased

jrha (approved):

ned21 (approved):

PR #1401, opened 2019-11-03
ncm-spma: yumdnf: initial dnf/yum4 based on yum

stdweird:
Initial DNF support in el8 (tested on centos 8.0) via yum4 backwards compatibility and yum backend.
Based on #1405


ncm-spma/src/main/perl/spma/yumdnf.pm
```diff
@@ -0,0 +1,28 @@
...
+
+
+# Completes any pending transactions
+sub _do_complete_transaction
+{
+    my ($self) = @_;
+
+    # Could be implmented using e.g. dnf remove $(dnf repoquery --duplicated --latest-limit=-1 -q)
+    #   not sure it does what you think it does
+    $self->verbose("Skipping complete_transaction, no DNF equivalent implemented");
```
ned21: This doesn't feel like a message for users but instead programmers so should be debug() not verbose()? Or even disabled all together?

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -293,17 +294,24 @@ sub execute_yum_command
 
     $opts{stdin} = $stdin if defined($stdin);
 
-    my $cmd = CAF::Process->new(_set_yum_config($command), %opts);
+    my $cmd = CAF::Process->new($self->_set_yum_config($command), %opts);
 
     $cmd->execute();
-    $self->warn("$why produced warnings: $err") if $err;
+    if ($err) {
+        my $warn_logger = ($err =~ m/\A[\s\n]*Last metadata expiration check.*[\s\n]*\z/m) ? 'verbose': 'warn';
```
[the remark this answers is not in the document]
stdweird: dnf always reports "Last metadata expiration check" to stderr, even if there's nothing wrong. now, if there's only that message, it's not considered a warning, and reported verbose.

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -248,11 +249,11 @@ sub generate_repos
 # the first element is the executable
```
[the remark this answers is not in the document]
stdweird: maybe the lines before make it clearer?

ncm-spma/src/main/perl/spma/yum.pm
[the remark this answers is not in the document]
ned21: Can you add that as a comment please? Especially as that sounds like a defect in dnf which might get fixed in the future.

jrha:
Bumped. Seems like this needs a bit longer to cook.

ncm-spma/src/main/pan/components/spma/yum/light.pan
```diff
@@ -2,8 +2,10 @@
 # ${developer-info}
 # ${author-info}
 
-@{spma backend specific statements for spmalight}
+@{spma yum backend specific statements for spmalight}
```
ned21, line 5: I think I missed something - where do we describe the purpose of these "light" variants?

ncm-spma/src/main/perl/spma/yum.pm
```diff
@@ -66,6 +67,7 @@ use constant YUM_CMD => qw(yum -y shell);
 use constant YUM_DISTRO_SYNC => qw(yum -y distro-sync);
 
 use constant VALID_PACKAGE_NAME => qr{^([\w\.\-\+]+)[*?]?};
+use constant REPO_INCLUDE => 1;
```
ned21, line 70: Use Readonly for newly introduced constants?

ncm-spma/src/main/perl/spma/yumdnf.pm
```diff
@@ -0,0 +1,91 @@
+use strict;
+use warnings;
+
+package NCM::Component::spma::yumdnf;
+
+use NCM::Component::spma::yum;
+push our @ISA , qw(NCM::Component::spma::yum);
+
+use constant YUM_CONF_FILE => "/etc/dnf/dnf.conf";
```
ned21, line 9: Use read-only instead since this is new code?

ncm-spma/src/main/pan/components/spma/yum/light.pan
stdweird: https://github.com/quattor/configuration-modules-core/blob/master/ncm-spma/src/main/pan/components/spma/light.pan

ncm-spma/src/main/perl/spma/yum.pm
stdweird: i prefer not to mix those, the inheritance is also a bit different, so i prefer not to

ncm-spma/src/main/perl/spma/yumdnf.pm
stdweird: need to be the same as in the parent `yum.pm`

ncm-spma/src/main/pan/components/spma/yum/light.pan
ned21: 👍  - this issue came up just this week, I didn't realise this was designed to fix!

jrha:
@hpcugentbot please test this

jrha:
@hpcugentbot test this please

jrha:
@stdweird I'm happy with the code (we're using it at RAL), are you able to deal with @ned21's last couple of comments?

ncm-spma/src/main/perl/spma/yumdnf.pm
jrha: I've made this a `debug 1` message.

ncm-spma/src/main/perl/spma/yum.pm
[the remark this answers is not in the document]
jrha: I've added a comment based on @stdweird's explanation.

jrha:
@ned21 I've made the changes that you requested on @stdweird's behalf.

ned21 (approved):

ned21:
🚢 it

PR #1402, opened 2019-11-03
ncm-network: always enable the network service

stdweird:
based on #1372 
Relying on only ncm-chkconfig can give issues in case of failures with ncm-chkconfig (and network service is not even present by default on EL8)


ncm-network/src/main/perl/network.pm
```diff
@@ -1983,6 +1983,14 @@ sub Configure
             # check if this was due to addition of resolv_mods / peerdns
             my $no_resolv = $self->make_ifcfg($ifacename, $iface, $ipv6, resolv_mods => 0, peerdns => 0);
             $self->legacy_keeps_state($file_name, $no_resolv, $exifiles);
+        };
+
+        if ($exifiles->{$file_name} == $UPDATED) {
+            # This is a UGent issue: a version with resolv_mods only was once build and deployed.
+            # Adding only the peerdns causes a restart
+            # TODO: get rid of this after a weeks
```
ned21: Should this merged as-is if it's a UGent only temp fix?

stdweird: @ned21 no, we can get rid of this next week i think. (but we need this in a PR for our internal build tools)

jrha:
I have removed the temporary commit and rebased on master.

jrha (dismissed):
LGTM AFAICT

jrha:
rebased

jrha (approved):
LGTM

ned21 (approved):

PR #1404, opened 2019-11-07
ncm-spma DNF backend

gombasg:
This PR is supposed to supersde PR #1309 . I've modified Jindrich's code to adapt it to RHEL8, and added a few improvements. There is no support for module streams yet.


ncm-spma/pom.xml
```diff
@@ -213,12 +213,6 @@
           <groupId>org.codehaus.mojo</groupId>
           <artifactId>rpm-maven-plugin</artifactId>
           <configuration>
-            <requires>
```
stdweird: euhm, not really acceptable imho. it's an issue but this is not the solution imho.

gombasg: I'm open to ideas, I've even looked into the sources of `rpm-maven-plugin` to see if it is possible to inject custom .spec file fragments, but I did not find a better solution so far. This change at least allowed me to install the RPM on RHEL 8.

CCM has a similar problem, let me just file an issue...

gombasg: https://github.com/quattor/CCM/issues/202 is similar in nature.

stdweird: yep, i also looked into it. for now, i made a dummy rpm that resolves the deps issues so i can continue testing. 
maven-rpm canbnot do this (ie generate different rpms with different requires). i only see a way to have the release script use rpmrebuild to make el8 packages with proper dependencies.

gombasg: Btw. I'd like to add a proper `systemd` unit to `cdp-listend` and `ncm-cdispd`, so I'd support replacing `rpm-maven-plugin` with a direct tall to `rpmbuild`. I just don't know how to do that nicely with maven.

gombasg: I've removed this commit from the PR, meaning the RPM cannot be installed on RHEL8.

jrha (approved):
LGTM, we don't use this code path, but have tested packages built with it included.

PR #1406, opened 2019-11-13
ncm-metaconfig: add dellnetworking configuration

stdweird:
Tested on OS10 only, probably needs splitted TT structure to support OS9


ned21:
😕 It says ncm-network but it changes ncm-metaconfig? Is https://jenkins0.ugent.be likely to come back so we can test this?

Is it possible to fix the panlint errors?

stdweird:
@ned21 that was a bit sloppy, i fixed the title and the test failure.
not fixing panlint, as it is in a test profile and this is done the same way as all other services here (the inclusion of the metaconfig profile is done)

Issue #1395, opened 2019-08-26
ncm-systemd: try-restart should be skipped for units which do not allow manual start/stop


gombasg:
If a unit has either `RefuseManualStart=true` or `RefuseManualStop=true`, then the `systemctl try-restart` command emitted by `ncm-systemd` will fail. The component should detect the presence of those options (even if these options are not coming from the Quattor-managed part of the unit definition), and should not try to restart such services.

As a workaround, allowing the profile to say a service should not be restarted even if its unit changes would also work, and could actually be useful regardless of this issue.

PR #1407, opened 2019-11-14
ncm-systemd: Do not try to restart sevices which cannot

gombasg:
Attempt to address issue #1395, but only if the RefuseManual* directives
are defined in the profiles. If those directives are coming from unit
fragments which are not managed by Quattor, then restart is still
attempted, and the component will fail.

### DIFF
--- a/ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
+++ b/ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
@@ -443,8 +443,17 @@ sub configured_units
                 next;
             } elsif ($changed) {
                 if ($detail->{startstop}) {
-                    $self->verbose("Going to issue conditional restart for $detail->{name} due to changed unitfile");
-                    push(@tryrestart, $detail->{name});
+                    $ufile->{config}->reset;
+                    my $utree = $ufile->{config}->getTree;
+                    # This is incomplete, because the RefuseManual* options may come from
+                    # unit fragments not managed via templates. But $unit_cache does not seem
+                    # to be initialized here, so we cannot check "systemctl show" data easily.
+                    if ($utree->{unit}->{RefuseManualStop} || $utree->{unit}->{RefuseManualStart}) {
+                        $self->verbose("$detail->{name} is configured to disallow manual restart, skip");
+                    } else {
+                        $self->verbose("Going to issue conditional restart for $detail->{name} due to changed unitfile");
+                        push(@tryrestart, $detail->{name});
+                    }
                 } else {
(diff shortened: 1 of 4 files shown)


ncm-systemd/src/test/resources/service-unit_simple_services.pan
```diff
@@ -1,19 +1,51 @@
 unique template service-unit_simple_services;
 
+include if (false) { "components/systemd/config"; };  # Try to game panlint
```
jrha: I don't want to see this.
The panlint issue is fixed by quattor/pan#236.

jrha:
I've removed the line gaming panlint as the issue it was working around no-longer exists.

Pansanel (dismissed):
The changes look fine to me.

ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
```diff
@@ -442,9 +442,18 @@ sub configured_units
                 $self->error("Unitfile configuration failed, skipping the unit ", $self->unit_text($detail));
                 next;
             } elsif ($changed) {
-                $self->verbose("Going to issue conditional restart for $detail->{name} due to changed unitfile");
-                push(@tryrestart, $detail->{name});
-            };
+		$ufile->{config}->reset;
```
stdweird, line 445: i sort of missed this. seems to be related to #1453 but instead of configuring th eunit with 'Refuse', i disabled startstop.
but indeed, this is better detected automatically.

we can initiailse the unit_cache, we'll just have to do that again after th eunitfiles have been generated. not sure how slow it is.

however, you also need it where startstop is handled elsewhere. a unit filw 'Refuse' configured, should ignore startstop also in the Service code.

jrha: @stdweird does this imply changes to this PR, or new issues to be opened?

stdweird: at the very least open an issue. i think it ow only takes into consideration self defined units, not existig ones.
the 2nd remark is more hypothetical (why would you flag a unit startstop if it has Refused configured)

open an issue, i can have a look at it after this is merged

jrha:
@stdweird I've rebased this and resolved the conflict with the change you made in #1453, can you check that this still makes sense?

jrha (dismissed):

jrha:
@stdweird and @gombasg after looking at this again I realised I had screwed up the previous rebase. I am pretty sure this now makes sense again, but would appreciate someone else taking a look.

ned21 (approved):

PR #1411, opened 2019-11-29
ncm-symlink minor PAN template fixes

ned21:
ncm-symlink: tweak default replace options

    The "all" and "none" replace options are mutually exclusive. The prefix
    syntax meant there was no way for site-config to set replace/all without
    modifying this file to comment out the replace/none line in config.pan.
    By moving the setting one level higher in the tree, replace/none will
    only be set if site-config has not created the replace sub-tree.

ncm-symlink: update schema to use absolute_file_path type

    Specifying a symlink name with a trailing slash leads to a runtime
    error. Instead use the schema to catch and prevent these errors at
    compile time.


stdweird:
jenkins test this please

PR #1388, opened 2019-06-14
ncm-metaconfig: nginx add more ssl configuration options

wdpypere:
- added a bunch of settings to be able to better configure SSL on nginx.
- changed the ciphersuite options so it is possible to follow https://mozilla.github.io/server-side-tls/ssl-config-generator/
- changed the default of gzip to false. As I understand it, if misconfigured it leaves you open to BREACH.

PR #1412, opened 2019-12-09
ncm-metaconfig: nginx: configure multiple multiple listen per server

stdweird:


jrha:
I've rebased this to deal with the conflicts from merging #1388.

PR #1422, opened 2020-01-16
ncm-metaconfig: slurm setting escape fix

itkovian:
We need to have the escaped version here, since the contents will be unescaped when deploying the conf file.

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
@@ -86,7 +86,7 @@ type slurm_scheduler_parameters = {
     'bf_yield_interval' ? long(0..)
     'bf_yield_sleep' ? long(0..)
     'build_queue_timeout' ? long(0..)
-    'default_queue_depth' ? long(0..)
+    'default_5fqueue_5fdepth' ? long(0..)   # escaped version, otherwise unescaping produces some weird unicode
     'defer' ? boolean
     'delay_boot' ? long(0..)
     'default_gbytes' ? boolean
--- a/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
@@ -95,7 +95,7 @@ type slurm_scheduler_parameters = {
     'enable_user_top' ? boolean
     'Ignore_NUMA' ? boolean
     'inventory_interval' ? long(0..)
-    'kill_invalid_depend' ? boolean
+    'kill_5finvalid_5fdepend' ? boolean
     'max_array_tasks' ? long(0..)   # should be smaller than MaxArraySize
     'max_depend_depth' ? long(0..)
     'max_rpc_cnt' ? long(0..)
--- a/ncm-metaconfig/src/main/metaconfig/slurm/tests/profiles/config.pan
+++ b/ncm-metaconfig/src/main/metaconfig/slurm/tests/profiles/config.pan
@@ -107,7 +107,7 @@ prefix "/software/components/metaconfig/services/{/etc/slurm/slurm.conf}/content
 #SchedulerTimeSlice = 30;
 "SchedulerType" = "backfill";
 "SchedulerParameters" = dict(
-    "default_queue_depth", 128,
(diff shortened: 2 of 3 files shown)


stdweird:
`"convert/unescapekey" = true;` needed in the config.pan

ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
```diff
@@ -86,7 +86,7 @@ type slurm_scheduler_parameters = {
     'bf_yield_interval' ? long(0..)
     'bf_yield_sleep' ? long(0..)
     'build_queue_timeout' ? long(0..)
-    'default_queue_depth' ? long(0..)
+    'default_5fqueue_5fdepth' ? long(0..)   # escaped version, otherwise unescaping produces some weird unicode
```
jrha: I don't understand the need for this, are user really expected to assign to a path like this?
```pan
prefix "/software/components/metaconfig/services/{/etc/slurm/slurm.conf}/content"
"SchedulerParameters/default_5fqueue_5fdepth" = 128;
```
It seems very user-unfriendly, if there is a rendering issue, surely the schema is not the place to address it?

stdweird:
jenkins test this please

ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
jrha: Ah I see, so the issue is that escaping doesn't work in schema files?

jrha: Ok, so I've just tested this and `{}` escaping seems to work fine in types, I would therefore prefer the following:
```suggestion
    '{default_queue_depth}' ? long(0..)   # escaped version, otherwise unescaping produces some weird unicode
```

ncm-metaconfig/src/main/metaconfig/slurm/tests/profiles/config.pan
```diff
@@ -107,7 +107,7 @@ prefix "/software/components/metaconfig/services/{/etc/slurm/slurm.conf}/content
 #SchedulerTimeSlice = 30;
 "SchedulerType" = "backfill";
 "SchedulerParameters" = dict(
-    "default_queue_depth", 128,
+    "default_5fqueue_5fdepth", 128,
```
ned21, line 110: Should be:
"{default_queue_depth}", 128,
as per the comment above about how users are expected to specify config?

ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
ned21: Wouldn't this change need to apply to every attribute containing _ ?

jrha: I would have thought so, but perhaps the `_d` is what is tripping it up here.

jrha: I still don't fully understand why the key needs to be escaped.

ncm-metaconfig/src/main/metaconfig/slurm/tests/profiles/config.pan
jrha, line 110:
```suggestion
    "{default_queue_depth}", 128,
```

ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
```diff
@@ -95,7 +95,7 @@ type slurm_scheduler_parameters = {
     'enable_user_top' ? boolean
     'Ignore_NUMA' ? boolean
     'inventory_interval' ? long(0..)
-    'kill_invalid_depend' ? boolean
+    'kill_5finvalid_5fdepend' ? boolean
```
jrha, line 98:
```suggestion
    '{kill_invalid_depend}' ? boolean
```

jrha:
@itkovian can you make the suggested changes to use `{…}` escaping? With those made I would be happy to merge.

wpoely86:
All of these changes were part of #1493

PR #1425, opened 2020-01-16
ncm-metaconfig: devicemapper: fix bug in schema

kwaegema:
current type does not make sense

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/devicemapper/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/devicemapper/pan/schema.pan
@@ -61,7 +61,7 @@ type multipath_device = {
     include multipath_device_blacklist
     "revision" ? string
     "product_blacklist" ? string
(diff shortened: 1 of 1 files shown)


ncm-metaconfig/src/main/metaconfig/devicemapper/pan/schema.pan
```diff
@@ -61,7 +61,7 @@ type multipath_device = {
     include multipath_device_blacklist
     "revision" ? string
     "product_blacklist" ? string
-    "hardware_handler" ? string with match(SELF[1], "^(emc|alua|hp_sw|rdac)$") && is_long(SELF[0]) && SELF[0] == 1
+    "hardware_handler" ? string with match(SELF, '^"1 (emc|alua|hp_sw|rdac)"$')
```
jrha: How about using the `choice` type?
```suggestion
    "hardware_handler" ? choice("emc", "alua", "hp_sw", "rdac")
```

PR #1431, opened 2020-02-27
ncm-cron: Provide schema validation to match the component code

### LINKED ISSUE #1429: ncm-cron: Minutes field should not be optional if smear is used
The minutes field is used to calculate the start of the smear offset so cannot be left blank as this implies `*` (i.e. all minutes of an hour).

The same problem would also likely manifest if only day was specified, but some problems are too much for my brain.

The Perl code already validates this, but it would be more user-friendly if this was caught at compile time.

jrha:
Resolves #1429.


stdweird:
jenkins test this please

PR #1433, opened 2020-03-05
ncm-icinga: add missing notification options

wdpypere:
notification can be disabled with 'n', but the options were missing in the schema.


stdweird:
jenkins test this please

PR #1443, opened 2020-05-19
ncm-metaconfig: haproxy allow configuration of acl on the frontends

wdpypere:


ncm-metaconfig/src/main/metaconfig/haproxy/pan/schema.pan
```diff
@@ -190,10 +190,12 @@ type haproxy_service_bind_params = {
 };
 
 type haproxy_service_frontend = {
+    'acl' ? string[]
```
stdweird, line 193: make this a dict, the key is the name of the acl

PR #1446, opened 2020-05-22
ncm-ssh: Add schema entry for RevokedKeys option

drossy:
Add schema entry and validation for RevokedKeys option

Adds the RevokedKeys option to allow the banning of ssh keys that have become compromised.
Optional so I don't expect backwards compatibility issues.

I don't think the error call in the validation will ever be reached - in testing is_absolute_file_path() threw the error, but I don't know if that matters.

### DIFF
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -134,6 +134,12 @@ type ssh_daemon_options_type = {
     "Port" ? long
     "PrintLastLog" ? legacy_binary_affirmation_string
     "PrintMotd" ? legacy_binary_affirmation_string
+    "RevokedKeys" ? string with {
+        if(!((SELF == 'none' ) || is_absolute_file_path(SELF))) {
+            error("RevokedKeys must either be a file path or none")
+        };
+        true;
+    }
     "RhostsAuthentication" ? legacy_binary_affirmation_string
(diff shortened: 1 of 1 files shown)


stdweird:
@drossy feel free to simpligy the `with` part if the custom error is not going to be called at all

out of interest, how do you plan to manage and/or distribute the revokedkeys file?

drossy:
> @drossy feel free to simpligy the `with` part if the custom error is not going to be called at all

Will do.

> out of interest, how do you plan to manage and/or distribute the revokedkeys file?

I was just going to filecopy it in, but haven't gotten to that part yet.

ned21:
https://jenkins0.ugent.be/job/NCM_Components-pull_requests/2940/ doesn't load for me. @drossy if you rebase this branch on master you'll pick up a GitHub action that runs the tests and you should then be able to see the error from the unit tests.

stdweird:
jenkins test this please

stdweird:
@ned21 a dep was removed on the jenkins server, is fixed now

drossy:
> > @drossy feel free to simpligy the `with` part if the custom error is not going to be called at all
> 
> Will do.

@stdweird After thinking about it I think I'll leave it as - it currently matches the way the existing validations are structured in this file, and its somewhat defensively coded if is_absolute_file_path() changes in the future to working how I thought it worked (just returning true/false rather than return true/throw error)

PR #1448, opened 2020-06-05
ncm-systemd: Add support for Socket in units

wpoely86:
This add supports for a `[Socket]` section in a systemd unit.

It builds on top of and requires #1396 

Please have look @stdweird

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -45,6 +45,9 @@ type ${project.artifactId}_valid_unit = string;
 # executable paths can have a number of special prefixes
 type ${project.artifactId}_valid_execpath = string with match(SELF, '^([@+!:-]|!!)?/');
 
+# type for a relative directory: no leading / and may not include ".."
+type ${project.artifactId}_relative_directory = string with !match(SELF, '(^/|\.\.)');
+
 # adding new ones
 # go to http://www.freedesktop.org/software/systemd/man/systemd.directives.html
 # and follow the link to the manual
@@ -149,9 +152,9 @@ valid for [Service], [Socket], [Mount], or [Swap] sections
 }
 type ${project.artifactId}_unitfile_config_systemd_exec = {
     'CacheDirectoryMode' ? type_octal_mode
-    'CacheDirectory' ? string[]
+    'CacheDirectory' ? ${project.artifactId}_relative_directory[]
     'ConfigurationDirectoryMode' ? type_octal_mode
-    'ConfigurationDirectory' ? string[]
+    'ConfigurationDirectory' ? ${project.artifactId}_relative_directory[]
     'CPUAffinity' ? long[][] # start with empty list to reset
     'CPUSchedulingPolicy' ? string with match(SELF, '^(other|batch|idle|fifo|rr)$')
     'CPUSchedulingPriority' ? long(1..99) # 99 = highest
@@ -178,19 +181,19 @@ type ${project.artifactId}_unitfile_config_systemd_exec = {
     'LimitSIGPENDING' ? long(-1..) # Specifies the limit on the number of signals that may be queued for the real user ID of the calling process.
     'LimitSTACK' ? long(-1..) # The maximum size of the process stack, in bytes.
     'LogsDirectoryMode' ? type_octal_mode
-    'LogsDirectory' ? string[]
+    'LogsDirectory' ? ${project.artifactId}_relative_directory[]
     'Nice' ? long(-20..19)
     'OOMScoreAdjust' ? long(-1000..1000)
     'PrivateTmp' ? boolean
-    'RootDirectory' ? string
+    'RootDirectory' ? ${project.artifactId}_relative_directory
     'RuntimeDirectoryMode' ? type_octal_mode
     'RuntimeDirectoryPreserve' ? choice('yes', 'no', 'restart')
-    'RuntimeDirectory' ? string[]
+    'RuntimeDirectory' ? ${project.artifactId}_relative_directory[]
     'StandardError' ? ${project.artifactId}_unitfile_config_systemd_exec_stdouterr
     'StandardInput' ? string with match(SELF, '^(null|tty(-(force|fail))?|socket)$')
     'StandardOutput' ? ${project.artifactId}_unitfile_config_systemd_exec_stdouterr
     'StateDirectoryMode' ? type_octal_mode
-    'StateDirectory' ? string[]
+    'StateDirectory' ? ${project.artifactId}_relative_directory[]
     'SupplementaryGroups' ? defined_group[]
     'SyslogFacility' ? syslog_facility
     'SyslogIdentifier' ? string
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -42,6 +42,9 @@ type ${project.artifactId}_unit_virtualization = string with match(SELF,
 #    make this more finegrained, e.g. has to be existing unit; or check types
 type ${project.artifactId}_valid_unit = string;
 
+# executable paths can have a number of special prefixes
+type ${project.artifactId}_valid_execpath = string with match(SELF, '^([@+!:-]|!!)?/');
+
 # adding new ones
 # go to http://www.freedesktop.org/software/systemd/man/systemd.directives.html
 # and follow the link to the manual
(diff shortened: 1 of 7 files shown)


ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -233,6 +247,77 @@ type ${project.artifactId}_unitfile_config_service = {
...
+the [Socket] section
+http://www.freedesktop.org/software/systemd/man/systemd.socket.html
+}
+type ${project.artifactId}_unitfile_config_socket = {
+    include ${project.artifactId}_unitfile_config_systemd_exec
+    include ${project.artifactId}_unitfile_config_systemd_kill
+    'ListenStream' ? string[]
+    'ListenDatagram' ? string[]
+    'ListenSequentialPacket' ? string[]
+    'ListenFIFO' ? string with match(SELF, '^/')
```
stdweird, line 260: there is an absolute_file_path type somewhre

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -233,6 +247,77 @@ type ${project.artifactId}_unitfile_config_service = {
...
+    'ListenSpecial' ? string with match(SELF, '^/')
+    'ListenNetlink' ? string
+    'ListenMessageQueue' ? string with match(SELF, '^/')
+    'ListenUSBFunction' ? string
+    'SocketProtocol' ? string with match(SELF, '^(udplite|sctp)$')
```
stdweird, line 265: use `choice`

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -233,6 +247,77 @@ type ${project.artifactId}_unitfile_config_service = {
...
+    'BindIPv6Only' ? string with match(SELF, '^(default|both|ipv6-only)$')
+    'Backlog' ? long(0..)
+    'BindToDevice' ? string
+    'SocketUser' ? defined_user
+    'SocketGroup' ? defined_group
+    'SocketMode' ? long # octal
```
stdweird, line 271: these probably can also be limited `(0..1023)` (ot sure anyone wants to set the sticky bit, otherwise stop at 511

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -233,6 +247,77 @@ type ${project.artifactId}_unitfile_config_service = {
...
+    'SELinuxContextFromNet' ? boolean
+    'PipeSize' ? long(0..)
+    'MessageQueueMaxMessages' ? long
+    'MessageQueueMessageSize' ? long
+    'FreeBind' ? boolean
+    'Transparent' ? boolean
+    'Broadcast' ? boolean
+    'PassCredentials' ? boolean
+    'PassSecurity' ? boolean
+    'TCPCongestion' ? string with match(SELF, '^(westwood|veno|cubic|lp)$')
```
stdweird, line 302: choice

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -233,6 +247,77 @@ type ${project.artifactId}_unitfile_config_service = {
...
+    'TimeoutSec' ? long(0..)
+    'Service' ? string
+    'RemoveOnStop' ? boolean
+    'Symlinks' ? string[]
+    'FileDescriptorName' ? string with match(SELF, '^[^:]{1,255}$')
+    'TriggerLimitIntervalSec' ? long(0..)
+    'TriggerLimitBurst' ? long(0..)
+} with {
+    if(exists(SELF['Service']) && exists(SELF['Accept']) && SELF['Accept']) {
+        error('You can only specify a service when Accept=no');
```
stdweird, line 316: `service` -> `Service`, and `Accept=no` -> `Accept=false`

stdweird:
@wpoely86 minor remarks, thanks for this!

wpoely86:
@stdweird I just noticed a thing: `ExecStartPost` (and friends). I've defined them as a list as systemd allows multiple lines of those. In the `Service` section, they are also defined but as string. The tt files do however have them in `list_of_lines` (together with `ExecStart` which is not correct according to the man page). So it's unclear if that can work at all to me.
I can change it, but that's a breaking change.

stdweird:
@wpoely86 `ExecStart` and friens should work as list_of_lines. it's doucmented and even has an example.

wpoely86:
@stdweird can you point tot that documentation? I can't find it.

As far as I understand how this works (and my knowledge is pretty limited), it shouldn't work.

Multiple `ExecStart` are in any case forbidden by systemd (according to the man page).

stdweird:
@wpoely86 where is that stated?
```
If more than one command is specified, the commands are invoked sequentially in the order they appear in the unit file. If one of the commands fails (and is not prefixed with "-"), other lines are not executed, and the unit is considered failed.
```
and a bit lower
```
Example:

Environment=ONE='one' "TWO='two two' too" THREE=
ExecStart=/bin/echo ${ONE} ${TWO} ${THREE}
ExecStart=/bin/echo $ONE $TWO $THREE

This results in /bin/echo being called twice, the first time with arguments "'one'", "'two two' too", "", and the second time with arguments "one", "two two", "too". 
```
i don't know if i ever tried this

wpoely86:
```
       ExecStartPre=, ExecStartPost=
           Additional commands that are executed before or after the command in ExecStart=, respectively. Syntax is the same as for ExecStart=, except that multiple command lines are allowed and the commands are executed one after the other, serially.
```

wpoely86:
OK, the docs are confusing:
```
Unless Type= is oneshot, exactly one command must be given. When Type=oneshot is used, zero or more commands may be specified. 
```

So it's fine, but only when it's a oneshot.

But in any case, how you specify this? Currently `ExecStart` is a string?

stdweird:
@wpoely86 ok, i'm also confused now what should be working. 
but indeed, either the schema becomes a list of strings or the TT shouldn't use list_of_lines. for now, i would remove ExecStart from list_of_lines and see what the unittests do.

wpoely86:
The unit tests keep on working just fine if you remove it. I'm pretty sure it doesn't work right now. You can only specify a single command. The `? string` needs to be `? string[]`.

I will open a PR for it once this is merged.

stdweird:
@wpoely86 thanks. can you rebase on the other PR also cleanup the commit history here.

wpoely86:
@stdweird done

stdweird:
@wpoely86 for me this is ok. i'll let @ned21 decide if using long type and assumig users will know what they are doing is the way to go. i'm in favour, but i understand that it's error prone

stdweird:
@wpoely86 @ned21 actuallu, this is already in use elsewhere https://github.com/quattor/configuration-modules-core/blob/master/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan#L91
so i'm going to approve this PR

wpoely86:
@stdweird all checks finally passed!

stdweird:
@wpoely86 clearly the unittests do not try to include the component config.pan
the `%{}` mahgic in that file binds to `systemd_component` type, the curren scheam has `component_systemd` type. can you fix the schema ?

wpoely86:
@stdweird fixed and tests are go

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -145,6 +145,10 @@ http://www.freedesktop.org/software/systemd/man/systemd.exec.html
 valid for [Service], [Socket], [Mount], or [Swap] sections
 }
 type ${project.artifactId}_unitfile_config_systemd_exec = {
+    'CacheDirectoryMode' ? type_octal_mode
+    'CacheDirectory' ? string[]
```
ned21, line 152: Looks like systemd only accepts relative paths here, and `..` is banned - should we have a type that enforces that?

ned21 (commented):
Can you squash e769272 into the previous commit please since it seems to be a fixup?

ncm-systemd/src/main/pan/components/systemd/schema.pan
wpoely86: I can add something like:

```
# type for a relative directory (no leading /) and may not include ".."
type ${project.artifactId}_relative_directory = string with !match(SELF, '(^/|\.\.)');
```

But currently we don't really check anything like this. The `WorkingDirectory` and `RootDirectory` are also just strings.

wpoely86:
@ned21 commits squashed

ncm-systemd/src/main/pan/components/systemd/schema.pan
ned21: Yep,  I think that would make a lot of sense if you can add it easily enough.

wpoely86: Done. Please have a look.

ned21 (dismissed):

kwaegema:
@wpoely86 has merge conflicts now

wpoely86:
@kwaegema fixed.

ned21:
Any particular reason you merged from master instead of rebaseing ? It keeps the commit history easier to read if it's linear?

wpoely86:
@ned21 rebased instead of merged now.

ned21 (approved):

PR #1450, opened 2020-06-05
ncm-ssh: add support for Match lines in daemon

wpoely86:
sshd support 'Match' clause where you can change a limited set of options depending on the user, gorup, address, ...

I'm trying to add support for that.

@stdweird Can you have a look if this is a good idea? I still need to figure out how to get the perl code to properly output this.

### DIFF
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -96,6 +96,7 @@ type ssh_core_options_type = {
 
 # These are the options allowed under a Match clause
 type ssh_daemon_match_options_type = {
+    include ssh_authkeyscommand_options_type
     @{AcceptEnv, one per line}
     "AcceptEnv" ? string[]
     "AllowAgentForwarding" ? legacy_binary_affirmation_string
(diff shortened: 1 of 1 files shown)


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -89,31 +94,65 @@ type ssh_core_options_type = {
...
+    include ssh_daemon_match_options_type
+};
+
+type type_matchrule = {
+    "User" ? string[]
+    "Group" ? string[]
+    "Host" ? string[]
+    "LocalAddress" ? string[]
+    "LocalPort" ? string[]
+    "Addres" ? string[]
```
stdweird, line 141: 2 's'es

stdweird:
@wpoely86 tbh, i don thtink this is a good idea. better and easier to generate everything with TT or even with metaconfig. the only tricky bit to use metacofig is the testing in the code.
you can certainly use a TT file and generate using current schema (and enable some force config option to either eidt the file or generate it from scratch)

wpoely86:
I don't understand what you're saying. You still need a schema for metaconfig?

How can you mix that with this component? Both cannot write to the `sshd_config` file?

gombasg:
The problem is, this will end up being a similar-but-incompatible schema than what’s already available under metaconfig – and having to remember you need to specify the same things on the server side in a different way than on the client side is messy. I’ve sent a pull request some time ago to allow metaconfig to be used for both the client and the server side. That is still pending and I no longer have time to work on it. But there should really be just one schema – either extend ncm-ssh, and drop the one under ncm-metaconfig, or the other way around.

wpoely86:
OK, I didn't know there was also a schema under metaconfig for sshd. Yes, it makes more sense to build on that one.

stdweird:
@gombasg ok, then i should look at your PR again.

wpoely86:
#1452 will do this way beter.

PR #1451, opened 2020-06-10
ncm-metaconfig: support running arbitrary commands in various steps

stdweird:

### DIFF
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -87,14 +87,19 @@ type ${project.artifactId}_textrender_convert = {
 type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
 
 type ${project.artifactId}_actions = {
-    @{Always run, happens before possible modifications}
+    @{Always run, happens before possible modifications.
+      A failure will cancel any file modification, unless the command is prefixed with -.}
     'pre' ? string
-    @{Always run, happens before possible modifications, the file content is passed om stdin.
+    @{Always run before possible modifications with the new (or unchanged) file content is
+      passed on stdin. A failure will cancel any file modification,
+      unless the command is prefixed with -.
       Runs with 'keeps_state' enabled, so do not modify anything with this command.}
     'test' ? string
-    @{Only run after file is modified, but before any daemon action is executed. A failure in this command has no effect on whether the daemon action is executed later.}
+    @{Only run after file is modified, but before any daemon action is executed.
+      A failure in this command has no effect on whether the daemon action is executed later.}
     'changed' ? string
-    @{Always run, happens after possible modifications (and after 'changed').}
+    @{Always run, regardless of whether file was modified or not, and after the 'changed' action
+      but before any daemon action. A failure of this command has no effect on the subsequent daemon action.}
     'post' ? string
 };
 
@@ -137,7 +142,7 @@ type ${project.artifactId}_config =  {
 } = dict();
 
 @{Command must start with absolute path to executable.
-  If the executable is preceded with a '-', it means that the a command failure
+  If the executable is preceded with a '-', it means that a non-zero exit code (i.e. failure) is
   treated as success w.r.t. reporting and continuation.}
 type ${project.artifactId}_command = string with match(SELF, '^-?/');
 
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -130,16 +130,22 @@ type ${project.artifactId}_config =  {
     'contents' : ${project.artifactId}_extension
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
-    @{Actions (i.e. named registered commands) to run on pre, validation and/or post step.
-      These are independent of daemons and are executed at time of processing each service.}
+    @{Actions (i.e. names found in /software/components/metadata/commands) to run when processing the service.
+      Refer to the metaconfig_actions type definition for the available hooks
+      for when a command may be run.}
     'actions' ? ${project.artifactId}_actions
 } = dict();
+
+@{Command must start with absolute path to executable.
+  If the executable is preceded with a '-', it means that the a command failure
+  treated as success w.r.t. reporting and continuation.}
+type ${project.artifactId}_command = string with match(SELF, '^-?/');
 
 type ${project.artifactId}_component = {
     include structure_component
     'services' : ${project.artifactId}_config{} with valid_absolute_file_paths(SELF)
     @{Command registry for allowed actions, keys should be used as action value}
-    'commands' ? string{}
+    'commands' ? ${project.artifactId}_command{}
 } with {
     foreach (esc_fn; srv; SELF['services']) {
         if (exists(srv['actions'])) {
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -86,6 +86,18 @@ type ${project.artifactId}_textrender_convert = {
 
 type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
 
+type ${project.artifactId}_commands = {
+    @{Always run, happens before possible modifications}
+    'pre' ? string
+    @{Always run, happens before possible modifications, the file content is passed om stdin.
+      Runs with 'keeps_state' enabled, so do not modify anything with this command.}
+    'test' ? string
+    @{Only run after file is modified, but after daemon action is added}
+    'changed' ? string
+    @{Always run, happens after possible modifications (and after 'changed').}
+    'post' ? string
+};
+
 type ${project.artifactId}_config =  {
     @{File permissions. Defaults to 0644.}
     'mode' : long = 0644
@@ -118,6 +130,9 @@ type ${project.artifactId}_config =  {
     'contents' : ${project.artifactId}_extension
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
+    @{Commands to run on pre, validation and/or post step.
+      These are independent of daemons and are executed at time of processing each service.}
+    'commands' ? ${project.artifactId}_commands
 } = dict();
 
 type ${project.artifactId}_component = {
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -92,7 +92,7 @@ type ${project.artifactId}_actions = {
     @{Always run, happens before possible modifications, the file content is passed om stdin.
       Runs with 'keeps_state' enabled, so do not modify anything with this command.}
     'test' ? string
-    @{Only run after file is modified, but after daemon action is added}
+    @{Only run after file is modified, but before any daemon action is executed. A failure in this command has no effect on whether the daemon action is executed later.}
     'changed' ? string
     @{Always run, happens after possible modifications (and after 'changed').}
     'post' ? string
--- a/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
+++ b/ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
@@ -86,7 +86,7 @@ type ${project.artifactId}_textrender_convert = {
 
 type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
 
-type ${project.artifactId}_commands = {
+type ${project.artifactId}_actions = {
     @{Always run, happens before possible modifications}
     'pre' ? string
     @{Always run, happens before possible modifications, the file content is passed om stdin.
@@ -130,12 +130,26 @@ type ${project.artifactId}_config =  {
     'contents' : ${project.artifactId}_extension
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
-    @{Commands to run on pre, validation and/or post step.
+    @{Actions (i.e. named registered commands) to run on pre, validation and/or post step.
       These are independent of daemons and are executed at time of processing each service.}
-    'commands' ? ${project.artifactId}_commands
+    'actions' ? ${project.artifactId}_actions
 } = dict();
 
 type ${project.artifactId}_component = {
     include structure_component
     'services' : ${project.artifactId}_config{} with valid_absolute_file_paths(SELF)
+    @{Command registry for allowed actions, keys should be used as action value}
+    'commands' ? string{}
+} with {
+    foreach (esc_fn; srv; SELF['services']) {
+        if (exists(srv['actions'])) {
+            foreach (action; cmd_ref; srv['actions']) {
+                if (!(exists(SELF['commands']) && exists(SELF['commands'][cmd_ref]))) {
+                    error('Found %s action %s for %s, but no matching command registered',
+                            action, cmd_ref, unescape(esc_fn));
+                };
+            };
+        };
+    };
+    true;
 };
--- a/ncm-metaconfig/src/main/perl/metaconfig.pm
+++ b/ncm-metaconfig/src/main/perl/metaconfig.pm
@@ -160,15 +160,121 @@ our $EC = LC::Exception::Context->new->will_store_all;
 
 our $NoActionSupported = 1;
 
+# Given metaconfigservice C<$srv> for C<$file> and hash-reference C<$actions>,
+# prepare the actions to be taken for this service/file.
+# C<actions> is updated in-place; does not return anything.
+sub prepare_action
+{
+    my ($self, $srv, $file, $actions) = @_;
+
+    # Not using a hash here to detect and support
+    # any overlap with legacy daemon-restart config
+    my @daemon_action;
+
+    my $file_msg = "for file $file";
+
+    foreach my $daemon (sort keys %{$srv->{daemons} || {}}) {
+        push(@daemon_action, $daemon, $srv->{daemons}->{$daemon});
+    }
+
+    if ($srv->{daemon}) {
+        $self->verbose("Deprecated daemon(s) restart via daemon field $file_msg.");
+        foreach my $daemon (@{$srv->{daemon}}) {
+            if ($srv->{daemons}->{$daemon}) {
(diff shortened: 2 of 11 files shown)


ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -223,16 +223,61 @@ sub process_actions
...
+{
+    my ($self, $commands, $type, $input) = @_;
+
+    my $command = $commands->{$type};
+    if ($command) {
+        $self->debug(1, "Going to run $type command '$command'");
+
+        my ($err, $out);
+        my %opts = (
+            shell => 1,
```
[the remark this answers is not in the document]
stdweird: @gombasg we can also do that. at some point it might make more sense to store this command/action registry somewhere central; and have more direct access to the commands from NCM::NCD or CAF::Proces.
i can start with putting where you suggest it, and see how useful it is.

(and obviously, commands need to be registered with `final` keyword.)

stdweird:
@gombasg i tried to add it. the naming is now a total mess far beyond my capabilties

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -118,9 +130,26 @@ type ${project.artifactId}_config =  {
     'contents' : ${project.artifactId}_extension
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
+    @{Actions (i.e. named registered commands) to run on pre, validation and/or post step.
```
ned21, line 133: Actions (i.e. names found /software/components/metadata/commands) to run when processing the service. Refer to the ${project.artifactId}_actions type definition for the available hooks for when a command may be run.

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -118,9 +130,26 @@ type ${project.artifactId}_config =  {
...
+      These are independent of daemons and are executed at time of processing each service.}
+    'actions' ? ${project.artifactId}_actions
 } = dict();
 
 type ${project.artifactId}_component = {
     include structure_component
     'services' : ${project.artifactId}_config{} with valid_absolute_file_paths(SELF)
+    @{Command registry for allowed actions, keys should be used as action value}
+    'commands' ? string{}
```
ned21, line 142: I think we need to say something about how commands are executed. i.e. (depending on discussion below): The commands will be executed via CAF::Process without shell.

ncm-metaconfig/src/main/perl/metaconfig.pm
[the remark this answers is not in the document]
ned21: I'd still vote for shell=>0. Looking at the docs, this should work just fine if we want the input to be strings?

```
my $cmd = CAF::Process->new(["ls -lh"], log => $self);
```

One of the reasons this structure is nice is that I can also nullify or replace any command I don't like with a central dict update:

```
'metaconfig/commands/not-for-me' = '/bin/true';
```

without having to loop through all `services` to find the right one.

I could potentially also do that conditionally using Aquilon's build status to e.g. avoid running commands that I know won't work during anaconda and but allowing them to run fine once we've left buildstatus=build.

(yeah, that's a bit ugly but potentially less so than the `cmd||/bin/true` hacks we have at the moment)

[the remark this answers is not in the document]
ned21: Just spotted this use of shell commands in our code base:
```
/usr/sbin/alternatives --set mta /usr/sbin/sendmail.sendmail; systemctl restart sendmail.service
```

But then with this code, I think we could still do that? The alternatives command would be run by the changed or post action, and then the restart would happen via the standard daemons attribute.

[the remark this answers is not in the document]
stdweird: the daemons are only triggered by a change, you can also run the alternatives in the pre section.
when is it triggered now? only on change?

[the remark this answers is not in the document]
ned21: Yep, on change only via file copy.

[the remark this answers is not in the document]
stdweird: in that case, yes this would work (but only because you are lucky enough that last command is a daemon action).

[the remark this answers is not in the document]
ned21: Right, but my point is that a lot of where we currently have shell "tricks" could be re-thought to be cleaner. But if we give people a $shell=>1 crutch they'll just port directly across from filecopy without trying to find a better way. 

What we found with file copy is that once sites are using shell=1, there's now way to back out of it without risk of breakage. So it's much better to start with shell=0 and then only turn it on if we find a proper use case that can't be solved any other way.

ned21:
This is a long thread so let me try summarise. I think we're pretty close to:

- [ ] I think we should use shell=0 rather than =1 because we can't disable it later.
- [ ] I made a wording suggestion on the docs.
- [ ] One of the points that Gabor made is that sometimes we don't care if a command fails (it's one of the drivers for shell syntax to allow `||/bin/true`), or rather we don't want it to cause the component to "fail" and thus prevent any downstream dependencies from running. Should we support a flag, or a different hook since it's only applicable to changed and post, that is "run but don't error on failure"?

stdweird:
@ned21 
- i'll look for a splitter, if you know one, please do share
- for gabor's request: we can do something systemd style, and the command that starts with a `-` can fail without error.

other question: can we force that all commands start with a absolute path to the executable? otherwsie, i have to invent my own path resolution too.

stdweird:
@ned21 i'll use `use Text::ParseWords qw(old_shellwords)` for the splitting (has el7 and el8 packages)

ned21:
> i'll use `use Text::ParseWords qw(old_shellwords)` for the splitting (has el7 and el8 packages)

What about el6? :-( Do we need a splitter though? the CAF::Process doc says it accepts a string with shell=0. Or is that a mis-reading on my part?

> for gabor's request: we can do something systemd style, and the command that starts with a - can fail without error.

I like that.

> other question: can we force that all commands start with a absolute path to the executable? otherwsie, i have to invent my own path resolution too.

Yes, absolutely paths only.

stdweird:
@ned21 for el6, i find no packages, and some sites seem to suggest it is part of perl 5.10 itself. can you check that? and where did you read that about CAF?

ned21:
Can't find any trace of Text-Parsewords on an EL6 system. However I tested:

```
use CAF::Process;
my $proc = CAF::Process->new(["ls /"]);
my $output = $proc->output();
$proc->execute();
```

Shows the contents of /.

```my $proc = CAF::Process->new(["ls /d*"]);```
gives: ```ls: cannot access /d*: No such file or directory.```

**Comment by stdweird** [comment]
@ned21 that splits on whitespace only, so no proper quote handling for args with space.
so `perl -e 'use Text::ParseWords'` fails on el6? what about `repoquery --whatprovides "perl(Text::ParseWords)"`

**Comment by gombasg** [comment]
On EL6, it is provided by “perl”. So if the Maven plugin uses the standard RPM macros to detect Perl dependencies, then everything should work. If not, then there’s a problem, because the dependency will need to be OS version specific – which needs to be implemented for RH8 support anyway…

**Comment by stdweird** [comment]
@gombasg only when the maven config has the dependecies explicitly set. otherwise, it's the usual perl.req script that adds them (in the `perl(A::B::C)` form).
can you open an issue about the el8 problems in the maven repo?

**Comment by stdweird** [comment]
@ned21 @gombasg i think i addressed the issues

**Comment by ned21** [comment]
ah, yep: `perl -e 'use Text::ParseWords'` was a good call. I was looking in `/usr/lib64/perl5/Text` not `/usr/share/perl5/Text/`

File: ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan [hunk 4]
```diff
@@ -118,9 +130,32 @@ type ${project.artifactId}_config =  {
...
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
+    @{Actions (i.e. names found in /software/components/metadata/commands) to run when processing the service.
+      Refer to the ${project.artifactId}_actions type definition for the available hooks
+      for when a command may be run.}
+    'actions' ? ${project.artifactId}_actions
 } = dict();
 
+@{Command must start with absolute path to executable.
+  If the executable is preceded with a '-', it means that the a command failure
```
ned21, line 140:
```suggestion
  If the executable is preceded with a '-', it means that a non-zero exit code (i.e. failure) is
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -118,9 +130,32 @@ type ${project.artifactId}_config =  {
     'contents' : ${project.artifactId}_extension
     @{Predefined conversions from EDG::WP4::CCM::TextRender}
     'convert' ? ${project.artifactId}_textrender_convert
+    @{Actions (i.e. names found in /software/components/metadata/commands) to run when processing the service.
+      Refer to the ${project.artifactId}_actions type definition for the available hooks
+      for when a command may be run.}
```
ned21, line 135: panlint is objecting to the indentation here: Indentation should be a multiple of four spaces

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -86,6 +86,18 @@ type ${project.artifactId}_textrender_convert = {
 
 type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
 
+type ${project.artifactId}_actions = {
+    @{Always run, happens before possible modifications}
+    'pre' ? string
+    @{Always run, happens before possible modifications, the file content is passed om stdin.
```
ned21, line 92:
```suggestion
    @{Always run, happens before possible modifications, the new file content is passed on stdin.
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -86,6 +86,18 @@ type ${project.artifactId}_textrender_convert = {
...
+      Runs with 'keeps_state' enabled, so do not modify anything with this command.}
+    'test' ? string
+    @{Only run after file is modified, but after daemon action is added}
```
ned21, line 95: What does "added" mean here? After the daemon action has been executed? Or just added to some queue for execution that will happen after these commands?

ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -153,19 +153,20 @@ use LC::Exception;
 use EDG::WP4::CCM::TextRender 18.6.1;
 use CAF::Service;
 use EDG::WP4::CCM::Path qw(unescape);
+use Text::ParseWords qw(old_shellwords);
```
ned21, line 156: Isn't is a bit risky to take a dependency on old_shellwords which is [deprecated](https://metacpan.org/pod/Text::ParseWords)? Why to use `shellwords`?

ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -265,55 +329,89 @@ sub handle_service
...
+
+    my $commands = {};  # this will trigger reporting that nothing is configured is this stays empty
+    foreach my $type (sort keys %$actions) {
+        my $action = $actions->{$type};
+        my $command = $command_registry->{$action};
+        if ($command) {
+            $commands->{$type} = $command;
+            $self->verbose("Resolved $type action $action to command '$command'");
+        } else {
+            # Not fatal, should be covered in schema already
```
ned21, line 361: What is meant by "not fatal" here? In some cases (i.e. for dependent components, any error() is "fatal". The rest of the comment reads as, "Should not happen due to this being validated in the schema"?

ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -265,55 +329,89 @@ sub handle_service
...
+            $self->error("Unable to resovle $type action $action to command");
```
ned21, line 362:
```suggestion
            $self->error("Unable to resolve $type action $action to command");
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
stdweird: this is a bug in panlint. i think it's confsued with closing `}` line before.

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
stdweird: it is always added to the queue (even if the changed command would fail), and then the changed command is executed. but the daemon action is only performed at the very end.

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: how did i miss the `shellwords`? i already foudn it very odd.  i'll defintely change this.

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: fatal as in stop parsing and do not execute the rest of the component code. thinking about it, i'd better make this fatal as it is the right thing to do, and the schema prevents that you can use such a config in a profile.

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
ned21:
```suggestion
    @{Only run after file is modified, but before any daemon action is executed. A failure in this command has no effect on whether the daemon action is executed later.}
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -86,6 +86,18 @@ type ${project.artifactId}_textrender_convert = {
...
+    'changed' ? string
+    @{Always run, happens after possible modifications (and after 'changed').}
```
ned21, line 97:
```suggestion
    @{Always run, regardless of whether file was modified or not, and after the 'changed' action but before any daemon action. A failure of this command has no effect on the subsequent daemon action.}
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
ned21:
```suggestion
    @{Always run before possible modifications with the new (or unchanged) file content is passed on stdin. A failure will cancel any file modification, unless the command is prefixed with -.
```

ncm-metaconfig/src/main/perl/metaconfig.pm
ned21: Seems harsh (since in theory the code could still deliver some useful value) but probably makes sense given it's a schema check.

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
stdweird: i find `new (or unchanged)` confusing. i would expect `new/changed (or unchanged)`,but maybe i need some sleep ;)
 
this is actually the generated content, and this runs before anything is written. on failure, it just stops, writing to disk or even comparing with previous content is not attempted (i think, i would have to look at the code in FileWriter).

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: i guess it's a personal preference. i prefer things stricter, i guess from your experience some useful value is probably better than nothing at all.
if you come up with better comment, i'll leave the code non-fatal ;)

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
```diff
@@ -86,6 +86,18 @@ type ${project.artifactId}_textrender_convert = {
 
 type caf_service_action = string with match(SELF, '^(restart|reload|stop_sleep_start)$');
 
+type ${project.artifactId}_actions = {
+    @{Always run, happens before possible modifications}
```
stdweird, line 90: maybe also add ` A failure will cancel any file modification, unless the command is prefixed with -.` here (because it's the case)?

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
ned21: I was unsure about "unchanged", if it clouds the situation then let's just go for:

Always run before possible modifications with the generated file content is passed on stdin. A failure will cancel any file modification, unless the command is prefixed with -.

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
ned21: Yes, good idea.

ncm-metaconfig/src/main/perl/metaconfig.pm
ned21:
```suggestion
            # Should not happen due to this being validated in the schema
```

ncm-metaconfig/src/main/pan/components/metaconfig/schema.pan
jrha: The same bug seems to exist in the syntax highlighting.

ned21 (changes requested):
Unless I have missed something, I think my documentation suggestions are still pending?

stdweird:
@ned21 please check again, i think it's ok now

ncm-metaconfig/src/main/perl/metaconfig.pm
ned21: Did `shellwords` not work after all?

ncm-metaconfig/src/main/perl/metaconfig.pm
ned21: Re-upping my suggested wording change.

ncm-metaconfig/src/main/perl/metaconfig.pm
ned21: typo

ncm-metaconfig/src/main/perl/metaconfig.pm
stdweird: oopsie, more comments i missed. i'll make the changes, can you do a `perl -e 'use Text::ParseWords qw(shellwords)'` to make sure it exists on el6?

stdweird:
@ned21 changes pushed

ned21 (dismissed):

stdweird:
@jrha conflicts resolved

ned21 (approved):

ned21:
Anything else that's needed here? We are really keen to have #1452 which depends on this.

stdweird:
@ned21 i'm happy with it :wink:

PR #1452, opened 2020-06-10
ncm-metaconfig: ssh daemon configuration support

stdweird:
Replaces #1377
Requires #1451


ncm-metaconfig/src/main/metaconfig/ssh/pan/server_config.pan
```diff
@@ -4,11 +4,12 @@ include 'metaconfig/ssh/schema';
 
 bind "/software/components/metaconfig/services/{/etc/ssh/sshd_config}/contents" = sshd_config_file;
 
-prefix "/software/components/metaconfig";
+# since final locks the whole path, bind it to a fix value and set it as default too
+#    TODO: support in compiler
+bind "/software/components/metaconfig/commands/sshd_test_stdin" =
```
ned21, line 9: So if a site wants to change this they need to patch the file rather than using e.g. site-config.pan?

stdweird: yes. i sort of see your point, but on the other hand also not. 

this is the `ncm-ssh` test code. you want to test the config in another way or something else enitrely, but if you know a better way to do so i'd like to hear about it.
but i guess this does not go well with eg OS changes.

i need to think about it. i would like that at least by default the templates provide something and that it is immutable. this is the closest i can think of right now.

if you have no real case, i'd like to keep it right now. PRs in future etc etc

ned21: We currently run sshd from AFS not /usr/ so one of our major motivations to allow arbitrary commands is to allow us to customise this. So making it necessary to fork it to modify this seems a little odd. The default makes perfect sense, but activating this module is optional, same as using ncm-ssh is optional today.

At some point I envisage us moving to /usr/bin/sshd, at which point happy to see this enabled.

ncm-metaconfig/src/main/metaconfig/ssh/pan/server_config.pan
```diff
@@ -0,0 +1,15 @@
+unique template metaconfig/ssh/server_config;
+
+include 'metaconfig/ssh/schema';
+
+bind "/software/components/metaconfig/services/{/etc/ssh/sshd_config}/contents" = sshd_config_file;
+
+# since final locks the whole path, bind it to a fix value and set it as default too
+#    TODO: support in compiler
```
ned21, line 8: Can we drop the TODO? Doesn't seem to add much information to the reader.

ncm-metaconfig/src/main/metaconfig/ssh/pan/server_config.pan
stdweird: @ned21 ok. i'll split the templates so it's easier to customise this bit.

ncm-metaconfig/src/main/metaconfig/ssh/pan/server_config.pan
stdweird: sure. out of in terest, do you want this better supported in the compiler, or do you think this overkill?

ncm-metaconfig/src/main/metaconfig/ssh/pan/server_config.pan
stdweird: @ned21 giving it some more thought, i'm not going to change it.
easiest way to customize it actually is to register a new command, and overwrite the `actions/test` below.

if you customise things before this is included, i can make it a default assignment below.

ned21: Good point - changing the config instead of changing the command makes a lot more sense.

ncm-metaconfig/src/main/perl/metaconfig.pm
```diff
@@ -265,55 +315,89 @@ sub handle_service
...
+    my $commands = {};  # this will trigger reporting that nothing is configured is this stays empty
+    foreach my $type (sort keys %$actions) {
+        my $action = $actions->{$type};
+        my $command = $command_registry->{$action};
+        if ($command) {
+            $commands->{$type} = $command;
+            $self->verbose("Resolved $type action $action to command '$command'");
+        } else {
+            # Not fatal, should be covered in schema already
+            $self->error("Unable to resovle $type action $action to command");
```
ned21, line 348:
```suggestion
            $self->error("Unable to resolve $type action $action to command");
```

jrha (approved):
AFAICT this is good to merge.

PR #1453, opened 2020-06-11
ncm-systemd: respect startstop when unitfile is changed

stdweird:


ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
```diff
@@ -442,8 +442,13 @@ sub configured_units
                 $self->error("Unitfile configuration failed, skipping the unit ", $self->unit_text($detail));
                 next;
             } elsif ($changed) {
-                $self->verbose("Going to issue conditional restart for $detail->{name} due to changed unitfile");
-                push(@tryrestart, $detail->{name});
+                if ($detail->{startstop}) {
+                    $self->verbose("Going to issue conditional restart for $detail->{name} due to changed unitfile");
+                    push(@tryrestart, $detail->{name});
+                } else {
+                    # Should we try to try-restart some dependency instead?
```
ned21, line 449: I don't understand this comment sorry?

ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
```diff
@@ -442,8 +442,13 @@ sub configured_units
...
+                    $self->verbose("Changed unitfile for $detail->{name} but no conditional restart with false startstop");
```
ned21, line 450:
```suggestion
                    $self->verbose("Changed unitfile for $detail->{name} but service restart disabled via startstop configuration");
```

ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
stdweird: i ran into this because i want to change the unitfile of a service that on itself cannot be started, it's only handled because it is a dependecy of anothet unit.
however, we are now in the situation where i modify a unit file but it changes nothing until this other service that might be running is restarted (well, i think). systemctl daemon-reload does run ofcourse.

so these changes might only become active when we do something somehow.

ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
stdweird: ok, but would like to keep the conditional restart mentioned in it

ned21: Have edited my suggestion.

ncm-systemd/src/main/perl/Systemd/Service/Unit.pm
stdweird: i removed the comment. obviously when startstop is false, you can get into this situation

stdweird:
@jrha @ned21 there's a bugfix in here that we should release

PR #1341, opened 2018-11-05
ncm-metaconfig: switch to CAF::ServiceActions

stdweird:
Requires https://github.com/quattor/CAF/pull/267

PR #1454, opened 2020-06-12
ncm-download: support daemon actions on change

stdweird:
Requires https://github.com/quattor/template-library-core/pull/201

### DIFF
--- a/ncm-download/src/main/pan/components/download/schema.pan
+++ b/ncm-download/src/main/pan/components/download/schema.pan
@@ -9,8 +9,7 @@ type component_download_file = {
       used. Local files can be used as source, such as
       file://localhost/etc/foo.txt or even file:///etc/foo.txt.}
     "href"    : string
-    @{Specify the command (no options allowed) to run
-      whenever the file is updated.
+    @{Specify the command (no options allowed) to run after non-failed download attempt.
       The filename is added as first and (only) argument.
       Note that if the update is
       optimised away by the download process (e.g. if the file is
@@ -41,6 +40,14 @@ type component_download_file = {
     "timeout" ? long
     @{allow older remote file}
     "allow_older" ? boolean
+    @{A dict of daemonname => caf_serviceaction indicating what actions
+      should be applied to daemonname if the file changes. This allows,
+      for example, the restarting of a daemon upon file change.
+      Even if multiple services are associated to the same daemon, each action
+      for the daemon will be taken at most once.
+      If multiple actions are to be taken for the same daemon, all actions
+      will be taken (no attempt to optimize is made).}
+    'daemons' ? caf_serviceaction{}
 };
 
 type download_component = extensible {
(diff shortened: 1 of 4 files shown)


ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -9,8 +9,8 @@ type component_download_file = {
       used. Local files can be used as source, such as
       file://localhost/etc/foo.txt or even file:///etc/foo.txt.}
     "href"    : string
-    @{Specify the command (no options allowed) to run
-      whenever the file is updated.
+    @{Specify the command (no options allowed) to run after non-failed download attempt
+      (i.e. also in case the file is not updated).
```
ned21, line 13: This line seems redundant - it just repeats line 15 & 16 in a less clear way?

ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -41,6 +41,13 @@ type component_download_file = {
     "timeout" ? long
     @{allow older remote file}
     "allow_older" ? boolean
+    @{An dict with foreach daemon the CAF::Service action to take
```
ned21, line 44: How about: 

```
A dict of daemonname => caf_serviceaction indicating what actions
should be applied to daemonname if the file changes. This allows,
for example, the restarting of a daemon upon file change.
Even if multiple services are associated to the same daemon, each action
for the daemon will be taken at most once.
If multiple actions are to be taken for the same daemon, all actions
will be taken (no attempt to optimize is made).}
```

stdweird: ok, i''l also change it in metaconfig when i redo #1341 on top of #1451

stdweird:
test this please

stdweird:
test this please

stdweird:
@jrha tests pass now

ncm-download/src/main/pan/components/download/schema.pan
```diff
@@ -9,8 +9,7 @@ type component_download_file = {
       used. Local files can be used as source, such as
       file://localhost/etc/foo.txt or even file:///etc/foo.txt.}
     "href"    : string
-    @{Specify the command (no options allowed) to run
-      whenever the file is updated.
+    @{Specify the command (no options allowed) to run after non-failed download attempt.
```
ned21, line 12: non-failed being different to successful?

stdweird: the code runs the post after a succesful `download` method call. that call can be succesful when nothing was retrieved (eg when timestamp was recent enough). it can also fail when retrieval is ok, but setiing permissions was not. `file is updated` definitely is not correct.
not sure how to phrase it better

ned21: Makes sense!

ned21 (approved):

PR #1457, opened 2020-07-16
ncm-authconfig: sssd: add override_shell option

kwaegema:
add schema option for sssd.conf

### DIFF
--- a/ncm-authconfig/src/main/pan/components/authconfig/sssd.pan
+++ b/ncm-authconfig/src/main/pan/components/authconfig/sssd.pan
@@ -108,6 +108,7 @@ type authconfig_sssd_domain  = {
     "dns_resolver_timeout" : long = 5
     "dns_discovery_domain" ? string
     "override_gid" ? long
+    "override_shell" ? string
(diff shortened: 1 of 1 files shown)


ncm-authconfig/src/main/pan/components/authconfig/sssd.pan
```diff
@@ -108,6 +108,7 @@ type authconfig_sssd_domain  = {
     "dns_resolver_timeout" : long = 5
     "dns_discovery_domain" ? string
     "override_gid" ? long
+    "override_shell" ? string
```
wpoely86, line 111: `absolute_file_path` instead of `string`?

wpoely86: Shouldn't this also be under sssd_nss?

jrha: > override_shell (string)
    Override the login shell for all users. This option can be specified globally in the [nss] section or per-domain.
    Default: not set (SSSD will use the value retrieved from LDAP)

PR #1461, opened 2020-08-13
ncm-ceph: mon_initial_members are ceph-ids (short hostnames)

kwaegema:
this gets rid of the deprecated warning too


kwaegema:
part of #1463

PR #1462, opened 2020-08-14
ncm-ceph: remove unsupported code (Jewel)

### LINKED ISSUE #1418: ncm-ceph: Why are there two files called (C|c)ommands.pm in the same dir?
`ncm-ceph/src/main/perl/Ceph` contains both `Commands.pm` and `commands.pm`. This is confusing for the user but also causes problems when using the repo on an OS with case-insensitive filesystems (e.g. macOS).

kwaegema:
Jewel was maintained until June 2018, should not be in use anywhere :)

Also fixes #1418


kwaegema:
@ned21 yes, I added the flag, but this indeed forces to use the new schema (since the old will be unusable anyway)

ned21:
@jrha What's our current position on merging things? Looks like 20.3 still has 55 open issues so can I go ahead and merge this to fix #1418 ?

PR #1464, opened 2020-09-02
ncm-systemd: all pre and post exec commands now take a list instead of a string

wpoely86:
As already discussed in #1448, ExecStartPost, ExecStartPre and ExecStopPost can appear multiple times within a systemd unit. Until now they were string, I've changed it to list. This is a breaking change.

ExecStart takes 1 command unless it's a one-shot in which case a list is also allowed. I've not changed that as that as it would really break everything.

Requires quattor/template-library-core/pull/212.


wpoely86:
Maybe useful for other too but I'm not sure what the apprioate place is to add this:

if you want to run the unit tests easily, use the included Docker.

```
cd /path/to/this/repo
podman build -t quattor_testing -f Dockerfile .

podman run --user=$(id -ur):$(id -gr) -it --userns=keep-id -v $PWD:/quattor_test localhost/quattor_testing
```

Or to run specific tests:

```
podman run --user=$(id -ur):$(id -gr) -it --userns=keep-id -v $PWD:/quattor_test -e MVN_ARGS="-Dunittest=00-tqu.t -pl ncm-systemd" localhost/quattor_testing
```

ned21:
> Maybe useful for other too but I'm not sure what the apprioate place is to add this:
> 
> if you want to run the unit tests easily, use the included Docker.
> 
> ```
> cd /path/to/this/repo
> podman build -t quattor_testing -f Dockerfile .
> 
> podman run --user=$(id -ur):$(id -gr) -it --userns=keep-id -v $PWD:/quattor_test localhost/quattor_testing
> ```
> 
> Or to run specific tests:
> 
> ```
> podman run --user=$(id -ur):$(id -gr) -it --userns=keep-id -v $PWD:/quattor_test -e MVN_ARGS="-Dunittest=00-tqu.t -pl ncm-systemd" localhost/quattor_testing
> ```

Could you add this to https://github.com/quattor/quattor.github.com/blob/master/_development/00-basics.md please?  Point 7 currently just says "run the unit tests" so you could add this info on how to accomplish that there.

ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
```diff
@@ -52,6 +52,7 @@ bind "/unitfile" = systemd_unitfile_config[];
         'TTYVHangup', false,
         'LimitSTACK', -1,
         'LimitNPROC', 100,
+        'ExecStartPre', list('/usr/bin/true', '/bin/foo'),
```
ned21, line 55: It would be nice to include a command with a special character here, e.g. `-/bin/false`

```suggestion
        'ExecStartPre', list('/usr/bin/true', '-/bin/false'),
```

ned21 (dismissed):
In general seems fine, but we need to think carefully how we merge this given it's backwards incompatible.

jouvin:
I'm hardly following what happens with Quattor development but I didn't disappear! I understand the intent of this change but such a backward incompatible change will pose a lot of problems and probably result in people staying in the version before the change for a long time or forking the component... I see 2 ways forward, used in the past in similar situations:

- Create a new property that will be a list, keeping the other one (with may be a deprecation warning), exclusive from the current one, and have the component check boths (if the schema enforce that it is one or the other, it should not result in complex checs in the component code)
- Take advantage that a property in pan can have several types and in the component code, check if it is a string or a list and process it accordingly (used in several components). It makes the transition easier and the current way can also be marked as deprecated if I'm right. Something like:

```pan
'ExecStartPre' ? string with deprecated(...)
'ExecStartPre' ? string[]
```

wpoely86:
@jouvin how does this several type feature work?

```
object template a;

type testschema = {
    'foo' ? string
    'foo' ? string[]
};

bind '/test' = testschema;

'/test/foo' = 'bar';
```
gives me:
```
validation error [/tmp/a.pan]
validation requires type of 'list' but element is of type 'string'
element path: '/test/foo'
type: 'testschema' [/tmp/a.pan:3.19-6.1]
path '/test' bound to type testschema in [/tmp/a.pan:8.16-8.25]
```

jouvin:
I have to dig into my memory... or archives!!! I'd say the way you did it but it seems I'm wrong...

jrha:
Briefly you need to use a type higher in the [type heirarchy](https://quattor-pan.readthedocs.io/en/stable/pan-book/pan-book.html#type-hierarchy) and validate with a function:
```
object template a;

type testschema = {
    'foo' ? element with is_string(SELF) || is_list(SELF)
};

bind '/test' = testschema;

'/test/foo' = 'bar';
```
See also `transitional_yes_no_true_false` in https://github.com/quattor/template-library-core/blob/master/pan/legacy.pan

jrha:
I've added a new type for this in quattor/template-library-core#212.

wpoely86:
Updated to use quattor/template-library-core#212 but now the tests fail with\
```
hashFiles('**/pom.xml') failed. Fail to hash files under directory '/home/runner/work/configuration-modules-core/configuration-modules-core'
```

Not sure what is the issue with this line.

jrha (approved):
LGTM

jrha:
Thanks for working out what the hashing error was about, could you move the fix to a new PR?

wpoely86:
@jrha It isn't fixed, just avoided. For some reason, doing `${{ hashFiles('**/pom.xml') }}` fails on when doing the save of the cache (but not always for a restore). I can't figure out why. Maybe it's a wrong directory but that makes no sense at all.

wpoely86:
Finally good to go with passing tests!

jrha (approved):

PR #1465, opened 2020-09-08
ncm-network: cleanup network interfaces with partition info in the name

stdweird:


jrha:
Clean-up? Doesn't this just define them as valid names?

stdweird:
it also removes any existing files with such naming

jrha:
Oh I see, cool.

jrha (approved):

PR #1466, opened 2020-09-15
ncm-openvpn: Modified to use generic "service" command rather than init.d script

Gargravarr2112:
Since EL7+ uses systemd, calling /etc/init.d scripts is no longer supported for service control. The generic command 'service' works on either sysvinit or systemd, so is backwards and forwards compatible.

### DIFF
--- a/ncm-openvpn/src/main/perl/openvpn.pm
+++ b/ncm-openvpn/src/main/perl/openvpn.pm
@@ -115,7 +115,7 @@ sub Configure
 
     if ($changed) {
 	$self->verbose("Restarting OpenVPN daemon");
-	CAF::Process->new([qw(/etc/init.d/openvpn restart)],
+	CAF::Process->new([qw(/sbin/service openvpn restart)],
 			  log => $self)->run();
 	return !$?;
     }


ncm-openvpn/src/main/perl/openvpn.pm
```diff
@@ -115,7 +115,7 @@ sub Configure
 
     if ($changed) {
 	$self->verbose("Restarting OpenVPN daemon");
-	CAF::Process->new([qw(/etc/init.d/openvpn restart)],
+	CAF::Process->new([qw(service openvpn restart)],
```
ned21, line 118: The real fix here is to move to CAF::Service which handles systemd natively, but as a quick fix I don't have a problem with this. I think an absolute path might be a better idea rather than relying on $PATH though?

ned21:
@hpcugentbot Test this please

ned21 (approved):

ned21:
Sorry, just realised I didn't read the commit message last night. Our standard is to put the component name as the first word of the first line to make it easy to automatically generate readable release notes. Could you please squash the two commits into 1, and then re-word the commit log to say:

```
openvpn: Use "service" command rather than init.d script

This allows the component to work on EL6 and EL7+.
```

ned21 (approved):

PR #1428, opened 2020-02-21
ncm-shorewall: add support for providers and rtrules configuration

stdweird:

PR #1467, opened 2020-10-09
ncm-shorewall: add support for snat and deprecate masq

kwaegema:
Based on #1428

### DIFF
--- a/ncm-shorewall/src/main/pan/components/shorewall/schema.pan
+++ b/ncm-shorewall/src/main/pan/components/shorewall/schema.pan
@@ -22,7 +22,21 @@ type component_shorewall_masq = {
     "origdest" ? string
     "probability" ? double(0..1)
 };
-
+# Keep this list in sync with list from TT file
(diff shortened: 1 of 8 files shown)


ncm-shorewall/src/main/pan/components/shorewall/schema.pan
```diff
@@ -22,7 +22,21 @@ type component_shorewall_masq = {
...
+    "action" : string with match(SELF, '^(CONTINUE|LOG|MASQUERADE|NFLOG|SNAT|ULOG)') # can have trailing parameters
+    "source" ? string
+    "dest" ? string[]
+    "proto" ? string
+    "port" ? string[]
+    "ipsec" ? string[]
+    "mark" ? string
+    "user" ? string
+    "switch" ? string
+    "origdest" ? string
```
jrha, line 37: Could any of these `string` types be restricted further?
Perhaps just by using `string_trimmed` to prevent leading or trailing white-space?

jrha:
It might be nice if the commit message explained why snat was being introduced and masq was being deprecated, I assume there is at least a Shorewall announcement/change log on the subject?

jrha (approved):
LGTM
Would be nice to tighten up the types, but not going force it.

PR #1437, opened 2020-03-30
ncm-ntpd: Add support for the 'interface' directive

gombasg:
See the title.

PR #1470, opened 2020-10-12
ncm-ntpd: Add support for the 'interface' directive

ned21:
Supersedes #1437 which was messed up by trying to use the webUI to make minor edits.

### DIFF
--- a/ncm-ntpd/src/main/pan/components/ntpd/schema.pan
+++ b/ncm-ntpd/src/main/pan/components/ntpd/schema.pan
@@ -96,6 +96,14 @@ type ntpd_system_options = {
     "stats" ? boolean
 };
 
+@documentation{
+    Defines a single rule for the "interface" statement. See the ntp_misc manpage.
+}
+type ntpd_interface_options = {
+    "action" : choice("listen", "ignore", "drop")
+    "match" : string
+};
+
 # logging configuration
 function valid_ntpd_logconfig_list = {
     # Loop over each component.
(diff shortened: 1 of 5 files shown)


ncm-ntpd/src/main/pan/components/ntpd/schema.pan
```diff
@@ -96,6 +96,14 @@ type ntpd_system_options = {
     "stats" ? boolean
 };
 
+@documentation{
+    Defines a single rule for the "interface" statement. See the ntp_misc manpage.
+}
+type ntpd_interface_options = {
+    "action" : choice("listen", "ignore", "drop")
+    "match" : string
```
jrha, line 104: So I read the [ntp_misc manpage](https://www.unix.com/man-page/centos/5/ntp_misc/), the match part seems to have some pretty specific restrictions which are not at all covered by `string`:
>     interface [listen | ignore | drop] [all | ipv4 | ipv6 | wildcard | name | address[/prefixlen]]

What is not clear is which are values and which are placeholders, and then `name` appears to mean network interface name and that is a whole can of worms, but maybe something like the following would cover it?
```suggestion
    "match" : string_trimmed with {
        match(SELF, 'all|ipv4|ipv6|wildcard') || is_ip || is_ipv4_netmask_pair(SELF) || is_ipv6_network_block(SELF) || path_exists(format('/system/network/interfaces/%s', SELF));
    }
```

jrha: Who needs documentation when you have source code?
https://github.com/ntp-project/ntp/blob/9c75327c3796ff59ac648478cd4da8b205bceb77/ntpd/complete.conf.in#L57-L63

jrha: Opened an issue (#1481) to look at this in future.

jrha (approved):

PR #1473, opened 2020-10-27
ncm-metaconfig: add chrony service

kwaegema:
chrony service for chrony.conf


ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -0,0 +1,85 @@
...
+include 'pan/types';
+
+type chrony_service_server_flag = choice(
+    'iburst', 'burst', 'offline', 'auto_offline', 'prefer', 'noselect', 'trust', 'require', 'xleave');
+
+@documentation{
+    the crony.conf configuration entry for server/pool/peer options
+}
+type chrony_service_server_options = {
+    'minpoll' ? long = -6 # power of 2 (1/64th of a second)
```
ned21, line 12: If these are not mandatory then would it be better to leave them without defaults and thus allow the upstream default to take precedence? I think we've had issues in the past where we've set a default for a piece of software, only for a later version of RHEL/SL to set a different default which we've then not picked up.

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -0,0 +1,85 @@
...
+    'offset' ? double
+    'minsamples' ? long(0..)
+    'maxsamples' ? long(0..)
+    'maxsources' ? long = 4
+    'filter' ? long(0..)
+    'polltarget' ? long(6..60) = 8
+    'port' ? type_port
+    'presend' ? long(0..)
+    'minstratum' ? long
+    'version' ? choice('3', '4') = 4
```
ned21, line 29: NB, default value has not effect for an optional field unless it is explicitly set to undef.

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -0,0 +1,85 @@
...
+    Normally chronyd will cause the system to gradually correct any time offset,
+    by slowing down or speeding up the clock as required.
+    In certain situations, the system clock might be so far adrift that this slewing process
+    would take a very long time to correct the system clock.
+
+    This directive forces chronyd to step the system clock if the adjustment is larger than a threshold value,
+    but only if there were no more clock updates since chronyd was started than a specified limit
+    (a negative value can be used to disable the limit)
+}
+
```
ned21, line 53: Drop blank line here to make sure annotation is attached to the type

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -0,0 +1,85 @@
...
+type chrony_service_makestep = {
+    'threshold' : double(0..)
+    'limit' : long
+};
+
+@documentation{
+    The allow/deny directive is used to designate a particular subnet from which NTP clients
+    are allowed/denied to access the computer as an NTP server.
+}
+
```
ned21, line 63: Remove blank line

ned21 (commented):
Looks good, aside from a couple of comments about default values and some spacing issues.

Anyone know why the unit test is failing? Looks unrelated.

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -0,0 +1,85 @@
...
+type chrony_service_network = {
+    'action' : choice('allow', 'deny')
+    'host' : type_network_name
+};
+
+@documentation{
+    the crony.conf configuration
+}
+type chrony_service = {
+    'server' ? chrony_service_server[]
```
stdweird, line 73: how much of all this is really optional?

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
kwaegema: Ah I didn't know default value for optional fields have ANY effect, so I used it merely as documentation/comment, will change this

ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
kwaegema: for use in isolated networks (computers that never have network conectivity to any computer which ultimately derives its time from a reference clock) the master does not have any of these, for all other cases at least one of server/pool/peer should be used

jrha:
Rebased and squashed.

jrha (approved):
LGTM

Like most other PRs against this repository, the Docker tests are failing with:
> maven-clean-plugin:2.4.1 or one of its dependencies could not be resolved
The actual component tests are passing without issue.

ned21:
```
020-12-17T10:56:31.4305136Z [ERROR] Plugin org.apache.maven.plugins:maven-clean-plugin:2.4.1 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-clean-plugin:jar:2.4.1: Could not transfer artifact org.apache.maven.plugins:maven-clean-plugin:pom:2.4.1 from/to central (https://repo.maven.apache.org/maven2): Connection timed out (Read failed) -> [Help 1]
2020-12-17T10:56:31.4308840Z [ERROR] 
```

That's not good. Do we need to run `clean` under Docker since the environment is built afresh each time. I can't remember where the source for `/usr/bin/mvn_test.sh` is kept. :(

jrha:
It's very odd, I can't reproduce it if I build the Dockerfile on my machine, it also successfully fetches a never version of `maven-clean-plugin` earlier on in the build.

`mvn_test.sh` currently comes from [perl-Test-Quattor](http://yum.quattor.org/devel/perl-Test-Quattor-18.3.0-SNAPSHOT20180406083650.noarch.rpm) but there is [a PR open](https://github.com/quattor/maven-tools/pull/175) to add it to maven-tools.

PR #1484, opened 2021-01-27
Replace travis checks with github action

wpoely86:
As travis is shutting down in the near future, let the panlint check also run as a github action.


.github/workflows/continuous-integration.yaml
```diff
@@ -1,10 +1,6 @@
 name: Run Tests using Docker
 
-on:
-  push:
-    branches: [ github-action ]
-  pull_request:
-    branches: [ master ]
```
wpoely86, line 7: @ned21 is there any reason why you blocked gh actions of running on all branches?

Witht this setup, you cannot run gh actions on your own fork? I've made it general.

ned21: This was my first GH action so I think it was out of an abundance of caution not to break anything or annoy anyone until it was working! Thanks for updating it now.

ned21 (dismissed):

wpoely86:
The github actions tests fail all the time with a timeout:

```
2021-01-27T20:13:07.8452192Z [ERROR] Plugin org.apache.maven.plugins:maven-clean-plugin:2.4.1 or one of its dependencies could not be resolved: Failed to read artifact descriptor for org.apache.maven.plugins:maven-clean-plugin:jar:2.4.1: Could not transfer artifact org.apache.maven.plugins:maven-clean-plugin:pom:2.4.1 from/to central (https://repo.maven.apache.org/maven2): Connection timed out (Read failed) -> [Help 1]
```

It already has a timeout of 15 minutes so not sure what can be done...

wpoely86:
I've reworked the github actions completely. They now directly run instead of building a container and then run the container. It now caches the maven cache which speeds up everything considerably (from 38 to 20 minutes). And the test now actually pass instead of failing on a maven download. @ned21 can you have a look and merge it? All new PR currently fail with maven downloads, this should fix it.

ned21 (approved):
Thanks for getting to the bottom of this problem! 🥇

ned21:
@jrha This will disable travis completely so are you happy with this before we merge it?

ned21:
I've updated the branch protection rules to require Ugent Jenkins and the GitHub actions and removed the requirement for travis. There's a lot of PRs backing up with checks failures that this would fix so I think this should get merged; my only reservation is that it might be nice to run this side-by-side with travis for a bit to make sure it's not missing something?

wpoely86:
The only thing that ran on travis was panlint, not the actual tests.

ned21:
@wpoely86 I merged it but now it's failing? https://github.com/quattor/configuration-modules-core/actions/runs/623827644

```
2021-03-05T09:05:33.7835590Z [INFO] Node Configuration Manager - components ........... FAILURE [15:51.992s]
```

but I can't find the "not ok" line that causes this.

wpoely86:
@ned21 it's the same issue as before as it seem to have a different cache for the master branch... It fails on:

```
2021-03-05T08:49:41.7748565Z Downloading: https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-clean-plugin/2.4.1/maven-clean-plugin-2.4.1.pom
```

And the cache is not saved on a failed action. The only thing I can think of is to let the tests run on a limited set of tests so we have a cache? Add `-pl metaconfig` or so to `MVN_ARGS` for one run?

ned21:
Makes sense. So that would require a PR to change tests to be just one test, merge it, let it run, then add a new PR to revert back to all?

wpoely86:
Yes, I'm doing that right now in https://github.com/quattor/configuration-modules-core/pull/1493
There is already a cache. Can you rerun the tests on master too?

ned21:
Yep, found how to rerun - rerunning now: https://github.com/quattor/configuration-modules-core/runs/2040122813

jrha:
> @jrha This will disable travis completely so are you happy with this before we merge it?
Too late, but yes, happy.

PR #1463, opened 2020-08-18
ncm-ceph: add support for centralized configuration management

kwaegema:
builds on and requires #1462 
Implementing configurating ceph using centralized config db:
https://ceph.io/community/new-mimic-centralized-configuration-management/

PR #1490, opened 2021-02-17
ncm-ceph: Add support for orchestrator/cephadm under octopus

kwaegema:
Added support for orchestrator with cepadm. This is yaml config file based. The component will still write a minimal ceph.conf, set config options in the ceph config database, and write and deploy orchestrator yaml-files.

Replaces https://github.com/quattor/configuration-modules-core/pull/1463 for configdb, which PR i will close in favor of this one.

configdb code was implemented under the (legacy) Luminous code in that PR, and I added it now also under the orchestrator code part, since I'll get rid of the Luminous code path when Luminus - Nautilus is not relevant anymore. But since this code was still not merged into master, i guess i can remove it as soon we're not using it anymore (probably June)


kwaegema:
Also needs EDG::WP4::CCM::TextRender version v20.12.1

jrha (approved):
LGTM
The panlint failures are existing missing component includes in tests and not related to this change.

PR #1493, opened 2021-03-04
ncm-metaconfig: add support for slurm 20.11

wpoely86:
This builds on top of #1422 and add/changes a couple of things so it works with slurm 20.11

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/slurm/pan/dbd.pan
+++ b/ncm-metaconfig/src/main/metaconfig/slurm/pan/dbd.pan
@@ -6,8 +6,8 @@ include 'metaconfig/slurm/schema';
 bind "/software/components/metaconfig/services/{/etc/slurm/slurmdbd.conf}/contents" = slurm_dbd_conf;
 
 prefix "/software/components/metaconfig/services/{/etc/slurm/slurmdbd.conf}";
-"owner" = "root";
-"group" = "root";
-"mode" = 0644;
+"owner" = "slurm";
+"group" = "slurm";
+"mode" = 0600;
 "module" = "slurm/dbd";
 "daemons/slurmdbd" = "restart";
--- a/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
+++ b/ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
@@ -243,7 +243,7 @@ type slurm_conf_control = {
     'GroupUpdateForce' ? boolean # 0/1
     'GroupUpdateTime' ? long(0..)
     'JobCheckpointDir' ? absolute_file_path
-    'JobContainerType' ? choice('cncu', 'none')
+    'JobContainerType' ? choice('cncu', 'tmpfs', 'none')
     'JobCredentialPrivateKey' ? absolute_file_path
     'JobCredentialPublicCertificate' ? absolute_file_path
     'JobFileAppend' ? boolean # 0/1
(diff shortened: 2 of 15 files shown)


wpoely86:
@stdweird Are there issue with Jenkins? I'm not getting a pass on any test (gh action is fine).

ncm-metaconfig/src/main/metaconfig/slurm/pan/dbd.pan
```diff
@@ -6,8 +6,8 @@ include 'metaconfig/slurm/schema';
 bind "/software/components/metaconfig/services/{/etc/slurm/slurmdbd.conf}/contents" = slurm_dbd_conf;
 
 prefix "/software/components/metaconfig/services/{/etc/slurm/slurmdbd.conf}";
-"owner" = "root";
-"group" = "root";
-"mode" = 0644;
+"owner" = "slurm";
+"group" = "slurm";
+"mode" = 0600;
```
jrha, line 11: Is this backwards compatible?

wpoely86: yes, the slurmdbd daemon already ran as the slurm user. They just made it more strict with this version: the config file has to be closed off (as it contains the mysql credentials to access the db).

jrha: Makes sense, thanks!

ncm-metaconfig/src/main/metaconfig/slurm/pan/schema.pan
```diff
@@ -86,7 +86,7 @@ type slurm_scheduler_parameters = {
     'bf_yield_interval' ? long(0..)
     'bf_yield_sleep' ? long(0..)
     'build_queue_timeout' ? long(0..)
-    'default_queue_depth' ? long(0..)
+    '{default_queue_depth}' ? long(0..)
```
jrha, line 89: Why are these keys being escaped in the schema?

wpoely86: Have a look at the discussion in #1422

wpoely86: I'm actually not 100% sure what the issue exactly is. @itkovian ?

jrha: Ah yes, I don't understand it either!

stdweird: i think  all keys are unescaped, but this means this generates invalid code unless escaped

wpoely86: When I drop the `{ }`, it fails with:

```
# -SchedulerParameters=bf_continue,bf_max_job_test=1024,bf_window=4320,default_queue�pth=128
# +SchedulerParameters=bf_continue,bf_max_job_test=1024,bf_window=4320,default_queue_depth=128
```

It smells like a bug? When doesn't this happen with other keys? For example `preempt_reorder_count` produces the correct result without `{ }`.

stdweird: this is not a bug, this the crappy unneeded escaping garbage in quattor. the escape function is so dumb, you get crap like this. anything that is `_[a-f][a-f]` can be unescaped, so you have to be really careful with simply escaping all because of the rare cornercases like ,_debug`, `_depth`, `_action` and a few others we encountered over the years.

wpoely86: OK but when I flip
```
"convert/unescapekey" = true;
```
to `false` and drop the `{ }` it gives the correct result. But that would revert #1422 which I guess was opened for good reason?

stdweird: there are some keys here and there that do require escaping since they contain a `/`. 
imho, this weirdness is a small price to pay. the alternative is to write much more complicated TT files.

and all this for some bad design decision of over 10years old, really not worth it.

wpoely86: OK, then we keep it as-is.

Other remarks @jrha ?

jrha: Ok I finally get it, we can probably live with it.

jrha (approved):

PR #1495, opened 2021-03-06
gh action test: first download all maven dependencies

wpoely86:
For reason unknown to me, downloading maven dependencies can fail if you run it with github actions. First downloading all dependencies and then running the tests does seem to work consistently.

@ned21 this works (so far) all the time for me. The caching with github action doesn't seem to be 100% reliable. Sometimes it gets a hit, sometimes not.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -43,7 +43,9 @@ jobs:
         # make sure it exists before chown
         mkdir -p /tmp/m2
         chown -R nobody:nobody . /tmp/m2
-        su --shell /bin/bash --preserve-environment --command="source /usr/bin/mvn_test.sh && mvn_test" nobody
+        # we have to run as a non-root user to pass the spma tests
+        # secondly, we first download all maven dependencies and then run the tests because it fails with hanging downloads otherwise.
+        runuser --shell /bin/bash --preserve-environment --command "source /usr/bin/mvn_test.sh && mvn_run \"dependency:resolve $MVN_ARGS\" && mvn_test" nobody
       env:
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -43,7 +43,9 @@ jobs:
         # make sure it exists before chown
         mkdir -p /tmp/m2
         chown -R nobody:nobody . /tmp/m2
-        su --shell /bin/bash --preserve-environment --command="source /usr/bin/mvn_test.sh && mvn_test" nobody
+        # we have to run as a non-root user to pass the spma tests
+        # secondly, we first download all maven dependencies and then run the tests because it fails with hanging downloads otherwise.
+        su --shell /bin/bash --preserve-environment --command="source /usr/bin/mvn_test.sh && mvn_run \"dependency:resolve $MVN_ARGS\" && mvn_test" nobody
```
ned21, line 48: Is this the best way to do this? From the `su(1)` man page it suggests runuser is better?

       su is mostly designed for unprivileged users, the recommended solution
       for privileged users (e.g. scripts executed by root) is to use non-set-
       user-ID  command  runuser(1)  that  does not require authentication and
       provide separate PAM configuration.

ned21 (approved):
Thanks for continuing to try and make this work. My guess is that we're hitting some sort of anti-DoS tarpit on the server that's marking the traffic as suspicious, so anything we can do to reduce downloads and cache locally will help.

.github/workflows/continuous-integration.yaml
wpoely86: I've switched to `runuser`

ned21 (approved):

ned21:
Merged 🤞

PR #1500, opened 2021-04-05
ncm-systemd: Add support for mount, automount and timer units

wpoely86:
Add support for:
-  `[Mount]` and `.mount` units
-  `[Automount]` and `.automount` units
-  `[Timer]` and `.timer` units

in ncm-systemd.


ned21 (approved):

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -321,6 +321,57 @@ type ${project.artifactId}_unitfile_config_socket = {
...
+the [Timer] section
+http://www.freedesktop.org/software/systemd/man/systemd.timer.html
+}
+type ${project.artifactId}_unitfile_config_timer = {
+    'OnActiveSec' ? long(0..)
+    'OnBootSec' ? long(0..)
+    'OnStartupSec' ? long(0..)
+    'OnUnitActiveSec' ? long(0..)
+    'OnUnitInactiveSec' ? long(0..)
+    'OnCalendar' ? string[]
```
wpoely86, line 363: I'm not sure what to do with this one. It supports a lot of different syntax: https://www.freedesktop.org/software/systemd/man/systemd.time.html#

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -375,7 +429,7 @@ type ${project.artifactId}_target = string with match(SELF, "^(default|poweroff|
 type ${project.artifactId}_unit_type = {
     "name" ? string # shortnames are ok; fullnames require matching type
```
wpoely86, line 430: Regarding this comment, is there a reason why it's not enforced by validation @stdweird ?

stdweird: because it's easier not to? feel free to add it, shouldn't be that hard

wpoely86: I'm more thinking about, if you have `data-mnt.mount` as filename, the `type` should match and if you have a file, it should have `mount` section.

wpoely86: I'm actually not sure what the `name` entry here does?

stdweird: i guess it forces the name?

ncm-systemd/src/main/pan/components/systemd/schema.pan
jrha: Oh wow.

Either leave it as is, or pick a subset of the supported syntax to validate.

ncm-systemd/src/main/pan/components/systemd/schema.pan
jrha: Are you using it?

ncm-systemd/src/main/pan/components/systemd/schema.pan
wpoely86: OK, leave as string.

ncm-systemd/src/main/pan/components/systemd/schema.pan
wpoely86: No

jrha: Probably best to remove it then.

wpoely86: OK but that might break templates for some people? Better deprecated it first?

stdweird: why remove it? adding the type check should be easy to do.

jrha: Sorry I misread the PR and thought it added it, it can be left as is then.

jrha (approved):

PR #1502, opened 2021-04-14
ncm-hostsfile: Cleanup and tests

jrha:
Add test for working IPv4 only behaviour.


ncm-hostsfile/src/main/perl/hostsfile.pm
```diff
@@ -2,31 +2,137 @@
...
+created using an FQDN formed by joining the shortname with
+this domain.
+
+=item * C<< /software/components/hostsfile/file >>
+
+The filename to modify, defaults to C<< /etc/hosts >>.
+
+=item * C<< /software/components/hostsfile/entries >>
+
+An nlist, keyed by hostname. The value of each hostname is an nlist
```
ned21, line 48: nlist => dict ?

ncm-hostsfile/src/main/perl/hostsfile.pm
```diff
@@ -2,31 +2,137 @@
...
+=cut
+
+package NCM::Component::hostsfile;
 
 use strict;
 use base 'NCM::Component';
 
 our $EC = LC::Exception::Context->new->will_store_all;
-our $NoActionSupported = 1;
+our $NoActionSupported = 0;
```
ned21, line 130: AFAIK LC works fine with NoAction (CAF just calls through to them) so I don't think it's necessary to disable this unless we know there's a piece of code that is not safe?

jrha: My mistake!

jrha:
I could do with some more examples of how this component is used in the wild to build more test cases, right now it looks to me like the behaviour of `takeover` doesn't match the documentation.

ned21:
Looks like we use takeover=true. What sort of examples would you like? Looks like our default profile is:

```
+-/software/components/hostsfile
   +-entries
      +-<<fqdn>>
         $ ipaddr : "<<ipaddr>>"
      +-localhost
         $ ipaddr : '127.0.0.1'
   $ takeover: true
```

stdweird:
test this please

ncm-hostsfile/src/main/perl/hostsfile.pm
```diff
@@ -2,18 +2,126 @@
...
+
+Returns error in case of a failure.
+
+=back
+
+=head1 RESOURCES
+
+=over 4
+
+=item * C<< /system/network/domainname >>
```
stdweird, line 35: i prefer to document the schema instead of the component.

ncm-hostsfile/src/main/pan/components/hostsfile/schema.pan
```diff
@@ -14,10 +14,21 @@ declaration template components/hostsfile/schema;
...
+    "comment" ? string_trimmed
+};
+
 type component_hostsfile_type = {
     include structure_component
     "file" ? string        # File to store in.  Default is /etc/hosts
-    "entries" : dict
+    "entries" : hostsfile_entry{} with {
+        foreach (k; v; SELF) {
+            if (! is_fqdn(k)) error('Invalid FQDN "%s" for hostfile entry', k);
```
ned21: We add localhost to our entries - would that pass the is_fqdn() check?

```
+-entries
  +-localhost
  $ ipaddr : '127.0.0.1'
```

I don't think we can remove that now either because all of our localhost entries now have the # NCM suffix.

ncm-hostsfile/src/main/pan/components/hostsfile/schema.pan
```diff
@@ -1,23 +1,81 @@
...
+    "comment" ? string_trimmed
+};
+
+@documentation{
+    Example - configuration defined like this:
+
+      include 'software/components/hostsfile';
+      "/software/components/hostsfile/active" = true;
+      "/software/components/hostsfile/file" = "/etc/hosts.local";
+      "/software/components/hostsfile/entries" = nlist (
```
ned21, line 23: nlist => dict

ned21 (changes requested):
Can you add a test for takeover=true using the sample profile I provided above please?

I've added backwards incompatible because this removes short name support, but looking at our /etc/hosts file, I believe we are still using it. Can you restore that functionality please?

ncm-hostsfile/src/main/pan/components/hostsfile/schema.pan
jrha: whoops, thanks

ncm-hostsfile/src/main/pan/components/hostsfile/schema.pan
jrha: ah ok, done

jrha:
> Can you add a test for takeover=true using the sample profile I provided above please?
> 
> I've added backwards incompatible because this removes short name support, but looking at our /etc/hosts file, I believe we are still using it. Can you restore that functionality please?

No problem, sample profile added and functionality restored.

jrha:
I think leaving the backwards incompatible flag in place is justified as the schema now has some type checking in place.

ned21 (approved):

PR #1505, opened 2021-04-27
ncm-grub: set grubby rpm dependency

stdweird:


jrha (approved):
LGTM, the package exists for all the OS flavours we currently support

PR #1507, opened 2021-05-02
[WIP] ncm-spma: yumdnf: use spmaleaves dnf plugin

stdweird:
Requires #1401

### DIFF
--- a/ncm-spma/src/main/perl/spma/yumdnf.pm
+++ b/ncm-spma/src/main/perl/spma/yumdnf.pm
@@ -20,7 +20,6 @@ use constant REPO_INCLUDE => 0;
 
 use constant MODULES_TREE => "/software/modules";
 
-
 # Completes any pending transactions
 sub _do_complete_transaction
 {
--- a/ncm-spma/src/main/perl/spma/yumdnf.pm
+++ b/ncm-spma/src/main/perl/spma/yumdnf.pm
@@ -4,6 +4,7 @@ use warnings;
 package NCM::Component::spma::yumdnf;
 
 use NCM::Component::spma::yum;
+use Set::Scalar;
 push our @ISA , qw(NCM::Component::spma::yum);
 
 use constant YUM_CONF_FILE => "/etc/dnf/dnf.conf";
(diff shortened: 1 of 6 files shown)


stdweird:
@jrha @ned21 i don't actually expect this gets merged ever (it ships a piece of GPL code). but this is my current attempt, not sure how deal with this (it's annoying i can't make el8 rpms yet, so i can ship this in a seperate rpm and have it installed via a rpm dependency).

ncm-spma/src/main/perl/spma/yumdnf.pm
```diff
@@ -8,6 +8,7 @@ use Set::Scalar;
 push our @ISA , qw(NCM::Component::spma::yum);
 
 use constant YUM_CONF_FILE => "/etc/dnf/dnf.conf";
+use constant YUM_CONF_CLEANUP_ON_REMOVE_VALUE => 0;
```
ned21, line 11: Why not?

ned21:
> @jrha @ned21 i don't actually expect this gets merged ever (it ships a piece of GPL code). but this is my current attempt, not sure how deal with this (it's annoying i can't make el8 rpms yet, so i can ship this in a seperate rpm and have it installed via a rpm dependency).

Sounds like the right approach - nice to see this functionality being maintained by someone other than us - but it appears to be part of RH maintained dnf-plugins-core so why would we need to make our own rpm rather than getting it from OS?

stdweird:
@ned21 yes and no. most of the code is the same, but the output is different from the leaves plugin. also, the el8 rpm for dnf-plugins-core does not include the leaves plugin at all (and it looks intentional).

ncm-spma/src/main/perl/spma/yumdnf.pm
stdweird: this is potentionally dangerous; and i was surprised it was enabled on yum.

the usual scenario is that a (installed) leave has a  (installed) dependency on something you have listed in the packages list (but the leave itself is not listed; only the dependency). the current code will generate a remove command for the leave, and no install command for the dependency (it's already nstalled). autoremove will remove the leave and the dependency; and you will end up with something you didn't ask for.

ned21:
> @ned21 yes and no. most of the code is the same, but the output is different from the leaves plugin. also, the el8 rpm for dnf-plugins-core does not include the leaves plugin at all (and it looks intentional).

That's a bit odd. Can you ping me some details on where you seeing it's missing and why you think it's intentional and maybe I can raise a support case asking them to add it? (Or do we need a different plugin of the same name? That's confusing... But again maybe we can ask for RH about it.)

ncm-spma/src/main/perl/spma/yumdnf.pm
ned21: OK, makes sense - thanks,

jrha:
As you stated that you have no expectation for this to ever be merged I've converted this PR to a draft.

jrha:
> i don't actually expect this gets merged ever (it ships a piece of GPL code)

We could of course ship the plugin as a separate RPM in the externals repo and have ncm-spma depend upon it.

wpoely86:
> > i don't actually expect this gets merged ever (it ships a piece of GPL code)
> 
> We could of course ship the plugin as a separate RPM in the externals repo and have ncm-spma depend upon it.

That sounds like an excellent idea to me. It's just this one python file (the leaves plugin itself) which is slightly adjusted from upstream if I understand it correctly? Can we make a dedicated repo for the plugin and spec file? Or do we have some other place to put it?

jrha:
I'd be tempted to try and get the now very small changes merged upstream if possible.

StephaneGerardVUB:
I've tried out this PR/WIP, and there are problems when dealing with several architectures of a same package. For example, if the package glibc-gconv-extra is installed with 2 architectures (i686 and x86_64), spma will remove both as they are not declared in /software/packages and userpkgs is set to 'no', but doing so, it also removes all Quattor and perl packages. The full story is the following: glibc-gconv-extra is a dependency of redhat-rpm-extra that is itself a dependency of perl (via perl-devel). You can remove one arch of glibc-gconv-extra, but not both.

PR #1511, opened 2021-06-02
ncm-pam: Use component config header macro

jrha:
This fixes missing the dependency on spma and support for site-config.

As highlighted in quattor/ncm-ncd#130.

### DIFF
--- a/ncm-pam/src/main/pan/components/pam/config.pan
+++ b/ncm-pam/src/main/pan/components/pam/config.pan
@@ -1,24 +1,8 @@
-# ${license-info}
-# ${developer-info}
-# ${author-info}
-# ${build-info}
-
-##########################################################################
-# Coding style: emulate <TAB> characters with 4 spaces, thanks!
-##########################################################################
-
-unique template components/${project.artifactId}/config;
-
-include 'components/pam/schema';
+${componentconfig}
 
 # standard component settings
-"/software/components/pam/version"    = '${no-snapshot-version}';
-"/software/components/pam/active"    ?=  true;
-"/software/components/pam/dispatch"  ?=  true;
 "/software/components/pam/directory" ?= "/etc/pam.d";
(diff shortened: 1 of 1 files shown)


ncm-pam/src/main/pan/components/pam/config.pan
```diff
@@ -1,25 +1,9 @@
...
-unique template components/${project.artifactId}/config;
-
-include 'components/pam/schema';
+${componentconfig}
 
 # standard component settings
-"/software/components/pam/version"    = '${no-snapshot-version}';
-"/software/components/pam/active"    ?=  true;
-"/software/components/pam/dispatch"  ?=  true;
 "/software/components/pam/directory" ?= "/etc/pam.d";
```
stdweird, line 4: the `{componentconfig}` ends with a relative prefix to the compont path, so you can replace this with relative paths like
```
"directory" ? = ...
```

jrha: I suppose, it could be a bit fragile if it changes though.

ncm-pam/src/main/pan/components/pam/config.pan
```diff
@@ -114,7 +114,7 @@ function pam_add_listfile_acl = {
     ret = pam_add(service, pamtype, control, "listfile", opts);
     # Now, grab the entry that was just put at the end of the list and
     # add in the ACL information.
-    ret[service][pamtype][length(ret[service][pamtype])-1][sense] = dict("filename", filename, "items", items);
+    ret[service][pamtype][length(ret[service][pamtype]) - 1][sense] = dict("filename", filename, "items", items);
```
stdweird, line 117: didn;t we add support for negative indices in pan a while back?

jrha: So we did!

ned21 (approved):

PR #1519, opened 2021-08-09
ncm-metaconfig: postfix - add missing smtp_tls_protocols option

wdpypere:


wdpypere:
panlint fails on:

```
ncm-postfix/src/main/pan/components/postfix/schema.pan:69: Global variables should be five or more characters
final variable DAYS = HOURS * 24;
               ^^^^

1 problems found in total
Error: Process completed with exit code 1.
```

Should I fix this? It seems panlint is a bit too aggressive here.

ncm-postfix/src/main/pan/components/postfix/schema.pan
```diff
@@ -462,6 +462,7 @@ type postfix_main = {
     "smtp_tls_note_starttls_offer" ? boolean
     "smtp_tls_per_site" ? string
     "smtp_tls_policy_maps" ? string
+    "smtp_tls_protocols" ? string
```
ned21, line 465: It would be nice to validate this, although I suspect the list would require regular maintenance as valid protocols are added and removed?

ned21:
@wdpypere - I am confused as to why those globals are even defined here. They don't appear to be used in config.pan, and they are not namespaced to the component so could easily conflict with similar definitions in more general code. If they need to be defined they belong in another repo, so I would vote to remove them rather than rename them.

@stdweird - I saw your revert PR. I will defer to @jrha  but I think I'm OK with this code so unless someone else has a major issue then I think it would have been merged anyway.

wdpypere:
@ned21 I removed the global variables from the schema in our tree and indeed it does not seem to affect anything. I will make a separate PR removing them.

jrha:
I agree that Panlint is probably too aggressive in this case, but as we're removing them anyway…

PR #1521, opened 2021-08-11
ncm-postfix: remove misplaced unused global variables

wdpypere:
As discussed in https://github.com/quattor/configuration-modules-core/pull/1519 the globals defined in the postfix schema should either be in another repo, or they should be removed. As I can't find any use of them in our tree, I'm removing them. However, I don't know if they are used in other sites.


ned21 (approved):

ned21:
LGTM. I have added the backwards compatible label to warn end users. I would merge it but I'm not sure where we are in the release cycle.

PR #1530, opened 2021-10-28
ncm-nfs: add lustre support

stdweird:


ncm-nfs/src/main/pan/components/nfs/schema.pan
```diff
@@ -10,7 +10,7 @@ type structure_nfs_exports = {
 type structure_nfs_mounts = {
     'device' : string
     'mountpoint' : string
-    'fstype' : string with match(SELF, '^(nfs4?|panfs|ceph|none)$')
+    'fstype' : choice('nfs', 'nfs4', 'panfs', 'ceph', 'none', 'lustre')
```
ned21, line 13: Can we put the `choice` options in a logical order? I'm OK with the most common coming first, but special types like `none` should be at the end IMHO.

`choice('nfs', 'nfs4', 'ceph', 'lustre', 'panfs', 'none')`

or just go full alphabetic.

ncm-nfs/src/main/perl/nfs.pm
```diff
@@ -261,7 +261,7 @@ The current managed entries are
 
 =over
 
-=item devices with filesystems C<nfs>, C<nfs4>, C<panfs> or C<ceph>.
+=item devices with filesystems C<nfs>, C<nfs4>, C<panfs>, C<ceph> or C<lustre>.
```
ned21, line 264: list the non-nfs types in alphabetical order?

ncm-nfs/src/main/pan/components/nfs/schema.pan
jrha: :+1: for full alphabetic

ned21 (approved):

PR #1512, opened 2021-06-02
ncm-resolver: Use component config header macro

jrha:
This mainly just cleans up and adds support for site-config.

PR #1533, opened 2021-12-20
Fix types for refactored components

jrha:
As a result of modernisation and refactoring work done to `ncm-resolver` and `ncm-pam`, the type definitions in the schema need renaming to match the standard format. e.g. `type component_resolver_type` should be `type resolver_component`.

The changes which caused this to be necessary were made in the following PRs, but the breakage was not noticeable at the time as macro expansion only happens during the release process (in this case for a release candidate) when the component templates are updated in the template library.
- quattor/configuration-modules-core#1511
- quattor/configuration-modules-core#1512

The schema should also not be bound to a path in the schema template, as this is inserted into `config.pan` by the macro.

See quattor/release#339.

### DIFF
--- a/ncm-pam/src/main/pan/components/pam/schema.pan
+++ b/ncm-pam/src/main/pan/components/pam/schema.pan
@@ -29,58 +29,57 @@ declaration template components/pam/schema;
 
 include 'quattor/schema';
 
-type component_pam_options = extensible {
+type pam_component_options = extensible {
 };
 
-type component_listfile_acl = {
+type pam_component_listfile_acl = {
+
     "filename" : string
     "items" : string[]
 };
 
-type component_pam_module_stack = {
+type pam_component_module_stack = {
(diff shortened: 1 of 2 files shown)


ned21:
Do you have a reference (eg. PR) for changes that broke this? I don't remember the refactoring and am in the dark as to what's caused this and thus how to review it for correctness.

ned21 (approved):
Code seems OK but it would be good to document the driver for the change more directly in case it causes problems for users so they can understand why this was necessary.

jrha:
Thanks @ned21, I've updated the PR description with links to the PRs that caused this.

jrha:
The type names themselves are not used anywhere outside of the component schema and config so really there should be no user facing changes from a rename.

jrha:
I've merged this into a [release branch](https://github.com/quattor/configuration-modules-core/tree/quattor-21.12.0) for testing.

PR #1540, opened 2022-02-04
ncm-grub: provide alternative way to set kernel args

stdweird:


ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -76,6 +84,8 @@ type grub_component = {
       If 'fullcontrol' is true then the current arguments passed to the
       kernel are substituted by the ones given in this entry.}
     "args" ? string
+    @{Same as args, but removal is flagged by setting enble=false}
```
jrha, line 87:
```suggestion
    @{Same as args, but removal is flagged by setting enable=false}
```

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -11,6 +11,14 @@ include 'pan/types';
  line, where the first field matches the file_user parameter.
 }
 
+type grub_argument = {
+    @{a value for the argument (creates key=value when present; only key is generated when missing
+      (a list is converted in comma-separated list of strings).}
+    "value" ? element with is_string(SELF) || is_long(SELF) || is_list(SELF)
```
jrha, line 17: Should we restrict leading/trailing white-space for string values?

ncm-grub/src/main/pan/components/grub/schema.pan
```diff
@@ -11,6 +11,14 @@ include 'pan/types';
...
+    @{if false, do not add/update this argument and remove this option from the current list}
```
jrha, line 18: Does this mean, if the argument already exists (i.e. the OS provides it by default) it will removed?

ncm-grub/src/main/pan/components/grub/schema.pan
stdweird: dunno but thanks for reminding i need to quote values with empty space in it

ncm-grub/src/main/pan/components/grub/schema.pan
stdweird: not sure what you exactly ask, but it's the same behaviour as the preceding with a `-` in the args case

ncm-grub/src/main/perl/grub.pm
```diff
@@ -613,6 +701,89 @@ sub get_info
...
+    my $add = {};
+    my $remove = {};
+
+    # Not caring about remove, these are faulty anyway
+    # For all add that have more than one value
+    foreach my $name (sort keys %{$currargs->{add}}) {
+        my $values = $currargs->{add}->{$name};
+        my $len = scalar(@$values);
+        if ($len > 1) {
+            $self->info("Found $len values for $name, replacing with last one: ",
```
kwaegema, line 759: just found out why installing latest kernel puts hosts with multiple mdadm arrays in dracut mode..
```2022/05/10-15:51:25 [INFO] Found 3 values for rd.md.uuid, replacing with last one: 8598f4d5:427993ed:e1092e63:748077b5 , 78d92008:5cd1d326:db95a259:9c632937 , 8795753f:47a571d7:410362b8:b05ac2b9```

thi should be possible though, from dracut manual:
```     rd.md.uuid=<md raid uuid>
           only activate the raid sets with the given UUID. This
           parameter can be specified multiple times.```

jrha (approved):

PR #1546, opened 2022-04-13
ncm-ipmi: Enhancements to IPMI module

Gargravarr2112:
* Add more user account access control functionality including Privilege level
* Explicitly enable IPMI over LAN
* Uses existing variables and structure
* Remove the mandatory, but unused `net_interface` property

### DIFF
--- a/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
+++ b/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
@@ -22,4 +22,3 @@ type component_ipmi_type = {
 };
 
 bind "/software/components/ipmi" = component_ipmi_type;
-
--- a/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
+++ b/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
@@ -2,22 +2,77 @@
 # ${developer-info}
 # ${author-info}
 
-
+@{
+    See Intelligent Platform Management Interface Specification v2.0 rev. 1.1
+    https://www.intel.co.uk/content/www/uk/en/products/docs/servers/ipmi/ipmi-second-gen-interface-spec-v2-rev1-1.html
+}
 declaration template components/ipmi/schema;
 
 include 'quattor/schema';
 
 type structure_users = {
-    "login" : string
-    "password" : string
-    "priv" ? string
-    "userid" ? long
+    @{
+        User Name String in ASCII (16 bytes)
+        See IPMI Spec v2.0 rev 1.1 section 22.28 Set User Name Command
+
+        Restricted to only printable ASCII characters
+    }
+    "login" : string_trimmed(1..16) with match(SELF, '^\p{Print}+$')
+
+    @{
+        Password String in ASCII (16/20 bytes)
+        See IPMI Spec v2.0 rev 1.1 section 22.30 Set User Password Command
+
+        Restricted to only printable ASCII characters
+    }
+    "password" : string_trimmed(0..20) with match(SELF, '^\p{Print}+$')
+
+    @{
+        Channel Privilege Level Limit (4 bits)
+        See IPMI Spec v2.0 rev 1.1 section 22.26 Set User Access Command
+
+        Standard Levels:
+            0 = Reserved (not supported)
+            1 = CALLBACK
+            2 = USER
+            3 = OPERATOR
+            4 = ADMINISTRATOR
+            5 = OEM Proprietary
+            15 = No access
+    }
+    "priv" ? long(1..15) with SELF <= 5 || SELF == 15
+
+    @{
+        Numeric User ID (6 bits)
+
+        Standard values:
+            0 = Reserved (not supported)
+            1 = The null user
+    }
+    "userid" ? long(1..63)
 };
 
 type component_ipmi_type = {
     include structure_component
 
-    "channel" : long = 1
+    @{
+        Channel Number (4 bits)
+        See IPMI Spec v2.0 rev 1.1 section 6.3 Channel Numbers
+
+        Standard Channels:
+            0       Primary IPMB (not supported)
+            1-11    Implementation-specific (normal range)
+            12-13   Reserved (not supported)
+            14      Reserved for identifying current channel (not supported)
+            15      System Interface (not supported)
+    }
+    "channel" : long(1..11) = 1
+
+    @{
+        User Configuration
+
+        List of structure_users type.
+    }
     "users" : structure_users[]
 };
 
--- a/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
+++ b/ncm-ipmi/src/main/pan/components/ipmi/schema.pan
@@ -19,7 +19,6 @@ type component_ipmi_type = {
 
     "channel" : long = 1
     "users" : structure_users[]
-    "net_interface" : string
 };
 
 bind "/software/components/ipmi" = component_ipmi_type;
--- a/ncm-ipmi/src/main/perl/ipmi.pm
+++ b/ncm-ipmi/src/main/perl/ipmi.pm
@@ -9,6 +9,15 @@ use CAF::Process;
 use constant IPMI_EXEC => "/usr/bin/ipmitool";
 use constant BASEPATH => "/software/components/ipmi/";
 
+sub ConfigureNetwork {
+  my ($self, $channel) = @_;
+  #Ensure IPMI over LAN is enabled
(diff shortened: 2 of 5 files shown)


ncm-ipmi/src/main/perl/ipmi.pm
```diff
@@ -29,22 +29,35 @@ sub Configure
       my $passwd = $user->{password};
       my $priv   = $user->{priv};
 
+      #Create and enable user account
       CAF::Process->new([IPMI_EXEC, qw(user set name), $userid, $login],
                         log => $self)->run();
       CAF::Process->new([IPMI_EXEC, qw(user set password), $userid, $passwd],
                         log => $self)->run();
+      CAF::Process->new([IPMI_EXEC, qw(user set priv), $userid, sprintf("0x%X", $priv)],
```
jrha, line 37: This requires `$priv` to be numeric, but in the test case below it is a string.

ncm-ipmi/src/test/resources/ipmi.pan
```diff
@@ -2,7 +2,9 @@ object template ipmi;
 
 prefix "/software/components/ipmi";
 
-"users/0/userid" = "userid";
+"channel" = 1;
+
+"users/0/userid" = 1;
 "users/0/login" = "login";
 "users/0/password" = "password";
 "users/0/priv" = "priv";
```
jrha, line 10: The code now requires this to be a long.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
```diff
@@ -2,25 +2,74 @@
...
 type structure_users = {
-    "login" : string
-    "password" : string
-    "priv" ? string
-    "userid" ? long
+    @{
+        User Name String in ASCII (16 bytes)
+        See IPMI Spec v2.0 rev 1.1 section 22.28 Set User Name Command
+    }
+    "login" : string_trimmed(1..16) with match(SELF, '^[ -~]+$')
```
stdweird, line 18: what is this match supposed to do?

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
```diff
@@ -2,25 +2,74 @@
...
+
+    @{
+        Password String in ASCII (16/20 bytes)
+        See IPMI Spec v2.0 rev 1.1 section 22.30 Set User Password Command
+    }
+    "password" : string_trimmed(0..20) with match(SELF, '^[ -~]+$')
```
stdweird, line 24: same question

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
```diff
@@ -2,25 +2,74 @@
...
+        Standard Levels:
+            0 = Reserved (not supported)
+            1 = CALLBACK
+            2 = USER
+            3 = OPERATOR
+            4 = ADMINISTRATOR
+            5 = OEM Proprietary
+            15 = No access
+    }
+    "priv" ? long(1..15) with SELF <= 5 || SELF == 15
```
stdweird, line 39: `choice` is string only?

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
```diff
@@ -2,25 +2,74 @@
...
+    }
+    "channel" : long(1..11) = 1
+
+    @{
+        User Configuration
+
+        List of structure_users type.
+    }
     "users" : structure_users[]
-    "net_interface" : string
```
stdweird, line 22: you remove a previously mandatory option. this is backwards incompatible, at least mention it in the PR

ncm-ipmi/src/main/perl/ipmi.pm
```diff
@@ -29,22 +37,27 @@ sub Configure
...
+      #Set remote access via LAN
+      CAF::Process->new([IPMI_EXEC, qw(channel setaccess), $channel, $userid, qw(link=on ipmi=on callin=on)],
+                        log => $self)->run();
   }
 
-  CAF::Process->new([IPMI_EXEC, qw(mc reset cold)],
-                    log => $self)->run();
+  $self->ConfigureNetwork($channel);
+
+  # Reset BMC to ensure changes have been made active
```
stdweird, line 56: how does this ensure it?

ncm-ipmi/src/test/resources/ipmi.pan
```diff
@@ -2,7 +2,9 @@ object template ipmi;
 
 prefix "/software/components/ipmi";
```
stdweird, line 3: bind the schema here as well, to test the schema

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: Restrict the field to only printable ASCII characters, which happens to be from space to to tilde.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: Restrict the field to only printable ASCII characters, which happens to be from space to to tilde.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: `choice` is indeed string-only.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: Good point

ncm-ipmi/src/test/resources/ipmi.pan
jrha: good call

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: It could instead be `^[\x20-\x7e]+$` if preferred.

Gargravarr2112: The hex-code does make it more obvious it's a range, if less nice to read.

jrha: AFAICT POSIX character classes aren't supported, but if they are then `^[[:print:]]+$`.

jrha: Ah, of course there's a special Java-way of doing POSIX character classes: `^\p{Print}+$`.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
jrha: Done.

ncm-ipmi/src/main/perl/ipmi.pm
jrha: @Gargravarr2112?

ncm-ipmi/src/test/resources/ipmi.pan
jrha: Done.

ncm-ipmi/src/main/pan/components/ipmi/schema.pan
stdweird: i thought as much, it should have been `match(SELF, '^[^ -~]+$')` :wink:

stdweird: ah, so it really was intended as a range. :shivers:

ncm-ipmi/src/main/perl/ipmi.pm
stdweird: i also do a cold reset when making changes, but then i also check if the changes are still present. however, this would make this component very slow.
also if you monitor your bmcs, these will go in error now and then (i don't know how this component works, but it's not stateful i think).

what you could try to do is dump eg `ipmitool lan print` to a semi permanent quattor-related file, and diff it upon a run. if there are actual changes before and after, only then reset and redump/recheck values?

stdweird: but tbh, i don't care that much. i am ok to merge this PR in its current state

jrha: I like the idea of tracking the state, but I'll open an issue for it.

jrha:
@stdweird I've created an issue to implement state tracking later, I think this is good to merge if you're happy with the rest of it.

stdweird (approved):

PR #1549, opened 2022-04-27
CI: fix hash key for m2 cache

wpoely86:
For some reason, hashlib function does no longer work in the post step.
I've changed it to generate the hash in the beginning and store it in a
variable.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -9,6 +9,9 @@ jobs:
     steps:
     - name: Checkout code
       uses: actions/checkout@v2
+    - name: Determine hash for caching key
+      id: cachekeystep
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -36,8 +39,8 @@ jobs:
       uses: actions/cache@v2
       with:
         path: /tmp/m2
-        key: ${{ runner.os }}-m2-${{ hashFiles('**/pom.xml') }}
-        restore-keys: ${{ runner.os }}-m2
+        key: ${{ runner.os }}-m2-${{ steps.cachekeystep.outputs.pomcachekey }}
+        restore-keys: ${{ runner.os }}-m2-
```
jrha, line 43: I don't understand the significance of the extra `-` at the end of this line. Could you elaborate?

jrha (approved):

PR #1550, opened 2022-06-01
ncm-chkconfig: ensure that the services property doesn't remain undefined

jouvin:
- With the migration to ncm-systemd, a configuration can define the legacy module and endup with an invalid
configuration if no service has been added to ncm-chkconfig


ned21 (approved):
panlint failures are unrelated.

jouvin:
They are in the component schema, I'll fix them.

jrha (approved):

PR #1551, opened 2022-07-05
ncm-metaconfig: Add service action condrestart

jose-caballero:
Current list of allowed service actions includes restart, reload, and stop_sleep_start.
This creates problems when the metaconfig component issues a reload action for a given service on a brand new host.
Since the host is new, that service has never been started, and therefore the reload command fails.
This also prevents the rest of Quattor components depending on metaconfig from running.
To solve this, allow service command condrestart, which works whether the service is already running or not.


ned21:
Thanks for the PR! The panlint errors are unrelated to the change but this LGTM if the main tests pass.

jose-caballero:
Thanks. Is there anything else I need to do?

PR #1555, opened 2022-09-19
metaconfig: haproxy: tunnel timeout settings

stdweird:


ncm-metaconfig/src/main/metaconfig/haproxy/pan/schema.pan
```diff
@@ -66,6 +66,9 @@ type haproxy_service_timeouts = {
     'connect' : long = 3500
     'client' : long = 10000
     'server' : long = 10000
+    'client-fin' ? long
+    'server-fin' ? long
+    'tunnel' ? long
```
jrha, line 71: How about some minimum bounds?
The manual seems to imply 4 seconds should be the minimum.
```suggestion
    'client-fin' ? long(4000..)
    'server-fin' ? long(4000..)
    'tunnel' ? long(4000..)
```

PR #1556, opened 2022-09-28
ncm-metaconfig: add missing kafka options

wdpypere:


wdpypere:
the should be `int.int` as a string. where it should match an actual kafka version. Do you have suggestions on how to improve the schema?

jrha:
How about…
```pan
string with match(SELF, '^\d+\.\d+$');
```

wdpypere:
thanks @jrha , fixed.

jrha (approved):
LGTM

PR #1557, opened 2022-09-28
ncm-metaconfig: add missing zookeeper 4lw options

wdpypere:


ncm-metaconfig/src/main/metaconfig/zookeeper/pan/schema.pan
```diff
@@ -22,6 +22,7 @@ type zookeeper_main = {
     "autopurge.snapRetainCount" ? long
     "autopurge.purgeInterval" ? long
     "syncEnabled" ? boolean
+    "4lw.commands.whitelist" ? string
```
jrha, line 25: Can this be anything more specific than just `string`?

wdpypere: it should be a comma separated list of ` conf, cons, crst, dump, envi, ruok, srst, srvr, stat, wchs, wchc, wchp, mntr`.

I needed it quickly on our site during an upgrade, but I should convert it to a choice list. It will need some adjustments to the TT file though. I'll see what I can do.

PR #1558, opened 2022-09-30
ncm-spma: dnf support for enabling module streams

aka7:
add support for enabling module stream using modulemd-defaults for dnf backend. Similar to yumdnf, but this configures via module defaults configs using drop-in files in /etc/dnf/modules.defaults.d, includes setting of default profiles.

plus
- providng a new function module_add
- make panlint happy

This shouldn't create any backward incompatibly. If no "/software/modules" is set, it should be a   noop.
fix few pantlint errors in functions.pan, could have done a separate commit, but don't think it's that big of change for it.


ncm-spma/src/main/pan/components/spma/dnf/config.pan
```diff
@@ -8,7 +8,11 @@ prefix '/software';
 # Package to install
 'packages' = pkg_repl("ncm-${project.artifactId}", "${no-snapshot-version}-${rpm.release}", "noarch");
 
+# modules can be empty, when nothing is set
+"/software/modules" = dict();
```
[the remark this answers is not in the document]
aka7: done, also just noticed software prefix above, so changed to "modules" ?= dict()  to keep it consistent.

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -22,4 +22,14 @@ type component_spma_dnf = {
     "run"                ? boolean  # Run the transaction after configuring DNF
 };
 
+@{DNF module structure. To be used as dict, with name as key.
+ to set default profiles for modules not enabled by default and change existing defaults.
+}
+type component_spma_dnf_module_defaults = {
+    'stream'   : string
+    'profiles' ?  string[]
```
jrha, line 30: I'm assuming neither of these should have any leading or trailing white-space?
```suggestion
    'stream'   : string_trimmed
    'profiles' ?  string_trimmed[]
```

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
```diff
@@ -22,4 +22,14 @@ type component_spma_dnf = {
...
+    @documentation{ defaults to 1 in component, provided in-case it changes in a future release }
+    'modulemd_version' : long = 1
```
jrha, line 32: Looks like [there is already a version 2](https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml), we should probably at least put a lower bound on this, but maybe just restrict it to `1` until support is implemented for another version?
```suggestion
    'modulemd_version' : long(1..) = 1
```

ncm-spma/src/main/pan/components/spma/dnf/schema.pan
aka7: interesting, our release doesn't seem to have this, updated anyway, we can look at this internally.

aka7:
Anyone else have any other thoughts on this? I would like this in the current release if possible so we can get this into our rhel8 release asap.

ncm-spma/src/main/pan/components/spma/functions.pan
```diff
@@ -447,3 +447,35 @@ function pkg_ronly = {
...
 };
+
+@documentation{
+    desc =  Add config for default module stream for overriding module defualt streams
+    arg = name
+    arg = stream
+    arg = (optional) profiles (type list(), set these as default profile)
+    arg = (optional) modulemd_verison (moduelmd-defualts profile version,it's currently 1 on rhel8 release)
+}
+function module_add = {
```
stdweird, line 458: out of curiosity, why do you have a function for this? is it to protect overriding existing streams (cfr th error)?

ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -102,6 +104,67 @@ sub get_installed_rpms
...
+
+    my $modules = $config->getTree(MODULES_TREE);
+
+    # For easier lookup
+    my %modulenames;
+    foreach my $module (sort keys %$modules) {
+        $modulenames{$module} = $module;
+    }
+    # Remove default module definitions not present in the profile
+    my @module_defaults = glob "/etc/dnf/modules.defaults.d/*";
```
stdweird, line 150: this should be `glob $directory`

ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -102,6 +104,67 @@ sub get_installed_rpms
...
+                                  log => $self);
+        $changes += $fh->close() || 0; # handle undef
+    }
+
+    return $changes;
+};
+
+
+# support to override/set default stream in /etc/dnf/modules.defaults.d/
+# directory is here for unittesting
```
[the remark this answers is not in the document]
aka7: I see dnf module doesn't have any tests at present but this is there for adding in future hopefully.  I haven't look deep on how the tests can be added and tested.

ncm-spma/src/main/pan/components/spma/functions.pan
aka7: partly that, similar to pkg_add_repl wanted to do any checks in one place.   then in pan, we can just use module_add(..)
but if you have better suggestion, then I'm all for it.

stdweird: depends a bit hat you want to configure imho. you can also use structure template(s) or plain diretc pan like
```
prefix /software/modules;
'python36/stream' = '3.6';
...
```

all i am saying is that function doesn't do that much imho.

aka7: So all depends on what we want here, should the last entry win? function checks for duplicate definition of a module and fails compile. AFIAK you can't activate/set default  of two streams of same module,  so we can either silently take the last entry or fail compile. Function is my way of providing this check and have it fail compile to alert user that two stream of same module cannot be  enabled.

taking your approach, but with a module that has different streams,  
```
prefix "/software/modules";
"postgresql/stream" = "11";
```
then somewhere else in template, (through another feature maybe), someone adds following, 
```
prefix "/software/modules";
"postgresql/stream" = "10";
```
should this fail compile?

setting through the function, it will fail compile. i.e 
```

"/software/modules" = module_add("postgresql", "11", list("server"));
....
"/software/modules" = module_add("postgresql", "10", list("server"));

```
but anyway maybe this is our use case and there are better ways to do this? So I'll be happy to remove the function if this isn't desirable upstream?

stdweird: @aka7 to be clear, your function is desirable ;) but the pan code over time has accumulated lots of code and variables that imho would be better hamdled with using path.

so just so you know, the function does not prevent anyone to override it later using the direct path.
there are several ways to deal with this problem, but all have their drawbacks.

if you want to hard protect a stream path from alteration, you can use `final` (but that also has drawbacks).
defaults are also nice to use.

we are used to having the `last one wins`, but you might have different experience

aka7: @stdweird we are also used to having last one wins, and you make valid point about function doesn't prevent someone from setting using direct path.   I think this probably requires further discussion if we are to provide a solution for duplicate entries where last one should not win.  This is why I'm suggesting maybe I remove function from pr for now for further discussion, mainly  so I can get rest of code in. We are keen to get a new point release with this change, if possible. If the rest of the code makes sense.

Reason for the rush is that, redhat annoying made lot of modules default which come EOL next year, where as they have different stream with longer support, but are not set as the default. We like to make all long term support modules the default before we release RHEL8 into the wild.  if that makes sense.

aka7: I had a offline conversation with @jrha  and we agreed to remove the function. This is now updated.

jrha (approved):
In general this could definitely do with unit-tests, but given the original yum-variant of this had none, this is no worse so is good enough for now.

PR #1563, opened 2022-12-09
Don't specify Python 3 minor version

jrha:
GitHub is updating runner images at the moment which makes the availability of specific versions unpredictable.

### DIFF
--- a/.github/workflows/panlint.yaml
+++ b/.github/workflows/panlint.yaml
@@ -13,7 +13,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
-          python-version: 3.6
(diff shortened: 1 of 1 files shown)


wpoely86 (approved):

wpoely86:
We would have to specify that we want the Ubuntu 20.04 runners to get python 3.6.

jrha:
Indeed, but as we don't require a specific version to test our code, only Python 3 to run Panlint we don't really care right?

wpoely86:
Yes, it should be fine. I originally picked 3.6 as that's the python of RHEL 8. The checks will tell us if panlint would not be compatibility with the latest python in any case.

PR #1567, opened 2022-12-14
ncm-systemd: All Exec* options can be lists of strings

### LINKED ISSUE #1262: ncm-systemd: ExecStart should be a list of string, not a string
`ExecStart` can appear multiple times in a unit file to execute several commands. Because of this, if you want to redefine the standard `ExecStart` attribute in a custom file (appearing in a `service.d` directory), you need to first reset the standard `ExecStart` and then define the new one, meaning you need to specifiy at least 2 values. This is not possible with the current schema, as of 17.12. The possibility to have several commands separated by a `;` doesn't apply to the reset case (an empty string before the `; ` is not allowed and fails the test that the command path must be absolute).

For the backward compatibility we could allow both a string and a list of string (I think pan allows this) and have the component do the right thing based on the actual value type....

jrha:
Resolves #1262.

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -252,11 +252,11 @@ type ${project.artifactId}_unitfile_config_service = {
     'BusName' ? string
     'BusPolicy' ? string[] with length(SELF) == 2 && match(SELF[1], '^(see|talk|own)$')
     'CapabilityBoundingSet' ? linux_capability[]
-    'ExecReload' ? string
-    'ExecStart' ? string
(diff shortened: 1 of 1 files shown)


ned21:
@aka7  - is this similar to the issue we just hit with importing the latest schema?

aka7:
> @aka7 - is this similar to the issue we just hit with importing the latest schema?
@ned21 yes this would be an issue for us to import I think.  Since we will have hosts running on older releases so is likely to have previous versions of ncm.  Therefore in order to import it, we have to make sure we won't be going back on older releases. We may need to think about how we provide these repository.  ( If I understood this right )

wpoely86 (approved):

Issue #1055, opened 2017-02-27
ncm-syslogng: Archive and remove from repository


jrha:
See #1027.

jrha:
@ned21 can this component be archived now?

jrha:
RAL is no longer using syslogng.

ned21:
Looks like it is still in use on ~100 hosts. Not sure why, looks like some niche configs were never migrated. @aka7 is probably the best person to dig into this these days.

jouvin:
GRIF is not using it.

aka7:
> Looks like it is still in use on ~100 hosts. Not sure why, looks like some niche configs were never migrated. @aka7 is probably the best person to dig into this these days.

created an internal jira to dig into  and will report back.

jrha:
@aka7 any update?

aka7:
@jrha Our usage has gone down since last update, however I'm still seeing some ~60 hosts using it.  I will chase them. Do you have date of when 23.6 will be released?

jrha:
Ok thanks, the aim for 23.6 is ~June 2023.

PR #1568, opened 2022-12-14
ncm-syslogng: Add deprecation warning to schema

jrha:
First step towards removal (see #1055).


ned21:
I think this is OK because we can just skip importing the deprecation warning, we know we need to kill the component.

stdweird (approved):

Issue #1496, opened 2021-03-22
ncm-hostsfile: Duplicate host entries get overwritten e.g. localhost


Gargravarr2112:
Similar to the request in #1184 , there is a bug in not being able to support duplicate host entries. Primarily, on EL7, the default hosts file contains entries for localhost for both IPv4 and IPv6. Since these are the same hostname and are stored under the same key, the IPv6 version overwrites the IPv4 version and the latter is lost.

I believe the problem is specifically here:
https://github.com/quattor/configuration-modules-core/blob/400aba5c9c5c344fc8b88f0a54209d038fbaedaa/ncm-hostsfile/src/main/perl/hostsfile.pm#L109

* The version of the component you are using (rpm -q $component)
_ncm-hostsfile-19.12.0-1.noarch_
* What OS version you are using (/etc/redhat-release or similar)
_Oracle Linux Server 7.9_
* Steps required to reproduce the problem
_Include components/hostfile and define a hosts entry_
* Actual result versus expected result
_Expected_: The existing hosts are preserved while defined host is added.
_Actual_: First (IPv4) localhost entry is lost, leaving only the IPv6 entry.

jrha:
I presume `/etc/hosts` looks something like this before the component runs?
```
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
````

**Comment by Gargravarr2112** [comment]
Hi James,
It does indeed look like that. However it removes the first (IPv4) line. I think this is because the key ('localhost') is being reused for the next line.
Cheers,
Rob

**Comment by jrha** [comment]
Looking at this more closely, if you wanted to set `'takeover' = true` to enfore both the IPv6 and IPv4 lines, the schema would also prevent `localhost` from being defined twice.

You can work around both issues using the `aliases` functionality though, you just need to ensure that the canonical hostname is unique.

For example:
```pan
prefix '/software/components/hostsfile';
'active' = true;

'localhost6/ipaddr' = '::1';
'localhost6/aliases' = 'localhost localhost.localdomain localhost6.localdomain6';
```

Or:
```pan
prefix '/software/components/hostsfile';
'active' = true;
'takeover' = true;

'localhost/ipaddr' = '127.0.0.1';
'localhost/aliases' = 'localhost.localdomain localhost4 localhost4.localdomain4';

'localhost6/ipaddr' = '::1';
[...]

PR #1571, opened 2023-01-03
ncm-hostsfile: Test IPv6 and dual stack localhosts

jrha:
In these we show how dual stack entries can set takeover and overwrite the existing localhost entry or not, specifying aliases for both stacks that allow `localhost` to resolve for both protocols while still avoiding duplicate entries.

This demonstrates and tests workarounds for the issue reported in #1496.


ncm-hostsfile/src/test/resources/dual_stack_localhost.pan
```diff
@@ -0,0 +1,15 @@
...
+
+prefix '/software/components/hostsfile';
+
+'active' = true;
+'file' = '/tmp/hosts.local';
+'takeover' = true;
+
+prefix 'entries';
+# IPv4
+'localhost/ipaddr' = '127.0.0.1';
```
stdweird, line 11: @jrha it might make sense to define 2 final variables with these ipv4 and ipv6 dicts, so other can use this workaround with ease.

i'm willing to merge in any case ;)

jrha:
@stdweird I've re-factored and provided two `dict` as suggested.

stdweird (approved):

PR #1572, opened 2023-01-03
ncm-spma: fix dnf provider not to remove running kernel

### LINKED ISSUE #1565: ncm-spma: dnf provider tries to remove running kernel-core and fail
The kernel-core package on RHEL8 is a protected package as it delivers the actual kernel files and core drivers. When moving from one kernel release to another, if the running kernel is not one of listed kernels in host profile, spma will try to remove it but dnf will prevent this from happening and spma run will fail.

Looking at the code, following does exists but doesn't appear to be working as intended. 

```
   # Do not remove currently running kernel.
    my $kvers = (POSIX::uname())[2];
    $self->info("Will skip removal of running kernel: $kvers");
 
    # Compose list of packages to be installed and removed.
    my $will_remove = $preinstalled - $to_install - "kernel-$kvers" - "kernel-core-$kvers";
    my $will_install = $to_install - $preinstalled;
    $self->info("removal list: $will_remove");
```

running kernel
```
$ uname -r
4.18.0-372.13.1.el8_6.x86_64
```
this version is not listed in host profile as one of kernels to be installed.
spma run will show this, 

`[INFO] Will skip removal of running kernel: 4.18.0-372.13.1.el8_6.x86_64  `

But in the  removal list is below, the  running kernel-$kvers kernel-core-$kvers is in the list to be removed.

`[INFO] removal list: (gpg-pubkey-0:d4082792-5b32db75.(none) gpg-pubkey-0:de57bfbe-53a9be98.(none) gpg-pubkey-0:e0178866-4aeeee16.(none) gpg-pubkey-0:fd431d51-4ae0493b.(none) kernel-0:4.18.0-372.13.1.el8_6.x86_64 kernel-core-0:4.18.0-372.13.1.el8_6.x86_64 kernel-devel-0:4.18.0-372.13.1.el8_6.x86_64 kernel-headers-0:4.18.0-372.13.1.el8_6.x86_64 kernel-modules-0:4.18.0-372.13.1.el8_6.x86_64 kmod-openafs-4.18.0-372.13.1-0:1.6.24-c14.msdw.1.el8.x86_64 krb5-libs-0:1.18.2-21.el8.i686 krb5-libs-0:1.18.2-21.el8.x86_64 krb5-workstation-0:1.18.2-21.el8.x86_64 libkadm5-0:1.18.2-21.el8.x86_64)`
 
This is a bug  as $kvers has not taken into account the $epoch value thats being set in $to_install.  therefore, in this case $kver is 4.18.0-372.13.1.el8_6.x86_64 and 

kernel-core-4.18.0-372.13.1.el8_6.x86_64 != kernel-core-0:4.18.0-372.13.1.el8_6.x86_64  (notice the -0:)

```
$ rpm -q ncm-spma
ncm-spma-21.12.1-1.noarch
```

```
$ cat /etc/redhat-release
Red Hat Enterprise Linux release 8.7 (Ootpa)
```

change host profile to include kernels to install which is not the running kernel.

Expected result:  running kernel should not be removed by spma
actual result: spma tries to remove kernel-core.

aka7:
Fix bug in dnf provider to correctly exclude removal of packages core to running kernel so it doesn't try to remove running kernel.
Fixes #1565

The code to exclude the running kernel doesn't take into account the epoch value that is added in the $will_remove var. (i.e kernel-core-0: vs kernel-core. 0 is the epoch value, 0 is when no epoch value is set).  To fix this, I've now moved to do this within the existing loop which does removal for whitelist exclusion. 

see issue #1565

Tested on a rhel8 host which is using the dnf provider.

### DIFF
--- a/ncm-spma/src/main/perl/spma/dnf.pm
+++ b/ncm-spma/src/main/perl/spma/dnf.pm
@@ -529,12 +529,11 @@ sub Configure
         return 0;
     }
 
-    # Do not remove currently running kernel.
+    # get currently running kernel version.
     my $kvers = (POSIX::uname())[2];
-    $self->info("Will skip removal of running kernel: $kvers");
 
     # Compose list of packages to be installed and removed.
-    my $will_remove = $preinstalled - $to_install - "kernel-$kvers" - "kernel-core-$kvers";
+    my $will_remove = $preinstalled - $to_install;
     my $will_install = $to_install - $preinstalled;
     my $whitelist = $t->{whitelist};
     my $whitelisted = Set::Scalar->new();
@@ -542,12 +541,19 @@ sub Configure
         if (substr($rpm, 0, 11) eq 'gpg-pubkey-') {
             $will_remove->delete($rpm);
         }
+        my $rpm_version = (split(':', $rpm))[-1];
+        
+        # Do not remove core packages for currently running kernel.
+        if ( ($rpm =~ /^(kernel|kernel-core|kernel-modules)-(\d+):/) &&
+            (index($rpm_version, $kvers) == 0 || match_glob($kvers, $rpm_version))) {
+            $self->info("Skip removal of core package(s) for running kernel: $rpm");
+            $will_remove->delete($rpm);
+        }
+
         # Do not remove whitelisted packages.
         if (defined($whitelist)) {
             for my $white_pkg (@$whitelist) {
-                my $rpm_noepoch = $rpm;
-                $rpm_noepoch =~ s/^.*://;
-                if (index($rpm_noepoch, $white_pkg) == 0 || match_glob($white_pkg, $rpm_noepoch)) {
+                if (index($rpm_version, $white_pkg) == 0 || match_glob($white_pkg, $rpm_version)) {
                     $will_remove->delete($rpm);
                     $whitelisted->insert($rpm);
                 }


ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -534,19 +534,26 @@ sub Configure
     $self->info("Will skip removal of running kernel: $kvers");
```
ned21, line 534: Not sure you need this info() log line any more since the real work is done later (and also logged via info()?) It records the value of $kvers but not sure that's useful enough for info.

ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -534,19 +534,26 @@ sub Configure
...
+    my $will_remove = $preinstalled - $to_install;
     my $will_install = $to_install - $preinstalled;
     my $whitelist = $t->{whitelist};
     my $whitelisted = Set::Scalar->new();
     for my $rpm ($will_remove->elements) {    # do not remove imported GPG keys
         if (substr($rpm, 0, 11) eq 'gpg-pubkey-') {
             $will_remove->delete($rpm);
         }
+        my $rpm_noepoch = $rpm;
+        $rpm_noepoch =~ s/^.*://;
```
ned21, line 546: My expectation of a variable called `$rpm_noepoch` is that it's the whole thing without an epoch, but looking carefully at the substitution it removes everything before the (last) colon, making the variable more accurately named, `$rpm_version_noepoch` ?

I might be tempted to go for something like this as being a bit more readable than a regex too (but maybe I'm just out of tune with Perl ;-)
```
$rpm_version = (split(':', $rpm))[-1];
```

ncm-spma/src/main/perl/spma/dnf.pm
aka7: was debating about this myself. I agree, don't think it's much of a useful info.

ncm-spma/src/main/perl/spma/dnf.pm
aka7: Make sense thanks.

ned21:
LGTM but I'll leave someone from outside MS to approve.

aka7:
@jrha would you be able to take a look at this? I believe only MS is using the dnf provider. 
If all is ok, we would like to get this into current release of 21.12.1 if possible ?

jrha (approved):

PR #1576, opened 2023-02-08
ncm-spma: enhance dnf provider to allow update to newer streams.

### LINKED ISSUE #1565: ncm-spma: dnf provider tries to remove running kernel-core and fail

aka7:
change dnf provider to reset module stream on config change to allow updates to happen on next spma run.

Fixes #1565

To allow changes to streams without having to rebuild, or run commands manually.
none.


ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -125,7 +125,15 @@ sub generate_dnf_modules {
         my $fh = $trd->filewriter("$directory/$name.yaml",
                                   header => "# File generated by " . __PACKAGE__ . ". Do not edit",
                                   log => $self);
-        $changes += $fh->close() || 0; # handle undef
+        $module_changed += $fh->close() || 0; #handle undef
+        $changes += $module_changed; 
+        if ($module_changed > 0 and !$CAF::Object::NoAction) {
+            my ($cmd_exit, $cmd_out, $cmd_err) = $self->execute_command([DNF_MODULE_RESET, $name], "module stream change detected. Reset managed module stream ".$name." for update", 1, "/dev/null", 1);
```
ned21, line 131: This line is unreadably long, can you break it up so it wraps a bit more easily please? Also I don't see any reason not to use string interpolation here? 

"module stream change detected. Reset managed module stream $name for update"

ncm-spma/src/main/perl/spma/dnf.pm
```diff
@@ -125,7 +125,15 @@ sub generate_dnf_modules {
...
+            if ($cmd_exit) {
+                $self->error("Error during reset module ". $cmd_err);
```
ned21, line 133: Use "Error during reset module $cmd_err"

The only reason to use concat (.) is if you need to use ' instead of "

ned21 (commented):
Minor formatting. Am I right in thinking we're the only site to use this so we're OK to merge this without explicit approval from another site?

ncm-spma/src/main/perl/spma/dnf.pm
aka7: updated

ncm-spma/src/main/perl/spma/dnf.pm
aka7: Thanks. updated

ncm-spma/src/main/perl/spma/dnf.pm
aka7: and yes I believe DNF provider is not used by anyone else. We started using on rhel8 builds.

ned21 (approved):

PR #1580, opened 2023-03-23
metaconfig: cumulus: basic initialisation script and bgp frr support

stdweird:


ncm-metaconfig/src/main/metaconfig/cumulus/pan/schema.pan
```diff
@@ -227,10 +227,23 @@ type cumulus_frr_route = {
...
+    @{ipv4 networks}
+    'ipv4' : type_ipv4_netmask_pair[]
+};
+
 @{Simple/minimal FRRouting config file, useful for static routing in VRF setup}
 type cumulus_frr = {
     @{Routes per VRF (key is VRF name)}
-    'vrf' : cumulus_frr_route[]{}
+    'vrf' ? cumulus_frr_route[]{}
+    @{List of bgp routers per CRF (VRF is the key)}
```
jrha, line 245:
```suggestion
    @{List of bgp routers per VRF (VRF is the key)}
```

jrha (approved):

PR #1582, opened 2023-04-12
ncm-systemd: add support for slice and more resource control

wpoely86:
The resource control options are from the man page of rockylinux 8.6. It's mostly about cgroupvs2.

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -222,18 +222,29 @@ valid for [Slice], [Scope], [Service], [Socket], [Mount], or [Swap] sections
 type ${project.artifactId}_unitfile_config_systemd_resource_control = {
     'CPUAccounting' ? boolean
     'CPUShares' ? long(2..262144)
+    'CPUWeight' ? long(1..10000)
+    'StartupCPUWeight' ? long(1..10000)
     'StartupCPUShares' ? long(2..262144)
     'CPUQuota' ? long(0..100)  # percentages
     'MemoryAccounting' ? boolean
     'MemoryLimit' ? long  # in bytes
+    'MemoryMin' ? long  # in bytes
+    'MemoryMax' ? long  # in bytes
+    'MemoryLow' ? long  # in bytes
+    'MemoryHigh' ? long  # in bytes
+    'MemorySwapMax' ? long  # in bytes
     'TasksAccounting' ? boolean
     'TasksMax' ? string with match(SELF, '^([0-9]+%?|infinity)$')
     'BlockIOAccounting' ? boolean
     'BlockIOWeight' ? long(10..1000)
+    'IOWeight' ? long(1..10000)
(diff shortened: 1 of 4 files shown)


ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -48,6 +48,23 @@ type ${project.artifactId}_valid_execpath = string with match(SELF, '^([@+!:-]|!
...
+@documentation{
+    Validate that a property is either a size (in bytes), a relative size (in %)
+    or 'infinity'. Used for memory limits in cgroups.
+}
+function is_absolute_or_relative_size = {
+    l = ARGV[0];
+    if (is_long(l) && l > 0) {
+        return(true);
+    };
+    if (is_string(l) && match(l, '^(infinity|1?[0-9]?[0-9]%)$')) {
```
jrha, line 60: This allows values up to and including `199%`, was this intended?

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -222,18 +239,29 @@ valid for [Slice], [Scope], [Service], [Socket], [Mount], or [Swap] sections
 type ${project.artifactId}_unitfile_config_systemd_resource_control = {
     'CPUAccounting' ? boolean
     'CPUShares' ? long(2..262144)
+    'CPUWeight' ? long(1..10000)
```
jrha, line 242: It's probably worth adding a `systemd_weight` type for all the weights that are `long(1..10000)`.

jrha (commented):
Looks good against upstream (https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html)

If I'm interpreting it correctly, the old cgroup options are deprecated:
- CPUShares
- StartupCPUShares
- MemoryLimit
- BlockIOAccounting
- BlockIOWeight
- StartupBlockIOWeight
- BlockIODeviceWeight
- BlockIOReadBandwidth
- BlockIOWriteBandwidth
and the new cgroup options are mutually exclusive with them.

Should we validate that both are not being used together?

wpoely86:
> Looks good against upstream (https://www.freedesktop.org/software/systemd/man/systemd.resource-control.html)

Keep in mind that systemd is a fast moving target. I'm only looking at the man page of systemd under RHEL 8.

 
> If I'm interpreting it correctly, the old cgroup options are deprecated:
> 
...
> 
> and the new cgroup options are mutually exclusive with them.
> 
> Should we validate that both are not being used together?

They are still valid in RHEL 7.x. I'm not sure how you deal with that in quattor?

As far as I understand, if both options would be present, systemd will simple user the newer ones.

```
       To ease the transition, there is best-effort translation between the two versions of settings. For each controller, if any of the settings for the unified hierarchy are present, all settings for the legacy hierarchy are
       ignored. If the resulting settings are for the other type of hierarchy, the configurations are translated before application.
```

So I would leave both here for the moment.

jrha (dismissed):
LGTM

jrha (approved):

PR #1583, opened 2023-04-13
ncm-systemd: Add path support

DavidFair:
Adds support for [path] directives. This is useful for credential handling, where we can detect the template file changing and trigger a one-shot service.

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -386,6 +386,23 @@ type ${project.artifactId}_unitfile_config_socket = {
 };
 
 @documentation{
+the [Path] section
+https://www.freedesktop.org/software/systemd/man/systemd.path.html
+}
+type ${project.artifactId}_unitfile_config_path = {
+    'PathExists' ? absolute_file_path
+    'PathExistsGlob' ? absolute_file_path
+    'PathChanged' ? absolute_file_path
+    'PathModified' ? absolute_file_path
+    'DirectoryNotEmpty' ? absolute_file_path
+    'Unit' ? string
+    'MakeDirectory' ? boolean
+    'DirectoryMode' ? type_octal_mode
+    'TriggerLimitIntervalSec' ? long(0..)
+    'TriggerLimitBurst' ? long(0..)
+};
+
+@documentation{
 the [mount] section
 http://www.freedesktop.org/software/systemd/man/systemd.mount.html
 }
@@ -460,6 +477,7 @@ type ${project.artifactId}_unitfile_config = {
     'socket' ? ${project.artifactId}_unitfile_config_socket
     'mount' ? ${project.artifactId}_unitfile_config_mount
     'automount' ? ${project.artifactId}_unitfile_config_automount
+    'path' ? ${project.artifactId}_unitfile_config_path
     'timer' ? ${project.artifactId}_unitfile_config_timer
     'unit' ? ${project.artifactId}_unitfile_config_unit
     'slice' ? ${project.artifactId}_unitfile_config_slice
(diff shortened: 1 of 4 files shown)


wpoely86:
I guess you also want `.path` files? In that case you need to add a couple of extra bits, see #1582 which does this for a `.slice` unit.

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -355,6 +355,24 @@ type ${project.artifactId}_unitfile_config_socket = {
...
+    'PathExists' ? absolute_file_path
+    'PathExistsGlob' ? string
+    'PathChanged' ? absolute_file_path
+    'PathModified' ? absolute_file_path
+    'DirectoryNotEmpty' ? absolute_file_path
+
+    'Unit' ? string
+    'MakeDirectory' ? boolean
+    'DirectoryMode' ? type_octal_mode
+    'TriggerLimitIntervalSec' ? string
```
jrha, line 372:
```suggestion
    'TriggerLimitIntervalSec' ? long(0..)
```

DavidFair: The documentation suggests users put a unit prefix (e.g. 1s or 1us): https://www.freedesktop.org/software/systemd/man/systemd.path.html
Do we have a type which supports this?

DavidFair:
> I guess you also want `.path` files? In that case you need to add a couple of extra bits, see #1582 which does this for a `.slice` unit.

I'm not sure what you mean sorry? I will add a section to the unit tests though based on that PR

wpoely86:
> > I guess you also want `.path` files? In that case you need to add a couple of extra bits, see #1582 which does this for a `.slice` unit.
> 
> I'm not sure what you mean sorry? I will add a section to the unit tests though based on that PR

Can't make a `.path` unit right now with this. You need to a path to the list on line 504 of `schema.pan`

ncm-systemd/src/main/pan/components/systemd/schema.pan
jrha: For simplicity we normally only support the base unit (in this case seconds).

ncm-systemd/src/main/pan/components/systemd/schema.pan
```diff
@@ -355,6 +355,24 @@ type ${project.artifactId}_unitfile_config_socket = {
     true;
 };
 
+@documentation{
+the [Path] section
+https://www.freedesktop.org/software/systemd/man/systemd.path.html
+}
+type ${project.artifactId}_unitfile_config_path = {
+    'PathExists' ? absolute_file_path
+    'PathExistsGlob' ? string
```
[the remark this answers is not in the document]
jrha: `absolute_file_path` doesn't care about globs, it just checks that it starts with a leading `/` and doesn't end with one.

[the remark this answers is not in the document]
jrha: See https://github.com/quattor/template-library-core/blob/master/pan/types.pan#L997-L1016

ncm-systemd/src/main/resources/regular/tests/profiles/unitfile.pan
```diff
@@ -69,6 +69,18 @@ bind "/unitfile" = systemd_unitfile_config[];
         'SocketGroup', 'pipegroup',
         'SocketMode', '660',
         ),
+    'path', dict(
```
wpoely86, line 72: you still need to add the expected output of this section to `ncm-systemd/src/main/resources/regular/tests/regexps/unitfile0/value`

jrha:
@wpoely86 I've rebased this and sorted out the tests, so could you re-review please?

wpoely86 (approved):

DavidFair:
Thanks for getting this over the line, I've been pretty wrapped up with leave so it unfortunately got left by the wayside

PR #1586, opened 2023-05-05
[DO NOT MERGE] skeleton nextgen network module

stdweird:


stdweird:
@victor-mendoza @aka7 It might be better to use a construct like this to do new work, without having to worry about current ncm-network code or behaviour

if you want to refactor the "old" code so it is easier to manipulate the behaviour, try to do so in separate PR or at least in single commit

stdweird:
@aka7 @victor-mendoza i also added NetworkManager-config-server as an rpm for more backwards compatible behaviour with initscripts

aka7:
@stdweird what I've been working on is fully to configure using keyfile format, rather than getting NM to work with ifcfg files. I'll leave ifcfg compatibility with @victor-mendoza .
 I started on this,  not yet complete or ready to create PR but your thoughts will be welcomed before I spend anymore time on this. 
https://github.com/quattor/configuration-modules-core/compare/master...aka7:configuration-modules-core:network_nm

stdweird:
@aka7 make a PR and set  it in draft, so i can comment (even in tests fail). but you only need to add the methods with changes vs network.pm. it looks like you made a full copy now.

aka7:
@stdweird  first crack at this. I had copy to methods over to make changes to it.
https://github.com/quattor/configuration-modules-core/pull/1587

PR #1587, opened 2023-05-16
ncm-network: networkmanager module to configure using keyfile format

aka7:
starting a new module to manage networking using networkmanager keyfile format. using most of the existing logic.
- slight tweak to file_dump and process_network to make it work with NM config.
- single interface, and bonding is currenlty supported.
- new nm_manage_dns option is added to schema to decided if NM should manage dns. NM by default manages dns which will intefere with ncm-resolver.
- ethtool,vlan,bridge configuration not yet supported.

NOTE: opening this up for comments and suggestions, not expecting this PR to be merged.  I have no doubt more work will be required on this. I've pulled in any functions that needs modification here purely so I don't make any changes to network.pm.

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -325,7 +325,14 @@ sub file_dump
     # Make hardlinked copy
     #   hardlink is ok, new file will be put in place, no editing of existing file
     if (!$self->file_exists($file) || $self->mk_bu($file, $testcfg)) {
-        my $fh = CAF::FileWriter->new($testcfg, log => $self, keeps_state => 1);
+        my $fh;
+        # files in nmconnection has to be mode 0400 otherwise networkmanager won't read it.
+        if ($file =~ /\.nmconnection$/) {
+            $fh = CAF::FileWriter->new($testcfg, log => $self, mode => 0400, keeps_state => 1);
+        } else {
+            $fh = CAF::FileWriter->new($testcfg, log => $self, keeps_state => 1);
+        };
+        
         print $fh join("\n", @$text, ''); # add trailing newline
 
         # Use 'scheduled' in messages to indicate that this method
@@ -839,6 +846,14 @@ sub process_network
             }
             $iface->{my_inner_ipaddr} .= "/$iface->{my_inner_prefix}";
         }
+
+        # Newly added used when generating the keyfile format config in networkmanager module
+        my $media =  $nic->{media};
+        if ($media) {
+            $iface->{media} = $media;
+            $self->debug(1, "Added media option $media for interface $ifname");
+        }
+
     }
 
     return $nwtree;
--- a/ncm-network/src/main/perl/networkmanager.pm
+++ b/ncm-network/src/main/perl/networkmanager.pm
@@ -2,16 +2,17 @@
 
 =head1 NAME
 
-network: Configure Network Settings using NetworkManager Keyfile format. Most functions and logic is taken from network module. 
+network: Extention of Network to configure Network settings using NetworkManager by configuring in Keyfile format.
+Most functions and logic is taken from network module to minimise changes to current network module.
 
 =head1 DESCRIPTION
 
 The I<network> component sets the network settings through C<< /etc/sysconfig/network >>
-and the keyfile format files in C<< /etc/NetworkManager/system-connections >>.
+and the NM keyfile settings in files C<< /etc/NetworkManager/system-connections >>.
 
 New/changed settings are first tested by retrieving the latest profile from the
-CDB server (using ccm-fetch).
-If this fails, the component reverts all settings to the previous values.
+CDB server (using ccm-fetch). 
+If this fails, the component reverts all settings to the previous values. This is no different to network module.
 
 During this test, a sleep value of 15 seconds is used to make sure the restarted network
 is fully restarted (routing may need some time to come up completely).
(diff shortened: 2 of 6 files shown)


ncm-network/src/main/perl/networkmanager.pm
```diff
@@ -0,0 +1,611 @@
...
+
+our $NoActionSupported = 1;
+
+use Readonly;
+
+Readonly my $HOSTNAME_CMD => '/usr/bin/hostnamectl';
+
+Readonly my $NETWORK_PATH => '/system/network';
+Readonly my $HARDWARE_PATH => '/hardware/cards/nic';
+Readonly my $IFCFG_DIR => "/etc/NetworkManager/system-connections";
```
aka7, line 36: reason for keeping this as $IFCFG_DIR was to see if  when it creates backup in this path, but it still doesn't. creates backup dir in network-scripts location

ncm-network/src/main/perl/network.pm
```diff
@@ -325,7 +325,14 @@ sub file_dump
     # Make hardlinked copy
     #   hardlink is ok, new file will be put in place, no editing of existing file
     if (!$self->file_exists($file) || $self->mk_bu($file, $testcfg)) {
-        my $fh = CAF::FileWriter->new($testcfg, log => $self, keeps_state => 1);
+        my $fh;
+        # files in nmconnection has to be mode 0400 otherwise networkmanager won't read it.
+        if ($file =~ /\.nmconnection$/) {
```
stdweird, line 330: hmm, i don't like copy paste. 
can you test

```
# files in nmconnection has to be mode 0400 otherwise networkmanager won't read it.
my $mode;
$mode = 0400 if ($file =~ /\.nmconnection$/);


my $fh = CAF::FileWriter->new($testcfg, log => $self, keeps_state => 1, mode => $mode);
```

ncm-network/src/main/perl/network.pm
```diff
@@ -839,6 +846,14 @@ sub process_network
             }
             $iface->{my_inner_ipaddr} .= "/$iface->{my_inner_prefix}";
         }
+
+        # Newly added used when generating the keyfile format config in networkmanager module
+        my $media =  $nic->{media};
```
stdweird, line 851: one space before `$nic`

ncm-network/src/main/perl/network.pm
```diff
@@ -839,6 +846,14 @@ sub process_network
             }
             $iface->{my_inner_ipaddr} .= "/$iface->{my_inner_prefix}";
         }
+
+        # Newly added used when generating the keyfile format config in networkmanager module
```
stdweird, line 850: i don't understand the comment

ncm-network/src/main/perl/network.pm
```diff
@@ -839,6 +846,14 @@ sub process_network
...
+        my $media =  $nic->{media};
+        if ($media) {
+            $iface->{media} = $media;
```
stdweird, line 853: is this something we should support in the schema? if you want to set (or is it preserve?) it, you should do this via the schema.

ncm-network/src/test/perl/networkmanager_simple.t
```diff
@@ -0,0 +1,28 @@
...
+
+
+is($cmp->Configure($cfg), 1, "Component runs correctly with a test profile");
+
+# From here, test custom methods
+command_history_reset();
+$cmp->disable_networkmanager(1);
+# check there a executed commands that match NetworkManager
+ok(command_history_ok(undef, ['NetworkManager']));
+
```
stdweird, line 26: some more tests might be good ;)

ncm-network/src/main/perl/networkmanager.pm
```diff
@@ -0,0 +1,724 @@
...
+# need this to remain same value as what is in network.pm
+Readonly my $REMOVE => -1;
+Readonly my $NOCHANGES => 0;
+Readonly my $UPDATED => 1;
+Readonly my $NEW => 2;
+# changes to file, but same config (eg for new file formats)
+Readonly my $KEEPS_STATE => 3;
+
+
+# NOTE: Had to pull in some fucntions from network.pm in order to satisfy location of new paths.
```
stdweird, line 54: this shouldn't be required. if you redefine the readonly here, you should be able to use them in the parent. but the parent might need changes. i can look for an example where i managed to do this in perl.

ncm-network/src/main/perl/networkmanager.pm
```diff
@@ -0,0 +1,724 @@
...
+Readonly my $HARDWARE_PATH => '/hardware/cards/nic';
+Readonly my $NMCFG_DIR => "/etc/NetworkManager/system-connections";
+Readonly my $BACKUP_DIR => "$NMCFG_DIR/.quattorbackup";
+Readonly my $NETWORKCFG => "/etc/sysconfig/network";
+
+Readonly my $RESOLV_CONF => '/etc/resolv.conf';
+Readonly my $RESOLV_CONF_SAVE => '/etc/resolv.conf.save';
+Readonly my $RESOLV_SUFFIX => '.ncm-network';
+
+# need this to remain same value as what is in network.pm
```
stdweird, line 45: you can import the values or perhaps use them straight away. there should be no need for this.

ncm-network/src/main/perl/networkmanager.pm
stdweird: ok, it's coming back.  readonly are real variables, as compared to the `use constant`. i would suggest to make subs like
```
sub _backup_dir { "whatever" }
````

and use `$self->_backup_dir` instead of `$BAKCUP_DIR` here or in the parent `network.pm`

File: ncm-network/src/main/perl/networkmanager.pm [hunk 9]
```diff
@@ -0,0 +1,724 @@
...
+{
+    my ($self, $file) = @_;
+
+    my $back = "$file";
+    $back =~ s/\//_/g;
+
+    return "$BACKUP_DIR/$back";
+}
+
+# Look for existing interface configuration files (and symlinks)
```
stdweird, line 68: what is diffeent here? is it only the NMCFG_DIR?

ncm-network/src/main/perl/networkmanager.pm
```diff
@@ -0,0 +1,724 @@
...
+            $msg = "file";
+            return (undef, undef) if ! defined($self->mk_bu($file));
+        }
+        $self->debug(3, "Found ifcfg $msg $file");
+    }
+
+    return (\%exifiles, \%exilinks);
+}
+
+# Create new backupdir
```
stdweird, line 108: idem, what is different here?

jrha:
Obsoleted by nmstate back-end added in #1601.

PR #1589, opened 2023-05-29
ncm-network: fix gso hash mapping

aka7:
update ethtool option mapping of gso, correct output seen from ethtool --show-offload DEVNAME is generic-segmentation-offload

setting gso in templates won't change this setting. 
see following in ncm log.

[INFO] running component: network
---------------------------------------------------------
[INFO] ethtool_set_iface_options: Skipping setting for eth2/offload/gso to off as not in ethtool

Not aware of any, checked output from ethtool -show-offload  DEVNAME on rhel7, 8 and 9.

### DIFF
--- a/ncm-network/src/main/perl/network.pm
+++ b/ncm-network/src/main/perl/network.pm
@@ -112,10 +112,10 @@ Readonly::Array my @ETHTOOL_OPTION_ROOT => qw(offload ring ethtool);
 
 Readonly::Hash my %ETHTOOL_OPTION_MAP => {
     offload => {
-        tso => "tcp segmentation offload",
+        tso => "tcp-segmentation-offload",
         tx  => "tx-checksumming",
         rx  => "rx-checksumming",
-        ufo => "udp fragmentation offload",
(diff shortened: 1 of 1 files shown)


jrha (dismissed):

jrha:
Thanks! Looks like `ufo` is also affected and should be `udp-fragmentation-offload`

jrha:
And possibly also `tcp-segmentation-offload` :disappointed:

aka7:
> And possibly also `tcp-segmentation-offload` 😞

ah my bad, should have checked them,  happy to add them to this pr.  
although ufo support has been removed and no longer available in rhel8+ anyway

UFO was removed from upstream
[Remove UDP Fragmentation Offload support](https://www.spinics.net/lists/netdev/msg443558.html)
Very few devices support this operation, it's usefullness is
quesitonable at best, and it adds a non-trivial amount of
complexity to our data paths.

https://access.redhat.com/solutions/4594671

jrha (approved):

PR #1593, opened 2023-06-27
workflows: Run unit tests on EL8

jrha:
EL7 is too old of a default environment for us now, although we will still build nightlies on it as long as we have to support it.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   runtests:
     runs-on: ubuntu-latest
-    container: centos:7
+    container: rockylinux:8
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/continuous-integration.yaml
```diff
@@ -14,20 +14,24 @@ jobs:
...
           perl-Data-Compare perl-Date-Manip perl-File-Touch perl-JSON-Any \
           perl-Net-DNS perl-Net-FreeIPA perl-Net-OpenNebula \
           perl-Net-OpenStack-Client perl-NetAddr-IP perl-REST-Client \
-          perl-Set-Scalar perl-Text-Glob cpanminus gcc wget
-        #perl-Git-Repository perl-Data-Structure-Util
-        # Hack around the two missing Perl rpms for ncm-ceph
-        cpanm install Git::Repository Data::Structure::Util
+          perl-Set-Scalar perl-Text-Glob cpanminus gcc wget \
+          perl-Git-Repository perl-Data-Structure-Util \
+          http://yum.quattor.org/devel/perl-Test-Quattor-18.3.0-SNAPSHOT20180406083650.noarch.rpm \
```
ned21, line 33: Is there a reason we need to use such an old version and a snapshot? Would it be worth documenting (e.g. in the commit) log what needs to be fixed in order to move back to an unversioned or non-snapshot release?

jrha: It's the same reason as in the original comments (lines 22-25), but yes it should be sorted out.

wpoely86: ok, let's put in a issue and not hold this back.

wpoely86 (approved):

jrha:
> It's a bit odd that we would have to take a rather old version to be able to update to rockylinux 8.

Yeah, unfortunately right now it's the _only_ version.

PR #1595, opened 2023-06-28
ncm-network core schema: allow realhostname to be a short hostname

jouvin:
- Allow better support of software like Ceph where it is the recommended setting

This change is backward compatible but relaxes the requirements on `realhostname` property. I also checked the `template-library-standard`, `template-library-os`, `template_library_core` and `template-library-grid` repositories, as well as `configuration-modules-core`. `realhostname` is not referenced in any of them (except ncm-network schema definition in`template_library_core`) . Only `hostname` is set in `template-library-standard` but this property is not affected by this PR.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -421,7 +421,7 @@ type structure_ipv6 = {
 type structure_network = {
     "domainname" : type_fqdn
     "hostname" : type_shorthostname
-    "realhostname" ? type_fqdn
+    "realhostname" ? string with is_shorthostname(SELF) || is_fqdn(SELF)
     "default_gateway" ? type_ip
     @{When default_gateway is not set, the component will try to guess the default
       gateway using the first configured gateway set on an interface.
--- a/ncm-network/src/test/perl/simple.t
+++ b/ncm-network/src/test/perl/simple.t
@@ -7,7 +7,7 @@ BEGIN {
 }
 
 use Test::More;
-use Test::Quattor qw(simple simple_ethtool simple_noethtool simple_realhostname simple_nobroadcast simple_tun);
+use Test::Quattor qw(simple simple_ethtool simple_noethtool simple_realhostname simple_shortrealhostname simple_nobroadcast simple_tun);
 use Test::MockModule;
 
 use NCM::Component::network;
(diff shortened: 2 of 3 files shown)


jrha:
I'm not really sure why `realhostname` was added, but it's definitely supposed to be a FQDN, the component uses it like this:

```perl
my $hostname = $nwtree->{realhostname} || "$nwtree->{hostname}.$nwtree->{domainname}";
```

[ncm-network/src/main/perl/network.pm#L1973C5-L1973C91](https://github.com/quattor/configuration-modules-core/blob/master/ncm-network/src/main/perl/network.pm#L1973C5-L1973C91)

jrha:
We don't use `realhostname` at RAL, what's the use-case?

jouvin:
@jrha not sure why you are saying this. What the component does it to define the host name based either on the `hostname` property with the domain name appended or the `realhostname` property used litteraly. `realhostname`, as the name suggests, is something that must be used as it is provided without any processing. The only requirement is that it is a valid host name value and both forms are valid (and as pointed out by the docs I mentionned, FQDN is not favoured when a machine has multiple interfaces/adresses like it is almost always the case on a Ceph machine.

jouvin:
At the end I don't care whether we relax the types allowed or add a new property like `shortrealhostname` (it seems to me it is an overkill if we look at the current usage of `realhostname`). My main point is: the `hostname` and `realhostname` properties are used by the component to  pass a valid value to define the actual host name but with the current schema we don't support all the valid values at the Linux level as there is **no way** to use a short hostname if we want, whatever the reason. IMO, this should be fixed.

jouvin:
@stdweird I tried to add a unit test for short hostname but it seems to fail miserably despite I mainly copied the test for `realhostname`. It is the usual message `  Parse errors: No plan found in TAP output`. Is there an obvious reason? How to get more details on the error? Thanks in advance if you can refresh my memory...

jouvin:
Unit tests are failing in `ncm-spma` so it doesn't seem to be related. `ncm-network` tests are ok.

stdweird:
i once proposed the idea to add methods to ccm to extract the fqdn hostname and make it similar to `realhostname || hostname.domainname`, so this could be used in components that now only use `hostname.domainname`.
so i don't care much tbh, we also don;t use it in our profiles. but be aware that some components use the fqdn as hostname.

jrha (approved):

PR #1597, opened 2023-07-03
ncm-shorewall: support audit in shorewall/iptables

stdweird:


ncm-shorewall/src/main/pan/components/shorewall/schema.pan
```diff
@@ -153,6 +155,10 @@ type component_shorewall_rules = {
     "headers" ? string
     "switch" ? string
     "helper" ? string
+    @{use the audit action}
+    "audit" ? boolean
+    @{action loglevel}
+    "loglevel" ? string
```
jrha, line 161: should this be a `choice`?

jrha: Seems not:
> Beginning with Shorewall 4.5.5, the level name or number may be optionally followed by a comma-separated list of one or more log options. The list is enclosed in parentheses. Log options cause additional information to be included in each log message.

jrha (approved):

PR #1604, opened 2023-07-14
ncm-ssh: add option prohibit-password to PermitRootLogin parameter

### LINKED ISSUE #1603: ncm-ssh 21.12.1: prohibit-password value not supported for PermitRootLogin
sshd option `PermitRootLogin` supports the value `prohibit-password` to allow root login through SSH with keys only. The schema doesn't allow this value.

jouvin:
Fixes #1603

### DIFF
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -128,7 +128,7 @@ type ssh_daemon_options_type = {
     "NoneEnabled" ? legacy_binary_affirmation_string
     "PermitEmptyPasswords" ? legacy_binary_affirmation_string
     "PermitRootLogin" ? choice(
-        'yes', 
+        'yes',
         'prohibit-password',
         'without-password',
         'forced-commands-only',
--- a/ncm-ssh/src/main/pan/components/ssh/schema.pan
+++ b/ncm-ssh/src/main/pan/components/ssh/schema.pan
@@ -127,7 +127,18 @@ type ssh_daemon_options_type = {
     "MaxStartups" ? long
     "NoneEnabled" ? legacy_binary_affirmation_string
     "PermitEmptyPasswords" ? legacy_binary_affirmation_string
-    "PermitRootLogin" ? string with match (SELF, '^(yes|without-password|forced-commands-only|no)$')
+    "PermitRootLogin" ? choice(
+        'yes', 
+        'prohibit-password',
+        'without-password',
+        'forced-commands-only',
+        'no'
+    ) with {
+        if (SELF == 'without-password') {
+            deprecated(0, '"without-password" is deprecated and should be updated to "prohibit-password"');
+        };
+        true;
+    }
     "PermitTunnel" ? string with match (SELF, '^(yes|point-to-point|ethernet|no)$')
     "PermitUserEnvironment" ? legacy_binary_affirmation_string
     "PidFile" ? string


ned21 (dismissed):
It would be good to switch this from a string to a `choice` type.

stdweird:
fyi @jouvin @ned21 we have switched to https://github.com/quattor/configuration-modules-core/pull/1452 completely

jouvin:
@stdweird I missed that... It would be good to mark it as deprecated in some way (may be I missed it).... I'll update our configs then.

ned21:
Having a separate component has some advantages. In our setup we use filecopy to deliver some .tt files as it takes too long to deploy an rpm, and that means metaconfig depends on filecopy. In some cases we've seen some config have filecopy declare a dependency on a third component and if either filecopy or that 3rd component has any errors, then metaconfig doesn't run and you end up with a machine without sshd, making troubleshooting tedious. The main risk comes from metaconfig having a pre-dep, so if you have a policy to prohibit that then it's probably OK.

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -127,7 +127,7 @@ type ssh_daemon_options_type = {
     "MaxStartups" ? long
     "NoneEnabled" ? legacy_binary_affirmation_string
     "PermitEmptyPasswords" ? legacy_binary_affirmation_string
-    "PermitRootLogin" ? string with match (SELF, '^(yes|without-password|forced-commands-only|no)$')
+    "PermitRootLogin" ? string with match (SELF, '^(yes|prohibit-password|without-password|forced-commands-only|no)$')
```
jrha, line 130:
```suggestion
    "PermitRootLogin" ? choice('yes', 'prohibit-password', 'without-password', 'forced-commands-only', 'no')
```

jrha (dismissed):

ncm-ssh/src/main/pan/components/ssh/schema.pan
jrha: Also perhaps also:
```suggestion
    "PermitRootLogin" ? choice('yes', 'prohibit-password', 'without-password', 'forced-commands-only', 'no') with {
    if (SELF == 'without-password') deprecated(0, '"without-password" is deprecated and should be updated to "prohibit-password"');
    true;
}
```

jouvin:
Thanks @jrha for the excellent suggestion, added to the PR

jouvin:
I had to remove the last comma in the `choice` to make panc happy! With the comma, there is syntax error (panc 10.7):

```
[panc-check-syntax] Encountered: )
[panc-check-syntax] Was expecting one of:
[panc-check-syntax]     <HEREDOC_STRING> ...
[panc-check-syntax]     <SINGLE_QUOTED_STRING> ...
[panc-check-syntax]     <DOUBLE_QUOTED_STRING> ...
```

ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -127,7 +127,18 @@ type ssh_daemon_options_type = {
     "MaxStartups" ? long
     "NoneEnabled" ? legacy_binary_affirmation_string
     "PermitEmptyPasswords" ? legacy_binary_affirmation_string
-    "PermitRootLogin" ? string with match (SELF, '^(yes|without-password|forced-commands-only|no)$')
+    "PermitRootLogin" ? choice(
+        'yes',
```
jrha, line 131: Really minor, but there appears to be a trailing space here:
```suggestion
        'yes',
```

jouvin:
@jrha sorry for missing the trailing whitespaces... I accepted your change. I'm lewfor one month of holidays with limited connectivity and no computer! So feel free to fix problems!

jrha:
@jouvin no problem, thanks for the fix! Enjoy your holiday!

PR #1614, opened 2023-08-09
ncm-network: nmstate support for configuring  dummy interfaces

aka7:
providing support for configuring dummy interfaces using nmstate module.

internally we use frr  to configure vips and routing, ffr configuration are using pan, this change is to allow ncm-network to create the loopback interface for the vips defined.  I don't see anything upstream about creating this type of interface, so schema change should not impact anything in network.pm.

 none, change is to nmstate.pm which shouldn't impact current network module.


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -101,6 +101,24 @@ type structure_interface_alias = {
...
+#
+############################################################
+@documentation{
+    Define vip interfaces for configuring loopback interface. Used with frr/zebra configuration
+}
+
+type structure_vip = {
+    "interfaces" : valid_interface[]
+    "ip"         : type_ip
+    "fqdn"       ? string
```
jrha, line 117:
```suggestion
    "fqdn"       ? type_fqdn
```

aka7: done.

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -231,6 +231,34 @@ sub nmstate_file_dump
     }
 }
 
+# given ip/netmask address return cidr mask length
+sub get_masklen {
+    my ($self, $ip_netmask) = @_;
+    $self->verbose("Converting $ip_netmask to cidr notation");
```
ned21, line 237: Is this relevant to an SA or just the programmer? Feels like debug would be more appropriate here

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -583,6 +610,21 @@ sub Configure
         $self->ethtool_opts_keeps_state($file_name, $ifacename, $iface, $exifiles);
     }
 
+    # configure vips defined under path /system/network/vips/ as dummy interface
+    # only if manage_vips is set to true.
+    my $vips = $net->{vips};
```
ned21, line 615: It's been a while since I wrote Perl, does it warn if a key is missing? I would have thought changing the order here would result in less chance of this emitting spurious warnings:

```
if ($net->{manage_vips} && defined($net->{vips})) {
  my $vips = $net->{vips};
  $self->verbose("Service address vips found, configuring dummy interfaces");
```

aka7: doesn't give a WARN, and an empty string in perl is considered to be false from what I read, AFAIK.

ncm-network/src/main/perl/nmstate.pm
aka7: yes agree. updated.

jrha (approved):

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -231,6 +231,34 @@ sub nmstate_file_dump
...
+# return hashrefs needed by nmstate interface config
+sub generate_vip_config {
+    my ($self, $iface_name, $vip) = @_;
+    my %dummy_iface;
+    my $ip_list = {};
+    my $netmask = $vip->{netmask} || "255.255.255.255";
+    my $ip = $vip->{ip};
+    $ip_list->{ip} = $ip;
+    $ip_list->{'prefix-length'} = get_masklen($self, "$ip/$netmask");
+    $dummy_iface{name} = $iface_name;
```
stdweird, line 252: you don't have to write ugly perl code ;)

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -231,6 +231,34 @@ sub nmstate_file_dump
...
+    $dummy_iface{'profile-name'} = $iface_name;
+    $dummy_iface{type} = "dummy";
+    $dummy_iface{state} = "up";
+    $dummy_iface{ipv4}->{enabled} = "true";
```
stdweird, line 256: if you want a yaml boolean out of this, use the yaml constant for true. what does this geneate? the string true incl quotes, that evaluates to true by accident?

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -231,6 +231,34 @@ sub nmstate_file_dump
...
+    $dummy_iface{ipv4}->{address} = [$ip_list];   
+    my $iface_cfg->{'interfaces'} = [\%dummy_iface];
```
stdweird, line 258: the single quotes are not needed when the key is unambigious, so here is. (`profile-name` isn't, perl thinks you are subtracting 2 function calls in that case)

so you can do
```
my $iface_cfg->{interfaces} = [{
  'profile-name' => $iface_name,
  type -> "dummy",
  state => "up",
  ipv4 => {
      enabled => $YTRUE,
      address => [$iplist],
  },
}]
```

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -583,6 +610,21 @@ sub Configure
...
     }
 
+    # configure vips defined under path /system/network/vips/ as dummy interface
+    # only if manage_vips is set to true.
+    if ($net->{manage_vips} && defined($net->{vips})) {
+        $self->verbose("Service address vips found, configuring dummy interfaces");
+        my $vips = $net->{vips};
+        my $idx = 0;
+        foreach my $name (sort keys %$vips) {
+            my $dummy_name="dummy$idx";
```
stdweird, line 620: spaces around `=`

but euhm, all dummy vip names might changes by adding a single interface with name that is earlier in alphabetic order? (i have no easy fix besides chagin the schema to a list, but just be aware of this "feature")

ncm-network/src/main/perl/nmstate.pm
stdweird: @aka7 can you fix this in some other PR?

aka7: yes, will go through all your comments and will do a new pr to fix all that. np.
Thanks for you review @stdweird  appreciate it.

PR #1623, opened 2023-08-22
ncm-network: add two  missing schema types

aka7:
add missing gso for ethtool offload
add priority used in  nmstate.pm

to able to set gso for offload and priority values for configuring network manager policy rule.

*None

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -74,6 +74,8 @@ type structure_rule = {
     "not" ? boolean
     @{routing table action}
     "table" ? network_valid_routing_table
+    @{priority, The priority of the rule over the others. Required by Network Manager when setting routing rules.}
+    "priority" ? long(0..0xffffffff)
     @{rule add options to use (cannot be combined with other options)}
     "command" ? string with !match(SELF, '[;]')
 } with {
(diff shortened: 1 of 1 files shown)


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -74,6 +74,8 @@ type structure_rule = {
     "not" ? boolean
     @{routing table action}
     "table" ? network_valid_routing_table
+    @{priroty also known as metric, needed by nm when setting routing rules)}
```
jrha, line 77:
```suggestion
    @{priority, also known as metric. Represents number of hops to the destination. Required by Network Manager when setting routing rules.)}
```
I assume this not required by other back-ends, so there is no need to make it required in the schema?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -74,6 +74,8 @@ type structure_rule = {
...
+    "priority" ? long
```
jrha, line 78: This should at least be a positive integer, experimentally it appears to be limited to 32-bits.
```suggestion
    "priority" ? long(0..0xffffffff)
```

ncm-network/src/main/pan/components/network/core-schema.pan
aka7: correct, only used in nmstate.  default is set to 100 in the code, I didnt want to set required on schema because changing existing profiles to network backend, what do you think?

also, sorry comment wasn't correct, I fixit now, its not metric, its actually the priority of the rule over the others, so you get something like this in rule. when setting any policy rule/rourte, if that makes sense.

# ip rule list
0:      from all lookup local
100:    from 1.2.1.0/24 lookup 3 proto static
200:    from all to 1.2.3.0/24 lookup 3 proto static
32766:  from all lookup main
32767:  from all lookup default

PR #1624, opened 2023-08-23
ncm-network: improvements made to perl code to construct dummy interface hash

aka7:
improvements made to perl code to construct dummy interface hash. Suggested in previous pr review https://github.com/quattor/configuration-modules-core/pull/1614. 
no functional change.

Improve code.

none.

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -248,14 +248,17 @@ sub generate_vip_config {
     my $netmask = $vip->{netmask} || "255.255.255.255";
     my $ip = $vip->{ip};
     $ip_list->{ip} = $ip;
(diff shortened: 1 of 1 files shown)


ncm-network/src/main/perl/nmstate.pm
```diff
@@ -248,14 +248,17 @@ sub generate_vip_config {
...
+    my $iface_cfg->{interfaces} = [{
+        'profile-name' => $iface_name,
+        name => $iface_name,
+        type => "dummy",
+        state => "up",
+        ipv4 => {
+            enabled => $YTRUE,
+            address => [$ip_list],
+            },
+        }];
```
jrha, line 261:
```suggestion
        },
    }];
```

jrha (approved):

PR #1625, opened 2023-08-24
tests: update dockerfile to rockylinux:8

wpoely86:
This updates the dockerfile to be on sync with what we do in the github actions.

@aka7 build this with podman:
```
podman build -t quattor_test .
```
And then use it like:
```
podman run --user=$(id -ur):$(id -gr)  -it --userns=keep-id -v /path/to/configuration-modules-core:/quattor_test:z -e MVN_ARGS="-pl ncm-network" localhost/quattor_test
```
This runs the tests on your checkout.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,34 +1,35 @@
-# Use an official centos image as a parent image
-FROM centos:7
+# Use an official RockyLinux image as a parent image
+FROM rockylinux:8
 
 # Set the working directory to install dependencies to /quattor
 WORKDIR /quattor
 
 # install library core in /quattor, tests need it
-ADD https://codeload.github.com/quattor/template-library-core/tar.gz/master /quattor/template-library-core-master.tar.gz
-RUN tar xvfz template-library-core-master.tar.gz
+ADD https://codeload.github.com/quattor/template-library-core/tar.gz/master template-library-core-master.tar.gz
+RUN tar -xzf template-library-core-master.tar.gz
 
-# Install dependencies
-RUN yum install -y maven epel-release
-RUN rpm -U http://yum.quattor.org/devel/quattor-release-1-1.noarch.rpm
+# point library core to where we downloaded it
+ENV QUATTOR_TEST_TEMPLATE_LIBRARY_CORE /quattor/template-library-core-master
+
+# Prepare to install dependencies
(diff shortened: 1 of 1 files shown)


jrha:
Could we publish the container this builds and then use it as the base for all of our GitHub actions? It would speed things up.

Dockerfile
```diff
@@ -1,34 +1,35 @@
...
-# point library core to where we downloaded it
-ENV QUATTOR_TEST_TEMPLATE_LIBRARY_CORE /quattor/template-library-core-master
+RUN dnf install -y maven which rpm-build panc ncm-lib-blockdevices \
+  ncm-ncd git libselinux-utils sudo perl-Crypt-OpenSSL-X509 \
+  perl-Data-Compare perl-Date-Manip perl-File-Touch perl-JSON-Any \
+  perl-Net-DNS perl-Net-FreeIPA perl-Net-OpenNebula \
+  perl-Net-OpenStack-Client perl-NetAddr-IP perl-REST-Client \
+  perl-Set-Scalar perl-Text-Glob cpanminus gcc wget \
+  perl-Git-Repository perl-Data-Structure-Util \
+  http://yum.quattor.org/devel/perl-Test-Quattor-18.3.0-SNAPSHOT20180406083650.noarch.rpm \
```
jrha, line 31: This should be included in releases for EL7 and EL8 now, so you can for example use the version from 23.6.0.

jrha, line 31:
```suggestion
  perl-Test-Quattor \
```

wpoely86:
> Could we publish the container this builds and then use it as the base for all of our GitHub actions? It would speed things up.

I believe so. There is ghcr.io. We already use it for other containers and let it (cron wise) weekly rebuild with github actions.

Do you want to keep it in this repo?

jrha:
Maybe it'd be better off replacing the [Dockerfile](https://github.com/quattor/release/blob/master/Dockerfile) in the release repo?

wpoely86:
> Maybe it'd be better off replacing the [Dockerfile](https://github.com/quattor/release/blob/master/Dockerfile) in the release repo?

Yes, let me put it there and remove it here.

jrha:
Cool, thanks! :grinning:

jrha (approved):

PR #1628, opened 2023-09-13
ncm-network: nmstate: handle command routes and rule cleanly

### LINKED ISSUE #1627: ncm-network: nmstate: Doesn't handle "command" routes cleanly
`nmstate` itself doesn't appear to support them, which is probably fine, but the module tries to add them anyway, which results in an invalid config.

For example, we have a number of hosts with `unreachable` routes set:
```pan
"/system/network/interfaces/eth0/route/24/command" = "unreachable 10.0.0.0/8";
"/system/network/interfaces/eth0/route/25/command" = "unreachable 192.168.0.0/16";
```

The module doesn't deal with the missing address, netmask etc and writes out the following:
```yaml
  - destination: /32
    next-hop-interface: eth0
  - destination: /32
    next-hop-interface: eth0
```

Which results in the following from `nmstatectl apply`:
```
NmstateError: InvalidArgument: Invalid IP address : invalid IPv4 address syntax
```

Desired behaviour here is probably:
1. Try and catch these with a validation rule in the schema when `nmstate` is in use.
1. Ignore these kinds of rules at runtime with a suitable warning.

aka7:
nmstate dossn't appear to handle route or rule commands. ignore any routes or rule which sets command with a warn.

Fixes #1627

see https://github.com/quattor/configuration-modules-core/issues/1627

when using Nmstate backend,  any routes and rules added with command will be ignored by Nmstate.

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -108,6 +108,10 @@ sub make_nm_ip_rule
 
     my @rule_entry;
     foreach my $rule (@$rules) {
+        if ($rule->{command}){
+            $self->warn("Rule command entry not supported with nmstate, ignoring '$rule->{command}'");
+            next;
+        };
         my %thisrule;
         my $priority = 100;
         $priority = $rule->{priority} if $rule->{priority};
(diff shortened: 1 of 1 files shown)


aka7:
this will ignore such routes and rules for Nmstate for now, but should still try and catch these with a validation rule in the schema when nmstate is in use but I'm not sure exactly how to do this yet.

I will also have chat with RedHat regarding Nmstate support for unreachable type. networkmanager does seem to have option for it, but can't see how to set this using Nmstate.

jrha:
Thanks @aka7, schema fix is in #1631, the command routes have always been a last resort to work around missing functionality so maybe `unreachable` should be supported in a different way in future.

jrha (approved):
LGTM

PR #1630, opened 2023-09-13
ncm-network: fix nmstate package definition in Kickstart config

### LINKED ISSUE #1629: ncm-network nmstate: invalid package definition in Kickstart configuration
In version 23.6.0, the addition to Kickstart of the package providing nmstate is invalid (not consistent withe the schema).

Michel

jouvin:
Fixes #1629

### DIFF
--- a/ncm-network/src/main/pan/components/network/config-nmstate.pan
+++ b/ncm-network/src/main/pan/components/network/config-nmstate.pan
@@ -13,4 +13,4 @@ prefix "/software/components/network";
 prefix '/software/packages';
 'nmstate' = dict();
 prefix "/system/aii/osinstall/ks";
-"NetworkManager-config-server" = dict();
+'packages' = append("NetworkManager-config-server");


aka7 (approved):

ncm-network/src/main/pan/components/network/config-nmstate.pan
```diff
@@ -13,4 +13,4 @@ prefix "/software/components/network";
 prefix '/software/packages';
 'nmstate' = dict();
 prefix "/system/aii/osinstall/ks";
-"NetworkManager-config-server" = dict();
+'packages' = append("NetworkManager-config-server");
```
jrha, line 16: Probably unnecessary, but for extra safety:
```suggestion
'packages' ?= list();
'packages' = append(SELF, 'NetworkManager-config-server');

```

stdweird: that should not be needed at all ever. single argument append does it all.

jrha: True, but the explicit `SELF` guards against users adding more items to the append.
Anyway, it's perfectly fine and now merged.

PR #1631, opened 2023-09-13
ncm-network: Throw error if commands used with nmstate

jrha:
This is the schema part of the fix for #1627.


aka7 (approved):
LGTM
great team work :) Thank you James.

Issue #1643, opened 2023-12-29
ncm-network: nmstate backend policy routing removal does not work without downing the interface.


aka7:
Currently the nmstate backend will run 'nmstatectl apply  /etc/nmstate/$interface.yml'

this works fine, except when there is config change where a policy route that was previously configured is no longer in config.  nmstate does not appear to remove them. Looks like nmstatectl apply just appends.

Only way this appears to work to down connection,  then apply the config.

* The version of the component you are using (rpm -q $component)
ncm-network-23.6.0-1

* What OS version you are using (/etc/redhat-release or similar)
```
# cat /etc/redhat-release
Red Hat Enterprise Linux release 9.3 (Plow)
```

* Steps required to reproduce the problem
add two 2 policy route to an interface so yaml file looks like this for routes. i.e
```
routes:
  config:
  - destination: 0.0.0.0/0
    next-hop-address: 10.10.10.65
    next-hop-interface: bond1
    table-id: '10'
  - destination: 10.10.10.64/28
    next-hop-interface: bond1
    table-id: '10'
  - destination: 10.10.12.0/24
    next-hop-interface: bond1
    table-id: '10'
```
3 routes added to table 10.

```
# nmstatectl apply /etc/nmstate/bond1.yml
# ip route show table 10
default via 10.10.10.65 dev bond1 proto static
10.10.10.64/28 dev bond1 proto static scope link
10.10.12.0/24 dev bond1 proto static scope link
```

now removed the route for dest 10.10.12.0/24 so file looks like this, 
```
routes:
  config:
  - destination: 0.0.0.0/0
    next-hop-address: 10.10.10.65
    next-hop-interface: bond1
    table-id: '10'
  - destination: 10.10.10.64/28
    next-hop-interface: bond1
    table-id: '10'
# nmstatectl apply /etc/nmstate/bond1.yml
```

nothing changes, removed route 10.10.10.0/24 still there.
```
# ip route show table 10
default via 10.10.10.65 dev bond1 proto static
10.10.10.64/28 dev bond1 proto static scope link
10.10.12.0/24 dev bond1 proto static scope link
```

down the interface first then apply
```
# nmcli conn down bond1
# nmstatectl apply /etc/nmstate/bond1.yml
# ip route show table 10
default via 10.10.10.65 dev bond1 proto static
10.10.10.64/28 dev bond1 proto static scope link
```

[...]

PR #1644, opened 2023-12-29
ncm-network: nmstate - down the interface connection before applying

aka7:
take down the connection using nmcli before applying changes.

nmstatectl apply does not replace running config instead it seems to append. This is not useful when settings have been removed such as policy routing. Therefore down the interface connection during configuration changes. This will be now similar to original main ncm network module where it runs ifdown and ifup during config changes.
#1643 

Get correct configs applied after a change to profile
 
None

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -518,6 +518,8 @@ sub nmstate_apply
             # apply config using nmstatectl
             my $ymlfile = $self->iface_filename($iface);
             if ($self->any_exists($ymlfile)){
+                $self->verbose("Applying config for interface $iface");
+                push(@cmds, [$NMCLI_CMD, "conn", "down", $iface ]) if $self->is_active_interface($iface);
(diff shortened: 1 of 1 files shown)


ncm-network/src/main/perl/nmstate.pm
```diff
@@ -518,6 +518,8 @@ sub nmstate_apply
             # apply config using nmstatectl
             my $ymlfile = $self->iface_filename($iface);
             if ($self->any_exists($ymlfile)){
+                $self->verbose("Apply config for inteface $iface");
```
ned21, line 521: `Apply -> Applying` to be consist with e.g. line 515
and typo: inteface -> interface.

aka7:
nmstate allows a way clear routes during apply. replaced with.
https://github.com/quattor/configuration-modules-core/pull/1645
so I think we can close this unless anyone thinks we should also down the interface before applying, just like in ifcfg approach? would like some input in this?

aka7:
closing PR in favor of one, https://github.com/quattor/configuration-modules-core/pull/1647

PR #1645, opened 2024-01-03
ncm-network: nmstate - remove config route entry not in profile

aka7:
nmstatectl apply does not replace running config instead merges current state config with desired state. This is not useful when settings have been removed such as policy routing.

'state: absent' will remove config routes entry that matches the interface as next hop and therefore will only apply the routes configured in host profile profile.

Allow capability to  change policy routes or adding static routes after host is built and without having to rebuild the host.

None.

#1643

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -159,6 +159,16 @@ sub make_nm_ip_route
     return \@rt_entry;
 }
 
+# create an absent route entry.
+# if you prepend the routes with the 'absent', then nmstate will clear the existing matches and apply the routes
+# This will allow nmstate to clear all routes for the interface and only apply routes defined in config.
+# useful when routes are changed later on in profile once host is built.
+# return arrayref
+sub make_nm_route_absent {
+    my ($self, $device) = @_;
+    return { 'state' => 'absent', 'next-hop-interface' => $device };
+}
+
 # group all eth bound to a bond together in a hashref for to be used as
 # - port in nmstate config file
 sub get_bonded_eth
@@ -377,13 +387,13 @@ sub generate_nmstate_config
     #     next-hop-interface:
     #  and so on.
     my $routes = [];
+    push @$routes, ($self->make_nm_route_absent($name));
+    push @$routes, \%default_rt if scalar %default_rt;
     if (defined($iface->{route})) {
         $self->verbose("policy route found, nmstate will manage it");
         my $route = $iface->{route};
-        $routes = $self->make_nm_ip_route($name, $route, $routing_table);
-        push @$routes, \%default_rt if scalar %default_rt;
-    } elsif (scalar %default_rt){
-        push @$routes, \%default_rt if scalar %default_rt;
+        my $policyroutes = $self->make_nm_ip_route($name, $route, $routing_table);
+        push @$routes, @{$policyroutes};
     }
 
     my $policy_rule = [];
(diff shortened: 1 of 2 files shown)


ncm-network/src/main/perl/nmstate.pm
```diff
@@ -159,6 +159,22 @@ sub make_nm_ip_route
...
+# create an absent route entry.
+# if you prepend the routes with the 'absent', then nmstate will clear the existing matches and apply the routes
+# This will allow nmstate to clear all routes for the interface and only apply routes defined in config.
+# useful when routes are changed later on in profile once host is built.
+# return arrayref
+sub make_nm_route_absent {
+    my ($self, $device) = @_;
+
+    my @rt_entry;
+    my %rt;
```
ned21, line 171: This is fine but doesn't feel very Perl-ish. My Perl is rusty but I think you want:

```
my $rt_hf = { 'state' => 'absent', 'next-hop-interface' => $device };
my @rt_entry = ($rt_hf);
```

Or I think you can do it in one:

```
return [ { 'state' => 'absent', 'next-hop-interface' => $device } ];
```

cf. https://stackoverflow.com/questions/17165675/how-to-reference-a-hash-in-perl and https://stackoverflow.com/questions/50155116/create-an-array-ref-in-one-line-in-perl

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -377,13 +393,13 @@ sub generate_nmstate_config
     #     next-hop-interface:
     #  and so on.
     my $routes = [];
+    push @$routes, @{$self->make_nm_route_absent($name)};
```
ned21, line 396: Making an array_ref just to turn it back into a real array here seems odd.

I haven't seen what other similar make_nm_ subroutines do but from a complexity POV, could you return a hashref from make_nm_route_absent
```return { 'state' => 'absent', 'next-hop-interface' => $device };```

and then this becomes:

```
push @$routes, ($self->make_nm_route_absent($name));
```

File: ncm-network/src/main/perl/nmstate.pm [hunk 3]
```diff
@@ -377,13 +393,13 @@ sub generate_nmstate_config
...
+    push @$routes, \%default_rt if scalar %default_rt;
```
ned21, line 397: Bit weird that at line 396 you are pushing on a single entry array containing a hashref but here you are appending a hashref instead. What does @$routes actually expect in it?

aka7: @$routes, it contains array of hashref.
so will contain values such as 
```
- next-hop-interface: bond0
  state: absent
- destination: 0.0.0.0/0
  next-hop-address: x.x.x.x
  next-hop-interface: bond0
```

ncm-network/src/main/perl/nmstate.pm
aka7: good shout, no idea why I didn't think of doing it like this. thanks :)

ncm-network/src/main/perl/nmstate.pm
ned21: Makes sense, so I think that means the previous line should be:
push @$routes, $self->make_nm_route_absent($name);

otherwise you're trying to append an array of hashrefs not a hashref. I _suspect_ Perl flattens the array or otherwise does something weird so I t still works but I think it's better to not rely on that.

ned21: I read `perldoc -f push` and it says that it appends each element in the second list to the first.  So line 390 is technically correct: the second arg must be a list. But I always recall using push like: `push @a, $b;` so I guess it works to have a scalar as the second argument. Perl is weird.

ned21:
LGTM but I'll let someone who actually understands the nm give the approval!

aka7:
closing PR in favour of one, https://github.com/quattor/configuration-modules-core/pull/1647

PR #1646, opened 2024-01-05
ncm-network: nmstate - add additional route rule parameters

aka7:
provide additional route rule parameters for nmstate config as defined in https://nmstate.io/devel/yaml_api.html#routes

Also add a default absent rule to clear route rules entries for the table. nmstate by default will merge rules therefore won't clear if there are any old rules already present. This will make sure only rules defined by profile is present.

Enhancement to policy based routing.
None.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -77,18 +77,28 @@ type structure_rule = {
     "table" ? network_valid_routing_table
     @{priority, The priority of the rule over the others. Required by Network Manager when setting routing rules.}
     "priority" ? long(0..0xffffffff)
+    @{nmstate-action used by nmstate module}
+    "nmstate-action" ? choice('blackhole', 'prohibit', 'unreachable')
+    @{nmstate-state used by nmstate module, Can only set to absent for deleting matching route rules}
+    "nmstate-state" ? choice('absent')
+    @{nmstate-iif used by nmstate module, Incoming interface name}
+    "nmstate-iff" ? string
+    @{nmstate-fwmark used by nmstate module. Select the fwmark value to match}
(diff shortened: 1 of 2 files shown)


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -77,18 +77,28 @@ type structure_rule = {
     "table" ? network_valid_routing_table
     @{priority, The priority of the rule over the others. Required by Network Manager when setting routing rules.}
     "priority" ? long(0..0xffffffff)
+    @{action used by nmstate module}
+    "action" ? choice('unreachable', 'blackhole', 'prohibit')
```
ned21, line 81: Would it make sense to sort the choices alphabetically? (b, p, u) ?

If this is exclusive to nmstate and ignored by other methods, use a nmstate- prefix (i.e. nmstate-action) for clarity?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -77,18 +77,28 @@ type structure_rule = {
...
+    @{state used by nmstate module, Can only set to absent for deleting matching route rules}
+    "state" ? choice('absent')
```
ned21, line 83: ditto about marking this as specific to nmstate via a prefix.

ncm-network/src/main/pan/components/network/core-schema.pan
aka7: make sense.

ned21:
Thanks for making the changes. I see there's now a merge commit in the branch. Our preference is for a linear history: any chance you can use rebase to bring the branch up to date with main instead of merge?

aka7:
closing PR in favor of one, https://github.com/quattor/configuration-modules-core/pull/1647

PR #1647, opened 2024-01-11
ncm-network: nmstate enhancements

### LINKED ISSUE #1643: ncm-network: nmstate backend policy routing removal does not work without downing the interface.

aka7:
Add the following enhancements:

`nmstatectl` apply does not replace running config, it instead merges current state config with desired state. This is not useful when settings have been removed such as routing and policy routing and rules.

`state: absent` will remove config routes entry that matches the interface as next hop and therefore will only apply the routes configured in host profile profile. so following changes added.

1. absent entry added when route for that interface. will clean all routes matching that interface when applying:
    - replaces #1644
    - replaces #1645
    - fixes #1643

2. absent entry add route-rules to match any rules with the table id and clear it when applying rules. 
    - replaces #1646
4.  made nmstate mimic bootproto setting of network.pm
     - replaces #1653
3. refactored configuration  settings of each interfaces type.
6. add more advance testing for interface configuration, bonding and valn.


aka7:
@ned21 thank your for review to the recent pr's I created,  I thought its best just do one PR because the changes are related, especially with the testing that I've added, it requires all 3 commits.  first 2 you already reviewed. hope that is ok. All changes are now in this one.

@jrha please can this be looked into in your own time.

aka7:
any of the maintainers are able to review this? I like to know if I need make any changes before I do a internal release please. or does it look good to you? @jrha @stdweird ?

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -77,18 +77,28 @@ type structure_rule = {
...
+    @{nmstate-fwmask used by nmstate module. Select the fwmask value to match}
+    "nmstate-fwmask" ? string
     @{rule add options to use (cannot be combined with other options)}
     "command" ? string with !match(SELF, '[;]')
 } with {
+    module = value('/software/components/network/ncm-module', '');
     if (exists(SELF['command'])) {
-        module = value('/software/components/network/ncm-module', '');
-        if (module == 'nmstate') error("Command routes are not supported by the nmstate backend");
+        if (module == 'nmstate') error("Command rule are not supported by the nmstate backend");
```
jrha, line 95:
```suggestion
        if (module == 'nmstate') error("Command rules are not supported by the nmstate backend");
```

ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -77,18 +77,28 @@ type structure_rule = {
...
+    @{nmstate-action used by nmstate module}
+    "nmstate-action" ? choice('blackhole', 'prohibit', 'unreachable')
+    @{nmstate-state used by nmstate module, Can only set to absent for deleting matching route rules}
+    "nmstate-state" ? choice('absent')
+    @{nmstate-iif used by nmstate module, Incoming interface name}
+    "nmstate-iff" ? string
+    @{nmstate-fwmark used by nmstate module. Select the fwmark value to match}
+    "nmstate-fwmark" ? string
+    @{nmstate-fwmask used by nmstate module. Select the fwmask value to match}
+    "nmstate-fwmask" ? string
```
jrha, line 89: Hmm, I'm not sure I like having nmstate specific stuff in the core schema, but right now I can't think of a better way to do it.

jrha: Acutally, maybe something along the lines of:
```pan
include format('%s-schema', value('/software/components/network/ncm-module', 'basic'));
```

With `nmstate-schema.pan` and `basic-schema.pan` both defining their own version of `structure_rule`?

aka7: if options are clearly stated is that such a big concern but happy to change it.

however, not sure I fully understand it, so that include goes in core-schema?
I tried separating it into own file then include is in corea-schema,  but tests seem to fail with 
`
Creating man page for target/lib/perl/NCM/Component/nmstate.pm, NCM::Component::nmstate.8[INFO]
[INFO] --- exec-maven-plugin:1.4.0:exec (prove) @ network ---
# VERBOSE No TQU cfg file tqu.cfg found
# VERBOSE No main::TQU cfg string found
# VERBOSE Guessed modules to load: NCM::Component::network
# VERBOSE Namespace regex pattern (?^:^(declaration|unique|object|structure)\stemplate\s(\S+);$)
# VERBOSE Found invalid template components/network/basic-schema.pan (expectedname components/network/basic-schema)
# VERBOSE Found matching template components/network/config-nmstate type unique
# VERBOSE Found matching template components/network/config type unique
# VERBOSE Found matching template components/network/core-schema type declaration
# VERBOSE Found invalid template components/network/nmstate-schema.pan (expectedname components/network/nmstate-schema)
# VERBOSE Found matching template components/network/schema type declaration
#   Failed test 'No invalid pan files found in target/pan (no namespace checked for annotations)'
#   at /v/global/user/k/ka/karimabd/github/configuration-modules-core/ncm-network/target/dependency/build-scripts/Test/Quattor/Doc.pm line 204.
#          got: '2'
#     expected: '0'
`
the structure is fine from what I can see.
happy push it as second commit if you want to see it?

and when I tried testing with a host profile with nmstate backend enabled its seems the path isn't available at that stage, so always includes basic-schema.  Clearly I'm probably including at wrong place, so any guidance here will be appreciated.  

to test, tried this, 
`include format('%s-schema', value('/software/components/network/ncm-module')); `
which results in 
`(/software/components/network/ncm-module) doesn't exist`

jrha: The first error just looks like the name in the template didn't match the file name `components/network/nmstate-schema.pan`.

The second error is because the `value()` call has no default fall-back value.

aka7: HI James, 
that's the thing, the name does match filename. I've doubled checked but still throws these errors, it seems any few files I add throws this error. 

yes, the second error is what I've done to see if there is any value of ncm-module even when I set it nmstate, but its still empty. not matter where I include it I can't seem get it to the value so as you say don't think this will work.

the additional nmsate-*  options I added was a good to have, so I can remove them for now if you want to  think about how we would do this, so the other changes can be merged? happy to take on any other suggestions.

jrha: :disappointed: yeah can you remove them for now then and stick them in another PR? The rest of this looks good to go.

aka7: I've removed the nmstate-* options for rules, will create a separate pr for that. 
we do need the action options at some point., so will have to find a solution, perhaps we can discuss on next call. thanks.

aka7:
@jrha made some changes based on what @wpoely86  discovered which I think makes code a bit simpler.
please can you review this please? would be great if we can get this merged.

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -121,7 +122,14 @@ sub make_nm_ip_rule
         $thisrule{'ip-to'} = $rule->{to} if $rule->{to};
         $thisrule{'ip-from'} = $rule->{from} if $rule->{from};
         push (@rule_entry, \%thisrule);
+        
+        # Add a default absent rule to match table defined. This will clear any existing rules for this table, instead of merging.
```
stdweird, line 126: ah, so this is what i hated about nmstate: the state should reflect the config, so if no rule is defined in the config, there should be no rules.
@aka7 thanks for finding this. but feel free to contact upstream developers and ask them if there is a config option to make nmstate "strict" instead us having to generate this in the config.

aka7: I have had chat with Redhat and they suggested this. The absent rules can be set as wildcard too.  I've had a discussion with them to see if they'll introduce an option to just replace instead of merge, but will see how far it goes.

for now though, I think this will be ok since we are adding rules/routes per interface and will clear all for that interface.

If you have any other suggestions happy to discuss it.

jrha (approved):
Looks like a lot of good improvements here, I'll have more of a think about how best to integrate the `nmstate` specific options.

wdpypere:
@aka7 I've been running some tests with this and it's working a lot better now, thanks for the work! However I ran into an issue where nmstate deletes the default gateway :smile: . I've logged it at https://github.com/quattor/configuration-modules-core/issues/1655 because it has little to do with this PR.

aka7:
@jrha is there anything outstanding on this PR or can this be merged?
I have a follow up PR but need to rebase once this is merge before I create it.

Regards

PR #1648, opened 2024-01-17
ncm-resolver: support to allow another local caching server to be used.

aka7:
option to turn off use of dnscache which only configures for djbdns. 
new option allows resolver to still be used to update resolv.conf but with local dns caching client already configured. so usual test to make sure it works before updating resolv.conf to be local.

use resolver outside of dnscache.
none

### DIFF
--- a/ncm-resolver/src/main/pan/components/resolver/schema.pan
+++ b/ncm-resolver/src/main/pan/components/resolver/schema.pan
@@ -24,6 +24,14 @@ type resolver_component = {
     include structure_component
     'servers' : type_ip[..3]
     'search' ? type_fqdn[..6] with { length(replace('(^\[ )|,|( \])$', '', to_string(SELF))) <= 256 }
+    @{dnscache - configure dnscache client, based on djbdns}
     'dnscache' : boolean = false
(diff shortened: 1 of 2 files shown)


aka7:
any of the maintainers are able to review this? I like to know if I need to make any changes before I do a internal release please. or does it look good to you? @jrha does anyone else use dnscache?

jrha:
At first glance this looks fine, we don't use `dnscache`, but do have a number of hosts that have a local resolver and set `/system/network/nameserver/0` to `127.0.0.1`.

aka7:
> At first glance this looks fine, we don't use `dnscache`, but do have a number of hosts that have a local resolver and set `/system/network/nameserver/0` to `127.0.0.1`.

ah, let me try to see if this will work for us, if it does then maybe we won't need this.
we also use actual nameservers in some circumstances to by disabling dnscache.  But I think maybe we can make it work in both circumstances.

jrha:
@aka7 did our solution do what you needed?

aka7:
> @aka7 did our solution do what you needed?
yes, closing

PR #1650, opened 2024-01-30
ncm-metaconfig: httpd - add missing cipher and sslopensslconfcmd

wdpypere:
- added missing cipher `DHE-RSA-CHACHA20-POLY1305`
 - added `opensslconfcmd` so we can set dhparams like: `sslopensslconfcmd DHParameters "/etc/httpd/dhparam.pem"`. Other options are probably also possible.


jrha (approved):
LGTM

Issue #1649, opened 2024-01-30
ncm-opennebula: Add ceph system datastore option


alvarosimon:
Use ceph system DS instead of system shared fs with ceph.

https://docs.opennebula.io/6.8/open_cluster_deployment/storage_setup/ceph_ds.html?highlight=ceph#opennebula-configuration

PR #1651, opened 2024-02-02
ncm-opennebula: Add Ceph system datastore support

alvarosimon:
This includes a new type of ceph datastore in opennebula. The ceph system datastore replaces the current shared system ds for live migrations. It should be the default one for ceph. 

* Fixes: #1649


alvarosimon:
Tested in altaria, now using the new datastore without fuse mount.

jrha (approved):
LGTM

wpoely86 (approved):

PR #1653, opened 2024-02-03
ncm-network: mirror "old" bootproto behaviour in nmstate

wdpypere:
I spent a good time figuring out why `nmstate.pm` was not changing IP's. The code of nmstate is:
```
    if (defined($eth_bootproto)) {
        if ($eth_bootproto eq 'static') {
```

In our templates we don't set `bootproto`.

but if I look in `network.pm` it does:
```
    # set the bootprotocol
    &$makeline('bootproto', def => 'static');

    my $bootproto = $iface->{bootproto} || 'static';
```

So I think `nmstate.pm` should mimic that behavior.

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -282,7 +282,7 @@ sub generate_nmstate_config
     my $iface = $net->{interfaces}->{$name};
     my $device = $iface->{device} || $name;
     my $is_eth = $iface->{set_hwaddr};
-    my $eth_bootproto = $iface->{bootproto};
+    my $eth_bootproto = $iface->{bootproto} || 'static';
     my $is_ip = exists $iface->{ip} ? 1 : 0;
     my $is_vlan_eth = exists $iface->{vlan} ? 1 : 0;
     my $is_bond_eth = exists $iface->{master} ? 1 : 0;


wdpypere:
tests fail because other parts of the code (vlans and bridges) seem to assume bootproto to be unset.  So this would need more fixing. But I do think nmstate should mimic network.pm, to match admin expectations.

aka7:
> tests fail because other parts of the code (vlans and bridges) seem to assume bootproto to be unset. So this would need more fixing. But I do think nmstate should mimic network.pm, to match admin expectations.
this PR should fix the failures you are seeing.
https://github.com/quattor/configuration-modules-core/pull/1647

wdpypere:
ah, I missed that. I'll close this PR. Thanks @aka7

aka7:
@wdpypere I think you will stiill need your PR if you want bootproto to be static by default.
but I guess either  after mine is merged and you rebase? or I can add it with that pr?

aka7:
>> @wdpypere I think you will stiill need your PR if you want bootproto to be static by default. but I guess either after mine is merged and you rebase? or I can add it with that pr?

I've made changes to the pr to include your suggestion and as a result made that part of the code a bit more simpler.  Please pull this in to test if you like. https://github.com/quattor/configuration-modules-core/pull/1647 @wdpypere and a review would be welcomed.

@jrha please can you review this please? would be great if we can merged.

wdpypere:
@aka7 thanks for adding to the other PR. Ill read it, but I don't think you want my review, I'm very bad at Perl. :D

PR #1656, opened 2024-02-18
ncm-network: nmstate dgw fix

### LINKED ISSUE #1655: ncm-network: nmstate default gateway issue
@aka7 I found another discrepancy with nmstate, with gateways. on our site we only set `/system/network/default_gateway`. something like:
```
+-/system/network
  $ default_gateway : '10.141.10.250'
  $ domainname : 'domain'
  $ hostname : 'hostname'
  +-interfaces
    +-eth0
      $ broadcast : '10.141.31.255'
      $ driver : 'bnx2'
      $ ip : '10.141.10.62'
      $ netmask : '255.255.224.0'
      +-route
        +-0
...
```

but in the code I see:
```
    if ($default_gw) {
        # create only default gw entry if gw entry match interface gateway defined                                                                           
        # otherwise this interface is not the default gw interface.                                                                                          
	if ((defined($iface->{gateway})) and ($iface->{gateway} eq $default_gw)) {
            $default_rt{destination} = '0.0.0.0/0';
            $default_rt{'next-hop-address'} = $default_gw;
            $default_rt{'next-hop-interface'} = $device;
	}
    }
```

the `defined($iface->{gateway})` always fails, so no default gateway is set. My perl is quite poor but it seems the code makes the assumption that a gateway is set on interface level. while network.pm does not. Unless I'm mistaken.

I would expect a route like the following to show up:
```
Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
0.0.0.0         10.141.10.250   0.0.0.0         UG    0      0        0 eth0
```

aka7:
Fix for setting default gateway in nmstate with correct interface by looking up if gateway falls within subnet boundary.
next-hop-interface is mandatory for nmstate therefore figure out which interface subnet this gateway falls in. 
This will provide backward compatibility with network.pm

closes #1655 

to provide backward compatibility with network.pm when setting default gateway. without having to set gateway address on interface.

None.

### DIFF
--- a/ncm-network/src/main/perl/nmstate.pm
+++ b/ncm-network/src/main/perl/nmstate.pm
@@ -320,6 +320,13 @@ sub find_vlan_id {
     }
     return $vlanid;
 }
+# Check if given ip belongs to a network
+sub ip_in_network {
+    my ($self, $check_ip, $ip, $netmask) = @_;
+    # is the given ip in his ip/netmask.
+    my $subnet = NetAddr::IP->new("$ip", "$netmask");
+    return NetAddr::IP->new("$check_ip")->within($subnet);
+}
 
 # generates the hashrefs for interface in yaml file format needed by nmstate.
 # bulk of the config settings needed by the nmstate yml is done here.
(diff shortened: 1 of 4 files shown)


wdpypere:
@aka7 thanks for the quick work. I can confirm this code now does with default gateways what I would expect!

aka7:
> @aka7 thanks for the quick work. I can confirm this code now does to default gateways what I would expect!

thanks for confirming, I've opened this PR for review now. I think this will be better then what I had before as I assumed every interface will have gateway defined.

Also bear in mind, nmstate module does not provide all functionalities currently provided by network.pm, as many of comments in the nmstate.pm states. @wdpypere

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -296,6 +296,17 @@ sub generate_vip_config {
     return $iface_cfg;
 }
 
+# Check if given ip belongs to a network
+sub ip_in_network {
+    my ($self, $check_ip, $ip, $netmask) = @_;
+    # is the given gw ip in his ip/netmask.
```
stdweird, line 302: remove `gw` in comment, this is about any ip

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -296,6 +296,17 @@ sub generate_vip_config {
...
+    my $thisip = NetAddr::IP->new("$check_ip");
```
stdweird, line 303: why not 
```
my $subnet = NetAddr::IP->new("$ip", "$netmask");
return NetAddr::IP->new("$check_ip")->within($subnet);
```

stdweird:
@aka7 can you also have a look at this?

stdweird (approved):

PR #1661, opened 2024-02-21
ncm-mysql: add an option to use encrypted user password

D3SX:
Add the possibility to store password in an encrypted way inside templates.
The schema is modified (`component_mysql_db_user` and `component_mysql_server_options` types) to add a boolean option `encrypted_adminpwd` which is false by default.

When `true`, the password has to be set in its encrypted form.
The easiest way to generate the encrypted password is the MySQL command `select password('plop');`

These modifications is fully backward compatible and could be use without changing anything in the templates.


jrha (approved):
This is a really nice enhancement, thanks!

PR #1667, opened 2024-03-05
ncm-network: provide backward compatibility for vlan interface config for nmstate.

### LINKED ISSUE #1660: ncm-network: nmstate - provide backward capability support for vlan interface in nmstate.
currently nmstate supports vlan interface configuration if its defined in following format.

```pan
"/system/network/interfaces/eth0.123"
"/system/network/interfaces/eth0.123/vlan"=true
```
see `vlaninterface.pan` in `test/`

Looking at network.pm code, it supports vlan interface if interface name is given with prefix vlan, such as this.
```pan
"/system/network/interfaces/vlan0" = create("defaultinterface");
prefix "/system/network/interfaces/vlan0";
"device" = "eth0.123";
"physdev" = "eth0";
"route/0" = dict("address", "1.2.3.4");
```

we use the format `"/system/network/interfaces/eth0.123"` so not a issue for us, but this may have impact on anyone else who is using vlan in previous release and want to move to nmstate.

Also  nmstate code assumes interface name and device name might always be same in few places, which won't be case here, so may need few fixes on this along the way.

* The version of the component you are using (`rpm -q $component`)
ncm-network-23.6.0
* What OS version you are using (`/etc/redhat-release` or similar)
Red Hat Enterprise Linux release 9.3 (Plow)

switch to use nmstate backend for previously configured vlan host.

interface configured with right vlan id etc in nmstate.

aka7:
improvements to nmstate to provide backward compatibility with network.pm for configuring vlan interface.
At morganstanley, we use path 

closes #1660

```
/system/network/interfaces/eth0.123
/system/network/interfaces/eth0.123/vlan = true

```
..etc

nmstate was development with that in mind, 

This change will now provide backward compatibility support with network.pm, so following config will also work.

```
prefix "/system/network/interfaces/vlan0";
"device" = "eth0.123";
"physdev" = "eth0";
```

 provide backward compatibility with network.pm for configuration vlan interface.

None.


ncm-network/src/main/perl/nmstate.pm
```diff
@@ -296,6 +296,26 @@ sub generate_vip_config {
     return $iface_cfg;
 }
 
+# find vlan id from either the name or device.
+# i.e eth0.123 will return 123. by checking iface name and device
+# returns vlan id.
+sub get_vlanid(){
```
ned21, line 302: I assume this works but replacing the () with a space would be consistent with line 278?

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -296,6 +296,26 @@ sub generate_vip_config {
...
+    my ($self, $iface, $device) = @_;
+    my $vlanid = $iface;
+    # a vlan interface can defined in two ways
+    # interface/name.vlanid/
+    # or interface/name/device=device.vlanid
+    # replace everything up-to and including . to get vlan id of the interface.
+    # favors ifacename.vlanid, then device.vlanid
+    $vlanid =~ s/^[^.]*.//;;
```
ned21, line 310: Why the double ;; ?

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -296,6 +296,26 @@ sub generate_vip_config {
...
+    # a vlan interface can defined in two ways
+    # interface/name.vlanid/
+    # or interface/name/device=device.vlanid
+    # replace everything up-to and including . to get vlan id of the interface.
+    # favors ifacename.vlanid, then device.vlanid
+    $vlanid =~ s/^[^.]*.//;
+    # if vlanid is empty here, lets check if device has vlan id.
+    if ((! $vlanid) && ($device)) {
+        $vlanid = $device;
+        $vlanid =~ s/^[^.]*.//;
```
stdweird, line 314: same issue, make a small privtae sub so you can reuse the same  regex code. or make it i full blown sub?

```
sub _get_vlan_id {
    my ($self, $name) = @_;
    return if ($name =~ m/\.(\d+)$/) $1;  
}
```
(not tested)

aka7: thanks, sorry for bad perl code :). updated.

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -330,10 +355,11 @@ sub generate_nmstate_config
...
         }
     } elsif ($is_vlan_eth) {
-        my $vlan_id = $name;
-        # replace everything up-to and including . to get vlan id of the interface.
-        # TODO: instead of this, should perhaps add valid-id in schema? but may not be backward compatible for existing host entreis, aqdb will need updating?
-        $vlan_id =~ s/^[^.]*.//;;
+        my $vlan_id = $self->find_vlan_id($name, $iface->{device});
+        # if vlan_id is empty, error
+        if (! $vlan_id) {
+            $self->error("Could not find vlan id for vlan device $name");
```
stdweird, line 361: shouldn't you return here? or skip the rest of the code?

aka7: erm not sure actually,  This will still create the config, with no vlan id:  , will it will be empty and therefore nmstate apply will also fail, which will be more clear as what has gone wrong?

stdweird: if you are sure it will fail, it's ok for me

stdweird (approved):

PR #1673, opened 2024-03-18
ncm-network: refactor network core schema

stdweird:
Refactors the network `core-schema` in structure that easier to modify and easier to move to template-library-core.
By default nothing changes: the original core-schema (now renamed to `core-schema-legacy`) is still used.

By setting the variable `variable QUATTOR_TYPES_NETWORK_LEGACY = false;` before the core-schema is included, you can test the new structure. It should compile and there should be no differences in the generated output.

There is a small chance of backwards incompatibility: some of the network types have been renamed from prefix `structure_` to `network_`, so there is a non-zero chance that this conflicts with self defined types or variables.


stdweird:
@ned21 @jrha @jouvin ideally i had avoided to use variables inside declaration templates, but i see no other way to control the rollout of these changes.

i am also not yet convinced the backend selection should be implemented like this, but this is not a regular component schema either.

i hope we can get this merged so we can start testing the refactor at least (and then drop the core-schema)

ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -0,0 +1,166 @@
...
+            error('networkinterface has gateway %s not reachable from ip %s with netmask %s',
+                            SELF['gateway'], SELF['ip'], SELF['netmask']);
+        };
+        if (exists(SELF['broadcast']) && ! ip_in_network(SELF['broadcast'], SELF['ip'], SELF['netmask'])) {
+            error('networkinterface has broadcast %s not reachable from ip %s with netmask %s',
+                            SELF['broadcast'], SELF['ip'], SELF['netmask']);
+        };
+    };
+
+    network_interface_tunnel_validate(SELF);
```
ned21, line 163: Would this make more sense at line 144? ie keep all the function calls together and then the block of code?

ned21 (commented):
It's a lot of change but the new structure looks good. As you say, it's nice to avoid globals but sometimes they are just the right tool for the job. I think there's some clean up / modernisation that could be done on the "modern" schema but that's better done in v2 once we've established it works.

ncm-network/src/main/pan/components/network/types/network/interface.pan
stdweird: i kept the order of tests, but sure, i can move this up

aka7:
@stdweird Just wondering where we are with this?  Are we separating the schemas in approach? 

I ask so I can then update this to provide the additional rules param.
https://github.com/quattor/configuration-modules-core/pull/1659

stdweird:
@aka7 euhm, i just force pushed a rebase on master, and now the PR is closed ;)

stdweird:
@aka7 will reopen.

i have no feedback saying i should modify anything, so waiting for a merge

stdweird:
@aka7 this is merged as part of #1676

jrha:
Whoops, I thought I'd merged this first.

PR #1683, opened 2024-05-03
ncm-network: nmstate change the way dummy interfaces are created.

aka7:
use key instead of index to label dummy interface. This will allow referencing the correct vip elsewhere in pan or external applications. Add tests.

Allow creation of dummy interface in nmstate using labels instead of indexing.

if using nmstate dummy interface, profile name will change from dummy0 to the label given to dumm interface, so if its myvip, it will change from dummy0 to  dummy_myvip.  I believe noone is using this currently. This was feature only added in nmstate.


jrha (approved):

ned21:
@jrha @aka7 - is the plan to squash these fixup commits and remove the merge commit from the branch before merging to keep the history easy-to-read-linear?

jrha (approved):

PR #1685, opened 2024-05-07
fixes #1684: NCM::Component::spma::dnf: Update dnf.pm

### LINKED ISSUE #1684: NCM::Component::spma::dnf does not honor repository option "module_hotfixes"
Ensure the title of this issue starts with the component name followed by a colon, e.g.
> ncm-spma:

The rhel8 host's profile has:
/software/repositories/0/module_hotfixes = true
...however the corresponding /etc/yum.repos.d/whatever.repo file does not include line "module_hotfixes=1"
This is due to the missing option "module_hotfixes" in 
https://github.com/quattor/configuration-modules-core/blob/master/ncm-spma/src/main/perl/spma/dnf.pm#L312

If this is a bug report, remember to include:

   ncm-spma-23.6.0-1.noarch
* What OS version you are using (/etc/redhat-release or similar):
    rhel8.9
* Steps required to reproduce the problem:
    see above
* Actual result versus expected result:
   see above

ulrich1919:
Add repository option "module_hotfixes"

> ncm-spma: add repository option "module_hotfixes"

  See #1684
   None

### DIFF
--- a/ncm-spma/src/main/perl/spma/dnf.pm
+++ b/ncm-spma/src/main/perl/spma/dnf.pm
@@ -309,7 +309,7 @@ sub Configure
             print $fh "baseurl=$urls\n";
         }
 
-        foreach my $attr (qw(enabled priority gpgcheck repo_gpgcheck gpgkey gpgcakey exclude includepkgs skip_if_unavailable)) {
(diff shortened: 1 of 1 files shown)


stdweird:
how odd. we have repo files with module_hotfixes...

jrha (approved):

PR #1689, opened 2024-07-03
ncm-sudo: support all values for lecture option and the lecture_file option

stdweird:
this is backwards incompatible

@jrha needs to be checked if this supported on el7 or ~el6~

### DIFF
--- a/ncm-sudo/src/main/pan/components/sudo/schema.pan
+++ b/ncm-sudo/src/main/pan/components/sudo/schema.pan
@@ -46,7 +46,9 @@ type sudo_default_options = {
     "mail_no_host" ? boolean
     "mail_no_perms" ? boolean
     "tty_tickets" ? boolean
-    "lecture" ? boolean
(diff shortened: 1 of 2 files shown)


ned21:
If I understand the comment correctly, a file with the old true/false values is accepted by sshd. However any site that sets config via quattor to true/false will need to modify all their templates to use once/never instead. That's a one-time sed fixup, but needs to happen in sync with the schema change and the code on the hosts needs to be ready to accept new version before the schema change can happen. 

Is there a way to permit both forms to co-exist for a release or two to provide a smooth transition path? Maybe something like

```
type sudo_lecture_transition = property with {
    if( is_binary(SELF) ) {
        warning("Deprecated ...);
    } else {
        # manual form of "choice" for transition, replace with choice when transition complete
        if( !match(SELF, '^(always|once|never$') ) {
            error(...);
        }
   }
   true;
};
```

Issue #1690, opened 2024-07-06
ncm-ssh: extend allowed options for AllowTcpForwarding


duncanatwork:
AllowTcpForwarding in `ncm-ssh` uses `legacy_binary_affirmation_string`, but other options such as `remote` and `local` are valid, so schema should be changed to support valid values

PR #1691, opened 2024-07-06
ncm-ssh: extend AllowTcpForwarding to support valid values

duncanatwork:
AllowTcpForwarding in ncm-ssh uses legacy_binary_affirmation_string, but other options such as "remote" and "local" are valid, so change schema to support valid values

Address #1690


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -97,7 +97,7 @@ type ssh_daemon_options_type = {
     "AcceptEnv" ? string[]
     "AllowAgentForwarding" ? legacy_binary_affirmation_string
     "AllowGroups" ? string
-    "AllowTcpForwarding" ? legacy_binary_affirmation_string
+    "AllowTcpForwarding" ? string with match (SELF, '^yes|no|all|local|remote$')
```
ned21, line 100: choice('yes', 'no', 'all', 'local', 'remote')

is probably better here than a regex

duncanatwork:
Apologies, I've made a mess of this branch somehow, so I'll just create a new PR

PR #1695, opened 2024-08-14
ncm-metaconfig: add schema for slurm 24.05

wpoely86:
New versioned schema for slurm 24.05.


jrha (approved):

jrha:
Adding `backwards incompatible` as this changes the default, but is otherwise backwards compatible.

PR #1697, opened 2024-08-19
ncm-postfix: add support for smtpd_relay_restrictions

wpoely86:
Add support for one extra postfix parameter, new since 2.10.


ncm-postfix/src/main/pan/components/postfix/schema.pan
```diff
@@ -507,6 +507,7 @@ type postfix_main = {
     "smtpd_recipient_restrictions" ? string[]
     "smtpd_reject_udicted_recipient" ? boolean
     "smtpd_reject_udicted_sender" ? boolean
+    "smtpd_relay_restrictions" ? string[]
```
jrha, line 510: Is this something that can be validated, or is it too complex?

wpoely86: I'm not 100% sure what values are all allowed here and we don't do any validation for any of the similar options (like `smtpd_recipient_restrictions`).

jrha (approved):

PR #1700, opened 2024-08-20
ncm-metaconfig beats: add seccomp to file beat schema to allow seccomp config for filebeat.yml

### LINKED ISSUE #1699: ncm-metaconfig beats: schema 6.3 seccomp
In Rocky 9 (On Linux 3.17 and later), the default seccomp settings may prevent filebeat from starting. 
See: [https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html](https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html)

Thus, we should add the config schema to allow users to configure this setting to set a allow policy.

DonaldChung-HK:
**Why the change is necessary.**
To append the schema to support configuration `seccomp` default action. This is because in Rocky 9 filebeat may be presented to start See: [https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html)](https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html) 

resolves #1699 

**What backwards incompatibility it may introduce.**
The change has been tested on both Rocky 8 and Rocky 9 aquilon hosts with filebeat. This change allows filebeat to work on Rocky 9 while not showing negative effects for existing Rocky 8 hosts.
Services which configures filebeat.yml using this module should ensure that it continues to produce the correct configuration.

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
+++ b/ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
@@ -133,7 +133,7 @@ type beats_logging = {
     (see https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html)
 }
 type beats_seccomp = {
-    'default_action' : string
+    'default_action' : choice('errno', 'trace', 'trap', 'kill_thread', 'kill_process', 'log', 'allow');
 };
 
 @documenation{
--- a/ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
+++ b/ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
@@ -128,6 +128,14 @@ type beats_logging = {
     'level' ? string with match(SELF, '^(critical|error|warning|info|debug)$')
 };
 
+@documentation{
+    Secomp settings for RHEL with Linux >3.16
+    (see https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html)
+}
+type beats_seccomp = {
+    'default_action' : string
+};
+
 @documenation{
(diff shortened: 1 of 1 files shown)


ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
```diff
@@ -128,6 +128,14 @@ type beats_logging = {
     'level' ? string with match(SELF, '^(critical|error|warning|info|debug)$')
 };
 
+@documentation{
+    Secomp settings for RHEL with Linux >3.16
+    (see https://www.elastic.co/guide/en/beats/filebeat/6.8/linux-seccomp.html)
+}
+type beats_seccomp = {
+    'default_action' : string
```
ned21, line 136: Would a choice type make more sense here to validate the entry supplied?

ned21 (commented):
Could you tweak the commit message please to make it a bit clearer and closer to PR description?

ncm-metaconfig: add seccomp to beats schema

Amend the beats schema to support configuration seccomp default action. This is because in Rocky 9+ filebeat may be prevented to start.

ncm-metaconfig/src/main/metaconfig/beats/pan/schema_6.3.pan
DonaldChung-HK: I think this is added as an optional workaround for users in particular sys admins on a system with more recent version of Linux. Thus, I don't think it is necessary to enforce/validate the value.

ned21:
Thanks for the amendment, looking good. Could you please squash these two commits into one to keep the change atomic and history easy to read?

jrha (approved):

PR #1701, opened 2024-08-22
ncm-metaconfig: Allow basic httpd remoteip config to be expressed

gregcorbett:
### Why the change is necessary.

- Needed to allow a basic remoteip config to be expressed for httpd, see https://httpd.apache.org/docs/2.4/mod/mod_remoteip.html.
- I have added the new definitions to the global section because, even though `RemoteIPProxyProtocol` can be defined at the vhost level, only one such definition will take affect, and do so globally.

### What backwards incompatibility it may introduce.
None, the new features have been added as optional.

### DIFF
--- a/ncm-metaconfig/src/main/metaconfig/httpd/config/global.tt
+++ b/ncm-metaconfig/src/main/metaconfig/httpd/config/global.tt
@@ -3,7 +3,10 @@
                 "hostnamelookups", 
                 ]
 -%]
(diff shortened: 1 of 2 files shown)


gregcorbett:
I'm not sure why runtests - it seems to error on `ncm-sysctl`. If it's because of this change I'll need some guidance to resolve.

jrha (approved):
LGTM, test failure is an existing problem with ncm-sysctl

PR #1708, opened 2024-09-10
ncm-hostsfile: Fix NoAction support declaration

jrha:
The component uses LC rather than CAF, so cannot support NoAction properly.


ned21:
ISTR LC supports NoAction, it's just not as detailed as CAF (i.e. no diffs). Do you have a specific case where  --noaction is doing the wrong thing?

Issue #1711, opened 2024-09-11
NetworkManager config changes sometimes not retrain on reboot.


aka7:
On one of our vcs cluster, which adds two static routs, 
after applying the interface config using nmstatectl apply (via ncm-network) on reboot it seems to revert back to previous config.  running nmstatectl apply again  fixes it.

on reboot it appears to use the connection from /run and not the update config in /etc and config in run does not match the updated config in /etc/. During reboot it seem to always apply the config in /run/.

nmcli dev shows this as externally connected.
```
$ nmcli dev status
DEVICE          TYPE      STATE                   CONNECTION
lo              loopback  connected (externally)  lo
dummy_hostname  dummy     connected               dummy_hostname
eth0            ethernet  connected (externally)  eth0
eth1            ethernet  connected (externally)  eth1

# nmcli -f filename conn
FILENAME
/run/NetworkManager/system-connections/lo.nmconnection
/etc/NetworkManager/system-connections/dummy_hostname.nmconnection
/run/NetworkManager/system-connections/eth0.nmconnection
/run/NetworkManager/system-connections/eth1.nmconnection
/etc/NetworkManager/system-connections/eth0.nmconnection
/etc/NetworkManager/system-connections/eth1.nmconnection
```

from NetworkManager.conf

  
```
keep-configuration
           On startup, NetworkManager tries to not interfere with interfaces that are already configured. It does so by generating a in-memory
           connection based on the interface current configuration.

           If this generated connection matches one of the existing persistent connections, the persistent connection gets activated. If there is no
           match, the generated connection gets activated as "external", which means that the connection is considered as active, but NetworkManager
           doesn't actually touch the interface.

           It is possible to disable this behavior by setting keep-configuration to no. In this way, on startup NetworkManager always tries to
           activate the most suitable persistent connection (the one with highest autoconnect-priority or, in case of a tie, the one activated most
           recently).

           Note that when NetworkManager gets restarted, it stores the previous state in /run/NetworkManager; in particular it saves the UUID of the
           connection that was previously active so that it can be activated again after the restart. Therefore, keep-configuration does not have
           any effect on service restart.

```
That suggests that 'keep-configuration no' might work on a start/reboot to ignore the previous uuid in /run and configure the device.
I'm not sure why on this particular vcs cluter we are seeing this, but this suggests that 'keep-configuration no'  work on a start/reboot to ignore the previous uuid in /run and configure the device. Setting this value fixes this problem.

We need an option to set this in ncm-network.

PR #1712, opened 2024-09-11
ncm-network: add support for device configuration dropin file

aka7:
provide support to create drop-in file for device config and add option to add keep-configuration option

fixes [1711](https://github.com/quattor/configuration-modules-core/issues/1711)

To able to turn off device keep-configuration option.
None.


aka7:
The schema update  is only for nmstate backend, but I'm not totally sure how to include this in nmstate schema, from the separation @stdweird did.  So some input would be appreciated.

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -103,6 +104,26 @@ sub disable_nm_manage_dns
...
+    my @data = ('[device]');
+
+    if ( $nm_device_config ) {
+        foreach my $key (sort keys %$nm_device_config){
+            push @data, $key."=".$nm_device_config->{$key};
+        };
+
+        $self->verbose("setting device configuration dropin");
+    }
+    my $fh = CAF::FileWriter->new($NM_DEVICE_DROPIN_CFG_FILE, mode => oct(444), log => $self, keeps_state => 1);
```
ned21, line 119: keeps_state seems wrong here if a successful write results in a network reload?

aka7: I don't think I fully understand what keeps_state purpose is?

jrha: It means that writing the file doesn't affect the state of the system, which isn't the case here?

> **keeps_state**
>
> A boolean specifying whether a file change respects the current system state or not. A file with keeps_state will be created/modified, regardless of any value for NoAction. This is useful when creating temporary files that are required for a NoAction run.
> By default, file changes modify the state and thus keeps_state is false.

https://quattor-documentation.readthedocs.io/en/stable/CAF/CAF_FileWriter.html#public-methods

aka7: ah thanks, will remove that then if default is false, also may need to fix it in sub disable_nm_manage_dns() too.

stdweird:
@aka7 not sure the comment https://github.com/quattor/configuration-modules-core/pull/1659#issuecomment-2361418243 was supposed to be added here ?

aka7:
Probably.  once the schema separation is understood and works, happy to
move it under nmstate schema only. I needed this quick to fix a bug we
noticed with network manager. So happy to move it under nmstate schema, if
that's what you mean?

jrha:
Schema migration in #1720.

jrha:
Rebased and updated to use nmstate specific schema.

jrha (approved):

PR #1715, opened 2024-09-12
/var/tmp is supposed to be reboot persistent 

stdweird:
Requires https://github.com/quattor/aii/pull/347


ned21:
I get that using /var/tmp makes sense if you need persistence but why is persistence useful here? Doesn't this leave a file lying around that might never get cleaned up? (I don't think EL has any sort of tmp clean up by default?)

stdweird:
@ned21 yes, and the way we have been deploying nodes has always relied on this. these files have also never been cleaned up (when written to /tmp). maybe we can add a systemd-tmpfiles rule to do this.

jrha (approved):

PR #1720, opened 2024-09-20
ncm-network: Move nmstate options to new schema

jrha:
@aka7 as promised this splits out the nmstate specific parts of the schema into the new structure @stdweird designed.

This also defaults to using the new schema and to be honest I don't see a reason to keep the legacy schema file around.

Backwards incompatible from 23.6.0 for anyone setting `"nm_manage_dns" = true` or more generally for any sites using the current code from master.

### DIFF
--- a/ncm-network/src/main/pan/components/network/core-schema.pan
+++ b/ncm-network/src/main/pan/components/network/core-schema.pan
@@ -2,7 +2,7 @@
 
 declaration template components/network/core-schema;
 
-final variable QUATTOR_TYPES_NETWORK_LEGACY ?= true;
+final variable QUATTOR_TYPES_NETWORK_LEGACY ?= false;
 
 include if (QUATTOR_TYPES_NETWORK_LEGACY) 'components/network/core-schema-legacy'
     else 'components/network/types/network';
--- a/ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
+++ b/ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
@@ -1,3 +1,6 @@
 declaration template components/network/types/network/backend/initscripts;
 
 @{implement types specific for initscripts / network.pm}
+
+type structure_network_backend_specific = {
+};
--- a/ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
+++ b/ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
@@ -4,3 +4,15 @@ declaration template components/network/types/network/backend/initscripts;
 
 type structure_network_backend_specific = {
 };
+
+function network_valid_route = {
+    if (exists(SELF['prefix']) && exists(SELF['netmask'])) {
+        error("Use either prefix or netmask as route");
+    };
+
+    if (exists(SELF['prefix'])) {
+        network_valid_prefix(SELF);
+    };
+
+    true;
+};
--- a/ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
+++ b/ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
@@ -8,3 +8,20 @@ type structure_network_backend_specific = {
     @{let ncm-network cleanup inactive connections}
     "clean_inactive_conn" : boolean = true
 };
+
+function network_valid_route = {
+    if (exists(SELF['command'])) {
+        if (length(SELF) != 1) error("Cannot use command and any of the other attributes as route");
+    } else {
+        if (!exists(SELF['address']))
+            error("Address is mandatory for route (in absence of command)");
+        if (exists(SELF['prefix']) && exists(SELF['netmask']))
+            error("Use either prefix or netmask as route");
+    };
+
+    if (exists(SELF['prefix'])) {
+        network_valid_prefix(SELF);
+    };
+
+    true;
+};
--- a/ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
+++ b/ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
@@ -1,3 +1,10 @@
 declaration template components/network/types/network/backend/nmstate;
 
 @{implement types specific for nmstate / nmstate.pm}
+
+type structure_network_backend_specific = {
+    @{let NetworkManager manage the dns}
+    "manage_dns" : boolean = false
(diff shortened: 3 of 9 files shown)


ncm-network/src/main/pan/components/network/core-schema.pan
```diff
@@ -2,7 +2,7 @@
 
 declaration template components/network/core-schema;
 
-final variable QUATTOR_TYPES_NETWORK_LEGACY ?= true;
+final variable QUATTOR_TYPES_NETWORK_LEGACY ?= false;
```
stdweird, line 5: @jrha @ned21 @jouvin i suppose we should give sites one or 2 weeks time to compare generated profiles with changing this boolean.

ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
```diff
@@ -1,3 +1,59 @@
...
+
+type structure_network_backend_specific = {
+};
+
+@documentation{
+    Add route (IPv4 of IPv6)
+    Presence of ':' in any of the values indicates this is IPv6 related.
+}
+type network_route = {
+    @{The ADDRESS in ADDRESS/PREFIX via GATEWAY}
```
stdweird, line 15: i would factor out all common bits. i don't know why you have this in the backend code

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
```diff
@@ -1,3 +1,60 @@
...
+
+type structure_network_nmstate = {
+    @{let NetworkManager manage the dns}
+    "manage_dns" : boolean = false
+    @{let ncm-network cleanup inactive connections}
+    "clean_inactive_conn" : boolean = true
+};
+
+type structure_network_backend_specific = {
+    "nmstate" : structure_network_nmstate
```
stdweird, line 15: why the nmstate subtructure? the options have a generic enough name

ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
jrha: This seemed the easiest option for having different validation rules per back-end, but I now realise I could redefine the validation function in each instead.

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
jrha: I thought it best to make it clear, but happy to do whichever people prefer.

ncm-network/src/main/pan/components/network/core-schema.pan
ned21: 👍  @aka7 is probably the best person to do that for MS. We have pretty good tools for this so it should be easy to generate, but may take a little bit to analyse any differences.

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
ned21: Could we ever have manage_dns and nmstate/manage_dns? If they might conflict then putting it in a subtree makes sense, but creating a nmstate subtree might also be problematic if RH rename nmstate or move the manage_dns out of it?

OTOH, this currently looks to be a single item tree? In which case I am not sure it makes sense to push things down a level? What things do we expect to appear in structure_network_backend_specific versus structure_network_backend_specific/nmstate ?

ncm-network/src/main/pan/components/network/core-schema.pan
aka7: @jrha  thank you for this.

I plan to look into this next week. 
I suspect backward incompatibility will be with nm_clean_inactive_conn too.

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
aka7: manage_dns was only added during nmstate configuration since network manager by default updates resolve.conf.
 
However upon looking at this few weeks ago for [device] configuration drop in file,
 I think we need schema to configure options for [main] droppin file. one of its option is dns=none, which manage_dns adds in [main] config section, so I wonder if this should be refactored to add any override options of [main] configuration of networkmanager.conf drop in file.

aka7:
@jrha  the being backward incompatible is ok with us,  We probably need to move manage_vips into nmstate path,  because that is currently only for nsmstate, unless we plan to look at creating loopback interface for network.pm? this we are going have backward incompatible worth looking at other nmstate on options I added recently.

but I'm struggling to override the two variable introduced in the schema at os level, the network types are included early on in quattor/types/system.pan (in our setup at least). So I'll be interested to know how others are going override these values QUATTOR_TYPES_NETWORK_BACKEND or QUATTOR_TYPES_NETWORK_LEGACY.

Has anyone else tested this? cc @stdweird ?

aka7 (approved):
Tested this internally at ms, LGTM.
Other than the comment made about overriding the variables.  and we can possibly move other options into nmstate schema, this can be done later  separate PR once this is merged?
Others should test and confirm though.

aka7:
@stdweird @jrha just checking what is decision on this? At ms, we have tested this,  with caveat that you can't override the two global variables early on before the network schema is included, but with some refactoring in our template code we can achieve this by putting in some logic to set the right values we need, legacy, vs initscripts, vs nmstate.  And we are ok with backward incompatibility.

I ask this, because there additional changes I have in flight which I'll create pr once this is merged. Also we are looking to go GA with rhel9 before the end of the year so any major changes, I would like to get this in before we go GA. hope this makes sense.

thanks for your help.

jrha:
Sorry for the delay, we've tested this at RAL with no resulting changes to host profiles, so we're happy and I'd like to go ahead and remove the legacy schema.

ncm-network/src/main/pan/components/network/types/network/backend/initscripts.pan
jrha: I'll try and refactor this tomorrow.

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
jrha: Ok cool, I'll drop it.

stdweird:
@jrha :+1:

Issue #1722, opened 2024-09-30
ncm-opennebula: remove qxl video driver


alvarosimon:
qxl video driver and SPICE are not supported in rhel9 anymore, so it does not make sense to keep those.

`
error: unsupported configuration: domain configuration does not support video model 'qxl'`

opennebula tt file should be updated to remove qxl. OpenNebula should choose the right video driver instead.

PR #1723, opened 2024-09-30
ncm-opennebula: do not use qxl video driver

alvarosimon:
Resolves:  #1722


stdweird:
@alvarosimon any issues wrt backwards compatibility?

alvarosimon:
@stdweird no, this change should be backwards compatible with older rhel releases

alvarosimon:
Tested with RHEL8 and RHEL9 hyps

jrha (approved):

PR #1725, opened 2024-10-22
ncm-network: Remove legacy schema

jrha:
Based on #1720.

The modernised schema has been proven to be equivalent and we really don't want to end up with the schema being forked.


aka7:
this has been tested at MS and works on rhel9 which is where we use this.
our internal schema is bet more complicated due to previous legacy ncm-network code that we have in rhel7/8, so the LEGACY option actually works for us, but that's fine to be removed here.
is this likely to be merged soon or we expecting any further changes?

jrha:
#1720 needs to be merged first, but otherwise this is good to go.

PR #1735, opened 2024-11-18
ncm-network: add additional bonding options

aka7:
add additional bonding options, as per https://docs.rs/nmstate/latest/nmstate/struct.BondOptions.html#fields

ability to set bonding options such as 
resend_igcmp and others, 

None. all options added are optional.


ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -34,13 +34,25 @@ type network_interface_alias = {
     Describes the bonding options for configuring channel bonding on EL5 and similar.
```
ned21, line 34: Maybe update comment to remove reference to EL5 (which might make someone think it's out of date) and add ref of https://docs.rs/nmstate/latest/nmstate/struct.BondOptions.html#fields ?

ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -34,13 +34,25 @@ type network_interface_alias = {
...
+    "all_slaves_active" ? long(0..1)
+    "arp_interval"      ? long
+    "arp_ip_target"     ? type_ip[]
+    "arp_validate"      ? choice('none', 'active', 'backup', 'all')
+    "downdelay"         ? long
+    "fail_over_mac"     ? choice('active', 'follow', 'none')
+    "lacp_rate"         ? long(0..1)
+    "min_links"         ? long
+    "miimon"            : long
+    "mode"              : long(0..6)
```
ned21, line 47: I can't find this on https://docs.rs/nmstate/latest/nmstate/struct.BondOptions.html#fields - is there modern doc? Or should be it switched to optional?

ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -34,13 +34,25 @@ type network_interface_alias = {
...
+    "ad_select"         ? choice('bandwidth', 'count', 'stable')
+    "all_slaves_active" ? long(0..1)
+    "arp_interval"      ? long
+    "arp_ip_target"     ? type_ip[]
+    "arp_validate"      ? choice('none', 'active', 'backup', 'all')
+    "downdelay"         ? long
+    "fail_over_mac"     ? choice('active', 'follow', 'none')
+    "lacp_rate"         ? long(0..1)
+    "min_links"         ? long
+    "miimon"            : long
```
ned21, line 46: Switch to optional? The [doc](https://docs.rs/nmstate/latest/nmstate/struct.BondOptions.html#fields) suggests default is 0 (i.e. off) -- is that more sensible than forcing a user to specify a value?

aka7: I havent  changed miimon, this was already there in current release.  happy to change it to optional if agreed.

ncm-network/src/main/pan/components/network/types/network/interface.pan
aka7: that comes under bonding mode https://docs.rs/nmstate/latest/nmstate/enum.BondMode.html, 
https://docs.rs/nmstate/latest/nmstate/struct.BondConfig.html#structfield.mode

and in this example get set in link-aggregation  https://nmstate.io/examples.html#interfaces-bond, 
which is how this is being set in nmstate.pm code.

ncm-network/src/main/pan/components/network/types/network/interface.pan
jouvin: If it makes sense either change it to optional or define a default value so that there is no requirement to add it if you don't need to.

ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -31,16 +31,30 @@ type network_interface_alias = {
...
+    "ad_select"         ? choice('bandwidth', 'count', 'stable')
+    "all_slaves_active" ? long(0..1)
+    "arp_interval"      ? long
+    "arp_ip_target"     ? type_ip[]
+    "arp_validate"      ? choice('none', 'active', 'backup', 'all')
+    "downdelay"         ? long
+    "fail_over_mac"     ? choice('active', 'follow', 'none')
+    "lacp_rate"         ? long(0..1)
+    "min_links"         ? long
+    "miimon"            ? long
```
jrha, line 48:
```suggestion
    @{802.3ad aggregation selection logic}
    "ad_select"         ? choice('bandwidth', 'count', 'stable')
    @{Drop (0) or deliver(1) duplicate frames on inactive ports}
    "all_slaves_active" ? long(0..1)
    @{ARP link monitoring frequency in milliseconds}
    "arp_interval"      ? long(0..)
    @{IP addresses to use as ARP monitoring peers when arp_interval is > 0}
    "arp_ip_target"     ? type_ip[]
    @{For which ports should ARP probes and replies should be validated or ignored completely}
    "arp_validate"      ? choice('none', 'active', 'backup', 'all')
    @{Milliseconds to wait before disabling a port after a link failure has been detected by miimon}
    "downdelay"         ? long(0..)
    @{MAC address assignment policy for failover bonds}
    "fail_over_mac"     ? choice('active', 'follow', 'none')
    @{Requested LACPDU packet rate in 802.3ad mode}
    "lacp_rate"         ? long(0..1)
    @{Minimum number of links that must be active before asserting carrier}
    "min_links"         ? long(0..)
    @{Milliseconds between link state checks}
    "miimon"            ? long(0..)
```

ncm-network/src/main/pan/components/network/types/network/interface.pan
```diff
@@ -31,16 +31,30 @@ type network_interface_alias = {
...
+    "mode"              : long(0..6)
+    "num_grat_arp"      ? long
+    "num_unsol_na"      ? long
+    "primary"           ? valid_interface
+    "primary_reselect"  ? choice('always', 'better', 'failure')
+    "resend_igmp"       ? long(0..255)
+    "use_carrier"       ? boolean
+    "updelay"           ? long
+    "xmit_hash_policy"  ? choice('0', '1', '2', 'layer2', 'layer2+3', 'layer3+4')
```
jrha, line 57:
```suggestion
    @{Transmit hash policy used in balance-xor, 802.3ad and tlb modes}
    "xmit_hash_policy"  ? choice('0', '1', '2', 'layer2', 'layer2+3', 'layer3+4')
```

jouvin:
@aka7 we just merge an update to the CI tests. It should not change the result but probably it a good idea to rebase to ensure everything is ok.

jrha (approved):

jrha:
Wait, are any of these new fields supported by the newly added initscripts backend?

aka7:
> Wait, are any of these newly added fields supported by the initscripts backend?
we use few already. We are already setting the following with ifcfg, which I it works.  I haven't yet checked  for others added.
in ifcfg, you get config generated such as below from network.pm (however this version is our forked version internally) so worth someone outside MS checking this please?

BONDING_OPTS='arp_interval=8000 arp_ip_target=10.x.x.x arp_validate=all mode=active-backup primary=eth0 primary_reselect=always resend_igmp=2'

PR #1739, opened 2024-11-20
ncm-network: dropin file support for main configuration of NetworkManager

aka7:
drop-in file support for main configuration section of NetworkManager

Allow capability for adding additional options for NetworkManager.conf [main] section.

manage_dns options will no longer be needed.


aka7:
not sure I understand the indentation failure here, indentation looks ok to me.

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
```diff
@@ -5,16 +5,31 @@ declaration template components/network/types/network/backend/nmstate;
 @documentation{
     NetworkManager device configuration for drop in config file.
 }
+
+type structure_nm_main_config = {
+    @{Set the DNS processing mode for NetworkManager}
+    "dns" : choice("dnsmasq", "default", "none", "systemd-resolved")
```
ned21, line 11:
```suggestion
    "dns" : choice("default", "dnsmasq", "none", "systemd-resolved")
```

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
```diff
@@ -5,16 +5,31 @@ declaration template components/network/types/network/backend/nmstate;
...
+    @{Specify devices for which NetworkManager should not create default wired connection}
+    "no-auto-default" ? string
```
ned21, line 13: Should this be a list of strings instead?

aka7: yes probably, will require a bit more thought on this as it can be interface names or mac addresses etc, 
I'm not sure who will need this right now, we don't, just added it as bonus, so I'm inclined not to include this in this PR, and look into later, perhaps, if need arise?

jrha:
Rebased on `main`.

PR #1742, opened 2024-11-25
ncm-sudo: Add support for use_pty with sudoers

jrha:
This allows commands to be run in a pseudo-pty.
This is required by the CIS Benchmarks.


jrha:
Fix for panlint tests in #1743.

wpoely86 (approved):

PR #1748, opened 2024-11-28
ncm-authconfig: Fix wrapping and indentation in pan

jrha:


ncm-authconfig/src/main/pan/components/authconfig/sssd/ldap.pan
```diff
@@ -8,14 +8,17 @@
 
 declaration template components/authconfig/sssd/ldap;
 
-type ldap_schema = string with match(SELF, "^(IPA|AD|rfc2307|rfc2307bis)") ||
-    error ("LDAP schema must be valid according to sssd-ldap: " + SELF);
+type ldap_schema = string with match(SELF, "^(IPA|AD|rfc2307|rfc2307bis)") || error(
```
ned21, line 11: I know this is supposed to be just clean up but there's a bug here, missing $ at end.

ned21 (approved):

ncm-authconfig/src/main/pan/components/authconfig/sssd/ldap.pan
jrha: It could also be migrated to `choice`, maybe I'll do another set of PRs for that sort of thing.

jrha: Looking at the this there's deeper issues, do I'll open a bugfix PR.

PR #1767, opened 2024-11-29
ncm-ssh: Cleanup pan templates

jrha:


ncm-ssh/src/main/pan/components/ssh/schema.pan
```diff
@@ -43,13 +43,32 @@ type ssh_gssapikexalgorithms = choice(
 );
 
 function is_valid_ssh_MAC = {
```
ned21, line 45: Away from useful workstation so can't easily check, but can choice be made a list like other types?

```
type foo = {
   a : string[] # list of strings
   b : choice('x', 'y')[]  # list of choice types
};
```

ned21 (approved):

ncm-ssh/src/main/pan/components/ssh/schema.pan
jrha: AFAICT yes that works, although maybe one for further cleanup and refactoring.

PR #1783, opened 2024-12-08
metaconfig/named: fix validation of logging channels in categories

jouvin:
In `ncm-metaconfig` named schema, the statement to validate the logging channel name refered by a logging category is wrong (missing `channels` in the path), leading to validation errors, if a category is defined.

This PR fixes the issue and add a category in the test profile to ensure that this validation statement is tested.


jouvin:
`indentation` errors are not mine! And I tend to disagree on what `indentation` suggests... but if you think it should be changed to comply to the suggestion, it's easy!

ncm-metaconfig/src/main/metaconfig/named/pan/schema.pan
```diff
@@ -96,7 +96,7 @@ type named_zone = {
 };
 
 type named_channel_name = string with
```
jrha, line 98:
```suggestion
type named_channel_name = string with exists(
```

ncm-metaconfig/src/main/metaconfig/named/pan/schema.pan
```diff
@@ -96,7 +96,7 @@ type named_zone = {
...
-    exists ("/software/components/metaconfig/services/{/etc/named.conf}/contents/logging/" + SELF) ||
+    exists ("/software/components/metaconfig/services/{/etc/named.conf}/contents/logging/channels/" + SELF) ||
     error (SELF + " doesn't refer to a logging channel");
```
jrha, line 100:
```suggestion
    "/software/components/metaconfig/services/{/etc/named.conf}/contents/logging/channels/" + SELF) || error(
        "%s doesn't refer to a logging channel", SELF
    );
```

jrha:
Fixes for existing issues in #1785.

jouvin:
@jrha hopefully, I manage to fix the `indentation` errors, even though I am not totally convinced by the readibility of the end result...

jrha:
Rebased on `main` and resolved conflicts.

PR #1788, opened 2024-12-11
ncm-systemd: ensure unit property is defined

### LINKED ISSUE #1787: ncm-systemd 24.10.0: validation fails if no unit is defined
In `ncm-systemd` since 21.12.0, a validation has been added to the schema for the main type `systemd_component` that accesses `unit` property in the structure. If the component has been loaded but no unit has been defined, this result in a validation error has the `unit` property is undefined.

jouvin:
Fixes #1787

### DIFF
--- a/ncm-systemd/src/main/pan/components/systemd/schema.pan
+++ b/ncm-systemd/src/main/pan/components/systemd/schema.pan
@@ -17,7 +17,7 @@ type hwloc_location = string with match(SELF, '^[\w:.]+$');
     syslog facility to use when logging to syslog
 }
 type syslog_facility = string with match(SELF,
-    '^(kern|user|mail|daemon|auth|syslog|lprnews|uucp|cron|authpriv|ftp|local[0-7])$'
+    '^(kern|user|mail|daemon|auth|syslog|lpr|news|uucp|cron|authpriv|ftp|local[0-7])$'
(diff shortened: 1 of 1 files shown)


jrha:
Isn't the issue here that the schema marks it as optional, but the validation requires it to be present?

jrha:
How about https://github.com/jrha/configuration-modules-core/commit/1abc38d4c7928385fa186398066f891724616dbf instead?

jouvin:
@jrha I'm happy with your patch too, it should have the same effect for the use case I want to address. I think the main choice criteria between both approaches is whether the configuration module code can deal properly with an undefined `unit` property. Looking quickly at the code, I have the feeling that the code is not really prepared to deal with no units at all being defined but the actual unit list is built in `Systemd/Service.pm` by `gather_configured_units()` which merges systemd units and services defined in `ncm-chkconfig` so the risk of an empty unit list looks very small and will probably not match a meaningful usecase... So probably it doesn't make a big difference.

jrha (approved):

jouvin:
@jrha please, could you merge this PR is you agree with as it would allow to fix errors in several other PRs currently open...

jrha:
Sorry, was waiting for the tests to finish and got distracted!

PR #1800, opened 2024-12-12
ncm-metaconfig: zookeeper: Fix wrapping and indentation in pan

jrha:


ned21 (commented):
This new formulation looks a bit weird? The [] at the end is really important and perhaps easy to overlook if it's on its own line?

jrha:
Good point, thanks. I've made it a new type for clarity.

ned21 (approved):

PR #1810, opened 2025-01-30
ncm-metaconfig: updated filebeat 7.x schema to fix #1728

### LINKED ISSUE #1728: ncm-metaconfig: fixing filebeat schema 7.0 and adding additional processor
Hi,

I would like to fix/enhance the following:
- removing the `registry_file` and `prospector` from the schema, these are deprecated and will cause filebeat to fail to start tested on `7.10.x`.
- add support for the following processor to support my work.
  - `dns` for DNS reverse lookup
  -  `dissect` for tokenizing console messages
  -  `decode_json_fields` for tokenizing JSON in the messages.

DonaldChung-HK:
fix #1728

remove deprecated keys: this prevents filebeats starting
added processors: dns, dissect, decode_json_fields, convert; 
add docker log support; add secomp config


ncm-metaconfig/src/main/metaconfig/beats/pan/schema_7.0.pan
```diff
@@ -163,24 +171,96 @@ type beats_service = {
...
     'seccomp.enabled' ? boolean
+    'seccomp' ? beats_seccomp
 };
 
 @documentation{
     Handle logmessages spread over multiple lines
 }
 type beats_filebeat_input_multiline = {
-    'pattern' ? string_trimmed
+    'pattern' ? string
```
jrha, line 181: If this field really needs allow leading/trailing white-space, then it might be good to add an annotation to reflect this, so that someone doesn't try to "fix" it again in future.

ncm-metaconfig/src/main/metaconfig/beats/pan/schema_7.0.pan
```diff
@@ -163,24 +171,96 @@ type beats_service = {
...
     Handle logmessages spread over multiple lines
 }
 type beats_filebeat_input_multiline = {
-    'pattern' ? string_trimmed
+    'pattern' ? string #string type to consider for potential whitespaced pattern
     'negate' ? boolean
     'match' ? choice('after', 'before')
     'max_lines' ? long(0..)
     'timeout' ? long(0..)
+    'max_lines' ? long(0..)
```
jrha, line 186: This is now a duplicated property (see two lines above)
```suggestion
```

jrha:
Note to self: Squash when merging.

Issue #1657, opened 2024-02-20
ncm-altlogrotate: with only the global entry, and no entries marked as global, main configfifle is not edited


stdweird:
issue is on L139

`if (@globals) {` -> `if (@globals || $overallglobal) {`

PR #1813, opened 2025-02-27
ncm-altlogrotate: global logrotate config fails to be modified in some cases

StephaneGerardVUB:
See issue #1657 for more details


wpoely86 (approved):

wpoely86:
This fixes #1657

ned21 (changes requested):
Sorry if this sounds a bit paranoid but a commit message which only references a ticket in the bug tracking system could lead to problems in the future since there's no guarantee we'll be able to migrate the bug data along with git repository to pastures new. Could you please provide a full explanation of the change being made in the commit log so it documents the change alongside the code?

Issue #1821, opened 2025-06-24
ncm-opennebula: add OS/MACHINE type option for VM templates


alvarosimon:
You can define  `OS/MACHINE` to set libvirt machine type from OS VM section template:

https://docs.opennebula.io/6.10/management_and_operations/references/template.html#os-and-boot-options-section

This allows to improve the performance (for instance using q35 model in RHEL9):

```
OS=[  
 BOOT="nic0,disk0",  
 MACHINE="pc-q35-rhel9.4.0" ]
```

This should be included from ncm-opennebula schema and vmtemplate tt file.

PR #1776, opened 2024-12-03
ncm-opennebula: Add cpu pinning support

alvarosimon:
Include support for VM CPU pinning and CPU topology in OpenNebula.
Removes old legacy code.
This PR is backwards incompatible for older ONE versions < `5.10.0`

Resolves: #1775 
Resolves:  #1745

PR #1822, opened 2025-06-25
ncm-opennebula: add libvirt hw machine type option

alvarosimon:
Required to add q35 hw machine types

#1776 must be merged first 

Resolves:  #1821

### DIFF
--- a/ncm-opennebula/src/main/resources/tests/profiles/vm.pan
+++ b/ncm-opennebula/src/main/resources/tests/profiles/vm.pan
@@ -230,3 +230,5 @@ prefix "/system/opennebula";
     "cores", 1,
 );
 
+"machine" = "pc-q35-rhel9.4.0";
+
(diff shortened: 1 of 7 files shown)


ncm-opennebula/src/main/resources/tests/profiles/vm.pan
```diff
@@ -230,3 +230,5 @@ prefix "/system/opennebula";
     "cores", 1,
 );
 
+"machine" = "pc-q35-rhel9.4.0";
```
wpoely86, line 233: Shouldn't the output of a tests also be changed?

alvarosimon: @wpoely86  indeed, I have included the missing regex

wpoely86: the regex turned out to be not very strict. Fixed also.

> 
> ah ok why it was working before (actually I also forgot to include the check) is because the test itself the regex is not quite strict as you can see it was:
> ```
> ^MEMORY\s?=\s*"\d+"\s*$
> ^OS\s?=\s?\[$
> ^\s*BOOT\s?=\s*""\s*$
> ^RAW\s?=\s?\[$
> ```
> 
> so this also work if I include something after OS = [    so the regex was working indeed before, now I have included the regex explicitly so if you modify MACHINE it will fail
> so the previous check only was looking for BOOT after OS and indeed it was right.. even if I include MACHINE between... now its more explicit

PR #1824, opened 2025-06-30
ncm-openldap: minimal support of mdb database

StephaneGerardVUB:
This change is required on EL9 machines that comes with OpenLDAP 2.6. This release doesn't support bdb backend anymore, so you are obliged to migrate to another one, and the closest seems to be mdb. The optional parameter 'maxsize' that only applies with mdb was also added. It you don't set this size, it is limited by default to 1MB, which is way to low for us.
This change is backward compatible.


ncm-openldap/src/main/pan/components/openldap/schema.pan
```diff
@@ -178,7 +178,7 @@ type ldap_global = {
 
 
 type ldap_database_string = string with match(SELF,
-    "^(bdb|config|dnssrv|hdb|ldap|ldif|meta|monitor|null|passwd|perl|relay|shell|sql)$"
+    "^(bdb|mdb|config|dnssrv|hdb|ldap|ldif|meta|monitor|null|passwd|perl|relay|shell|sql)$"
```
ned21, line 181: The list is currently sorted alphabetically so can you maintain than please?

ned21 (commented):
A couple of minor comments. Can you please also rewrite the commit message to follow the standard form:

```
ncm-openldap: minimal support of mdb database

This change is required on EL9 machines that comes with OpenLDAP 2.6. This release doesn't support bdb backend anymore, so you are obliged to migrate to another one, and the closest seems to be mdb. The optional parameter 'maxsize' that only applies with mdb was also added. If you don't set this size, it is limited by default to 1MB, which is likely too low for most users.

This change is backward compatible.
```

ncm-openldap/src/main/pan/components/openldap/schema.pan
StephaneGerardVUB: fixed

PR #1827, opened 2025-07-10
ncm-metaconfig: Update defaults for SSL protocol and ciphersuite

gregcorbett:
New SSL protocol and ciphersuite options have been added, and the defaults for those fields have been refreshed in reference to the latest Mozilla intermediate profile, see https://ssl-config.mozilla.org/

https://ssl-config.mozilla.org/#server=apache&version=2.4.60&config=intermediate&openssl=3.4.0&guideline=5.7
- Generated 2025-07-10, Mozilla Guideline v5.7, Apache 2.4.60, OpenSSL 3.4.0, intermediate config
- Supports Firefox 27, Android 4.4.2, Chrome 31, Edge, IE 11 on Windows 7, Java 8u31, OpenSSL 1.0.1, Opera 20, Safari 9.

Why the change is necessary.
Provide more modern defaults.

What backwards incompatibility it may introduce.
Those using the default options will see a change, for the better i.e. getting bumped from using TLSv1.0 to using TLSv1.2 or TLSv1.3.


jrha:
Looks like `src/test/perl/service-httpd.t` has a test in it that checks the defaults and looks for `TLSv1`, so that'll need updating.

gregcorbett:
Thanks for the pointer @jrha, I'll put this back into draft while looking into the tests.

jrha (approved):

PR #1828, opened 2025-07-16
ncm-network: nmstate support for adusting congestion window

aka7:
Supporrt for adusting congestion window size. cwnd, initcwnd and initrwnd. Special application protocols with specific requirements about the TCP congestion window size may have a valid use-case to override TCP slow-start.

Capability for nmstate to set cwnd size.

None. but this feature only available with nmstate >= 2.2.45
NOTE: support for this is only provided in nmstate.ppm and not network.pm.


aka7:
Any thoughts on this?

aka7:
> As these are only implemented for `nmstate`, should the schema changes be moved to `network/types/network/backend/nmstate.pan`?

yes you are right, updated.

PR #1829, opened 2025-10-28
ncm-metaconfig: Add chrony options for dealing with leap-second adjustments on time servers

jrha:


ncm-metaconfig/src/main/metaconfig/chrony/pan/schema.pan
```diff
@@ -65,6 +65,20 @@ type chrony_service_network = {
...
+    easier for them to track it and keep their clocks close together even when large offset or frequency corrections
+    are applied to the server’s clock, for example after being offline for a longer time.
+}
+type chrony_service_smoothtime = {
+    @{ Maximum frequency offset of the smoothed time to the tracked NTP time (in ppm) }
+    'max_freq' : double(0..100000)
+    @{ Maximum rate at which the frequency offset is allowed to change (in ppm per second) }
+    'max_wander' : double(0..100000)
+    @{ Only smooth leap seconds, ignore normal offset and frequency changes }
+    'leaponly' : boolean
```
ned21: https://chrony-project.org/doc/4.8/chrony.conf.html says this is optional. I see a value of "false" essentially removes it via the .tt file, and that's less code than having to check if it's set so makes sense to me... OTOH maybe we should set it as a default of false to avoid people having to specify it?

PR #1816, opened 2025-03-31
ncm-metaconfig: httpd: support requestheader in vhost

stdweird:

PR #1830, opened 2025-11-04
ncm-metaconfig: httpd: support casscope

GoodOldJack12:
Adds the CASScope directive for [mod_auth_cas](https://github.com/apereo/mod_auth_cas)
Depends on https://github.com/quattor/configuration-modules-core/pull/1816


ncm-metaconfig/src/main/metaconfig/httpd/pan/schema.pan
```diff
@@ -594,6 +594,9 @@ type httpd_expires = {
     "default" ? string
     "bytype" ? string{}
 };
+type httpd_casscope = {
+    "path" : string
```
jrha: This looks like it should at least not allow leading or trailing whitespace:
```suggestion
    "path" : string_trimmed
```

PR #1838, opened 2026-01-05
Cleanup metaconfig perfsonar

jrha:


jrha:
Rebased on `main` again

PR #1840, opened 2026-01-05
ncm-systemd: More schema cleanup

jrha:


ned21:
```
src/test/perl/utils.t ............. 
[VERB] Log options --verbose --debug 3
Error:  Unsafe path: ;!??
ok 1 - Insane path returns nothing
ok 2 - And an error is diagnosed
ok 3 - Sane path returns a valid path
Error:  Unsafe path: foo
```

I assume this comes from us running in taint warning mode as a precursor to hopefully turning on taint as a security measure?

jrha:
> I assume this comes from us running in taint warning mode as a precursor to hopefully turning on taint as a security measure?

I believe so yes, although that hasn't made much progress.

ned21 (approved):

PR #1847, opened 2026-01-06
ncm-metaconfig: rsyslog: Add condition support to all action types

SamC-STFC:
The rsyslog module previously could not use conditional filters for every type of action, this change updates the templates and the schema to add a new `condition` field to the generic rsyslog action template. Whatever is put in the condition field will be inserted as the condition of an `if` block surrounding the action, which is generated using the existing action template.
This means that the `condition `field can now be added to any action.

This should not cause any backwards incompatibility, since the existing action template is still used to generate the action itself.

Example configuration
```
dict(
  'fwd',
  dict(
    'condition', "<condition>",
    ...
    )
)
```
Example output
```
if <condition> then {
  action(
    ...
  )
}
```


ned21:
Thanks for the commit history clean up. LGTM -- although I am not a .tt expert.

PR #1860, opened 2026-03-11
ncm-metaconfig: httpd: add more CAS options

GoodOldJack12:
* Moves all CAS options to a seperate .tt file
* Adds the following CAS options:
   * CASCookie
   * CASSecureCookie
   * CASGatewayCookie
   * CASSSOEnabled


jrha (approved):

jrha:
Even though we're in RCs, I'm going to merge this into 26.2 to avoid creating a backwards incompatibility later.

PR #1864, opened 2026-03-26
ncm-network: add option to skip managing interface.

aka7:
Adding an optional ignore option, if set to true ncm-network will not manage the interface
closes #1858 

We have a niche requirement to add interfaces in aq but do not want ncm to manage that interface. If ignore is set then we need ncm to not manage this interface but still appear in /system/network path. 
None. This is an optional field, will not impact any existing config.


aka7:
@jrha  what we discussed few weeks back, I took the liberty create this PR but open to suggestion if you don't want this in /system/network for unmanaged interface config.

stdweird (approved):

PR #1865, opened 2026-04-14
ncm-network: nmstate support for alt-names

### LINKED ISSUE #1863: ncm-network: support for altnames in nmstate.
Provide support to add altnames to nmstate config. While altnames are fully supported by RedHat they aren’t with Aquilon today,

Nmstate – These are fully supported by nmstate but only from version 2.2.51.  Syntax is to add an altnames section into the interface:
```
interfaces:
- ipv4:
    address:
    - ip: 192.168.0.2
      prefix-length: 28
    dhcp: false
    enabled: true
  name: trade1
  profile-name: prof1
  alt-names:
    - name: myint1
  state: up
  type: vlan
  vlan:
    base-iface: bond1
    id: '997
```

the ask here is to, 
•	update nmstate Network interface schema to add support for altnames as a list?
•	code updates for ncm-network nmstate module to take these alt-names from the profile and write the nmstate code.

aka7:
Support for adding alt-names to nmstate config.   
altnames are fully supported by nmstate from version 2.2.51. This currently works well on physical interfaces, but there is a bug in adding alt-names on virtual interfaces, like bond or vlan interfaces. This has been raised with RHEL and fix will be provided soon.  This shouldn't' stop this change to go in but something to bear in mind.  Since the code for ncm-netowrk won't  change we would like to get this in.
closes #1863 

Allow adding alt-names to interface configs

None


aka7:
for the bug for non physical interfaces, this PR has been created.
https://github.com/nmstate/nmstate/pull/3129

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
```diff
@@ -52,6 +52,27 @@ type structure_network_route_backend_specific = {
...
+    "name" : string
+};
+
+@documentation{
+    network interface settings specific to nmstate backend
+}
+type network_interface_backend_specific = {
+    @{Alternative interface names.
+        These values are used by the nmstate backend to generate interface alt-names entries.
+        supported from nmstate version >= 2.2.51.
```
stdweird: i am not a big fan of version numbers and/or bug reports etc in docstrings; but at least replace the comment below with something like `virtual interfaces only supported since rhel 9.8` (and if there is a 10.X version with similar fix, also add its version; 10.2 i presume?)

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -446,6 +446,13 @@ sub generate_nmstate_config
 
     $ifaceconfig->{mtu} = $iface->{mtu} if $iface->{mtu};
     $ifaceconfig->{'mac-address'} = $iface->{hwaddr} if $iface->{hwaddr};
+    if ($iface->{altnames}) {
+        my @altnames = map {
+            my $entry = {name => $_->{name}};
```
stdweird: is the intermediate variable required?

ncm-network/src/main/pan/components/network/types/network/backend/nmstate.pan
aka7: ok.  updated.

stdweird (approved):

ncm-network/src/main/perl/nmstate.pm
aka7: no, was added as there maybe other options that will come in future, such as the property and alternativenamespolicy, which we requested as RFE with rhel.

so we can do something like this. add some proeprty options, and set name policy.
   - name: altname01
      property: 'NET_UDEV_NOT_UNDER_INITRD=1'
     alternativenamespolicy='database onboard slot path mac'
      
Anyway got ai to fix this for me so extra entry is now removed. so let me know if this works for you? am happy either way.

ncm-network/src/main/perl/nmstate.pm
```diff
@@ -446,6 +446,9 @@ sub generate_nmstate_config
 
     $ifaceconfig->{mtu} = $iface->{mtu} if $iface->{mtu};
     $ifaceconfig->{'mac-address'} = $iface->{hwaddr} if $iface->{hwaddr};
+    if ($iface->{altnames}) {
+        $ifaceconfig->{'alt-names'} = [map { +{name => $_->{name}} } @{$iface->{altnames}}];
```
stdweird, line 450: what is this `+{` syntax, and what minimal perl version does it require?

jrha: This forces Perl to interpret the `{ }` as a hash constructor (rather than a block) and has apparently been present since the first Perl 5, see https://perldoc.perl.org/perlref#Curly-Brackets).

aka7: sorry missed this.
so is this ok, my perl is not great, but this was suggestion from AI :)
 happy to change it if required as what you think is right, if you can suggest something better or easier to read, then will update it. please let me know, we are keen to use this code soon.

jrha: It might be the lack of caffeine, but I'm struggling to parse the line, so haven't got any alternative to suggest right now.

jrha: Okay, so as far as I can tell it's treating `$iface->{altnames}` as an array and building `$ifaceconfig->{'alt-names'}` as a hash with `name` as the key?

N.B. The `+` is also recommended in the Perl docs https://perldoc.perl.org/functions/map.

aka7: @jrha  is there anything outstanding needed on this? or can we start using it?

PR #1867, opened 2026-05-04
ncm-metaconfig: kerberos: domain_realm wrongly configured

kwaegema:
the config option is called domain_realm, without the trailing s


ned21:
https://manpages.org/krb5conf/5 confirms this is a bug, although confusingly there is also a section called "realms".

ned21 (approved):

PR #1870, opened 2026-05-11
ncm-freeipa: add realm to the principal, get rid of error messages on freeipa servers

kwaegema:
- Currently the default realm of the server  is used with AII, this will use the realm provided in the tree

-we seem to be running freeipa/aii server with a line commented out for a long time. 
I added some logic that does the same, only for those roles.
this effectively removes the error we get on the servers, but
I'm not completely sure what the impact of his is, @stdweird could you shine your light on this, is it safe to remove this context check for the servers?

### DIFF
--- a/ncm-freeipa/src/main/perl/freeipa.pm
+++ b/ncm-freeipa/src/main/perl/freeipa.pm
@@ -486,7 +486,7 @@ sub _set_ipa_client
         if ($role eq $IPA_ROLE_AII) {
             # AII acts on behalf of a host, but is run by admin on teh AII host.
             # For now, hardcoded. Could be made configurable via config file
-            $principal = $IPA_DEFAULT_AII_PRINCIPAL;
+            $principal = "$IPA_DEFAULT_AII_PRINCIPAL\@$tree->{realm}";
             $keytab = $IPA_DEFAULT_AII_KEYTAB;
             $self->verbose("AII role $IPA_ROLE_AII using predefined principal $principal keytab $keytab")
         } elsif ($role_principal) {
@@ -506,7 +506,13 @@ sub _set_ipa_client
         keytab => $keytab,
         log => $self,
         );
-    return $self->fail("Failed to get kerberos context: $_krb->{fail}") if(! defined($_krb->get_context(usecred => 1)));
+
+    # This always fails on AII and principal servers
+    if (!$role || ($role eq $IPA_ROLE_CLIENT)) {
+        return $self->fail("Failed to get kerberos context: $_krb->{fail}") if(! defined($_krb->get_context(usecred => 1)));
+    } else {
+        $self->info("Failed to get kerberos context: $_krb->{fail}") if(! defined($_krb->get_context(usecred => 1)));
+    };
 
     # set environment to temporary credential cache
     # temporary cache is cleaned-up during destroy of $krb


ncm-freeipa/src/main/perl/freeipa.pm
```diff
@@ -486,7 +486,8 @@ sub _set_ipa_client
         if ($role eq $IPA_ROLE_AII) {
             # AII acts on behalf of a host, but is run by admin on teh AII host.
             # For now, hardcoded. Could be made configurable via config file
-            $principal = $IPA_DEFAULT_AII_PRINCIPAL;
+            my $realm = $tree->{realm};
+            $principal = "$IPA_DEFAULT_AII_PRINCIPAL\@$realm";
```
stdweird: no need for the separate variable, just use `...\@$tree->{realm}"

ncm-freeipa/src/main/perl/freeipa.pm
```diff
@@ -506,7 +507,11 @@ sub _set_ipa_client
         keytab => $keytab,
         log => $self,
         );
-    return $self->fail("Failed to get kerberos context: $_krb->{fail}") if(! defined($_krb->get_context(usecred => 1)));
+
+    # This always fails on AII and principal servers
+    if (!$role || ($role eq $IPA_ROLE_CLIENT)) {
```
stdweird, line 511: i would not supress the message, but either make it fail or info+continue

ncm-freeipa/src/main/perl/freeipa.pm
```diff
@@ -506,7 +507,11 @@ sub _set_ipa_client
         keytab => $keytab,
         log => $self,
         );
-    return $self->fail("Failed to get kerberos context: $_krb->{fail}") if(! defined($_krb->get_context(usecred => 1)));
+
+    # This always fails on AII and principal servers
```
stdweird, line 510: is this the thing we used to patch out before?

kwaegema: currently on our servers the return is removed and this error of the component ignored :)

ncm-freeipa/src/main/perl/freeipa.pm
kwaegema: I tried this now with info, but the underlying kerberos code also spit an error:
```
2026/05/12-14:41:05 [ERROR] context init attempt failed GSSAPI::Cred=SCALAR(0x565094b92a18) GSSAPI::Name=SCALAR(0x565094b92dc0) Ticket cache: FILE:/tmp/CAF-Kerberos-My_nAK/tkt
Default principal: quattor-server@QA.....

Valid starting     Expires            Service principal
05/12/26 14:41:05  05/13/26 14:38:28  krbtgt/QA.....
: GSS Error GSSAPI::Context::init: MAJOR: Unspecified GSS failure.  Minor code may provide more information MINOR: Server principal valid for user2user only.
2026/05/12-14:41:05 [INFO] Failed to get kerberos context: GSS Error GSSAPI::Context::init: MAJOR: Unspecified GSS failure.  Minor code may provide more information MINOR: Server principal valid for user2user only
```
So the freeipa component still returns an error.
@stdweird not sure the underlying error is fixable?

stdweird: i don't know. i have never fully investiagted. if you are going to skip this sort opf error, i propose a boolean in the schema that explicitly enables skipping this fail (and thus fail -> info, and no return here)

